PR #6, opened 2015-08-15
Some fixes

mvdan:


gitlab.go
```diff
@@ -123,7 +123,10 @@ func NewClient(httpClient *http.Client, token string) *Client {
 	}
 
 	c := &Client{client: httpClient, token: token, UserAgent: userAgent}
-	c.SetBaseURL(defaultBaseURL)
+	if err := c.SetBaseURL(defaultBaseURL); err != nil {
+		// should never happen since defaultBaseURL is our constant
+		panic(err)
+	}
```
svanharmelen: Hmm...  Like you say in the comment, `defaultBaseURL` is a constant we defined ourselves so this should never give an error.

On the other hand it's of course a good habit to check error values that are returned. But if we decide to do that here as well, then I would prefer to just return the error instead of throwing a panic. It feels a little wrong and dirty to have a library package throw a panic. If it just returns an error, the program using this package can decide how to handle it instead.

That being said, I do wonder what the added value is for checking the error in this particular case. Maybe the first test case we should add in this package is a test for this function. That way the test case can make sure this function works when code in the package is changed, which would make this error check here even more useless.

Thoughts?

gitlab.go
```diff
@@ -291,8 +294,7 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
 
 	response := newResponse(resp)
 
-	err = CheckResponse(resp)
-	if err != nil {
+	if err := CheckResponse(resp); err != nil {
```
svanharmelen: Why change this? Seems just a matter of taste as the result is functionally 100% the same and technically is will add some (minor) costs for the additional shadow variable that this updated code will create.

gitlab.go
```diff
@@ -300,7 +302,7 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
 
 	if v != nil {
 		if w, ok := v.(io.Writer); ok {
-			io.Copy(w, resp.Body)
+			_, err = io.Copy(w, resp.Body)
```
svanharmelen: Good catch! This should indeed be checked!

projects.go
```diff
@@ -52,17 +52,17 @@ type Project struct {
...
-		OwnerID     int    `json:"owner_id"`
-		Path        string `json:"path"`
-		UpdatedAt   string `json:"updated_at"`
+		CreatedAt   time.Time `json:"created_at"`
+		Description string    `json:"description"`
+		ID          int       `json:"id"`
+		Name        string    `json:"name"`
+		OwnerID     int       `json:"owner_id"`
+		Path        string    `json:"path"`
+		UpdatedAt   time.Time `json:"updated_at"`
```
svanharmelen: If tested OK, this seems like a good and logical change. Did you have a change to test this already? Will do so myself as well comming Monday.

svanharmelen:
@mvdan first of all thanks for your PR!! Very much appreciated :smiley: 

I do have a couple of inline remarks and one additional question I will discuss here...

I was wondering about your reasoning to change all the `[]*struct` to `[]struct`? Of course a slice is already a referenced type, so in that sense I understand that it's fine to just return the slice as that will still be "just" the reference to the actual data in memory.

But when you are going to iterate over the slice (say for searching a specific commit or comment) in the `for ... range` loop it will copy the struct into a new variable that will be used in the body of the `for` loop. So IMO using a pointer will make sure that in these cases only the pointer is copied, instead of the whole data structure.

Of course we are not talking about massive data structures here so the real effect will be very (very) small, but still it seems like the most efficient way to do this. And of course functionally using either `[]*struct` of `[]struct` is exactly the same.

So why do you want to change this?

gitlab.go
mvdan: I use panics in my libraries, for example in `mvdan/xurls` where I use `regexp.MustCompile()` which under the covers uses a `panic` if `regexp.Compile()` errors.

Then of course, I have a set of tests that must pass at every commit. These tests actually use the regex, so I know that the library won't panic under any circumstances.

We do want to check for an error. If, for any reason - e.g. the user has a modified version of Go - this fails, we can't just ignore the error and act normally. Whether to panic or to error is more your decision. I think panicking is fine because you are already making sure that this doesn't happen under normal circumstances.

OTOH, if you're happier with returning an error, I can change that.

gitlab.go
mvdan: Are you saying that the new code will be slower because it uses an extra variable via shadowing? I highly doubt that's the case :)

This is the preferred style. On one hand, shadowing instead of reassigning. This way you can move the line around without worrying if there was a `var err error` already declared.

On the other hand, a separate `if err != nil {` only makes sense if the function returned more than just an error.

projects.go
mvdan: I haven't used the time stuff yet, no. But if the others work, these should work just as well.

I was wondering whether we had to specify the date format, since Gitlab must use a specific one that is probably different from the default one that Go uses for parsing. In such a case, we would just need to declare it and make it be used everywhere.

mvdan:
Hey @svanharmelen! Thanks for the quick review.

I stumbled upon your library yesterday, after days of fighting with a more popular go gitlab library that was just painful to use. I was writing my own from scratch - sigh - when I found yours :)

Using `[]*Foo` over `[]Foo` only makes sense in very, very narrow situations. One such situation would be using `sort.Sort()`. If your data structure `Foo` is very large, using sort over `[]Foo` would mean a lot of copying around, whereas `[]*Foo` would mean just moving pointers around. The end result would be that the underlying data would be all unordered in memory, but the process would be faster.

In your case, `[]*Foo` makes no sense. You say you're worried about the range stuff, but if you were, you could do:

```go
for i := range list {
    v := &list[i] // doesn't copy, pointer
    // use v here
}
```

This is a pretty normal pattern if you want to use pointers to elements in a slice, either because you want to modify them or because you want to avoid copying the data.

In general, unless you can prove that `[]Foo` performs poorly compared to `[]*Foo` in situations like `sort.Sort()`, you never want to use `[]*Foo`. See the go-github library for example - it does exactly that :) I hope I explained myself.

svanharmelen:
@mvdan that is exactly the path I followed... I also started with the existing package, but couldn't get past the way it's implemented so decided to write a package myself :wink:

Thanks for your explanation, but I don't think I agree with your statement:

> In general, unless you can prove that []Foo performs poorly compared to []\*Foo in situations like sort.Sort(), you never want to use []\*Foo.

Besides that you don't explain the technical reasons **why** this would be true, I did some searches through the source code of the standard packages and the [camlistore](https://github.com/camlistore/camlistore) code to check if (and if so how) they use slices with pointers to structs. And the results show that it's actually used quite a lot and hardly ever i.c.w. `sort.Sort()`. So I still only see technical benefits and no downsides of using `[]*Foo` over `[]Foo` in this package.

I would really like this package to follow the best Go idioms out there, but without some more resources/references/technical reasoning I don't agree `[]Foo` would be the better or more idiomatic choice in this case.

gitlab.go
svanharmelen: As for the `[]*Foo` vs `[]Foo` discussion, I also searched the standard packages for uses of this idiom. It's indeed used quite often, but in most cases when no `err` variable already exists it the current scope. In cases where a `err` variable already exists, I notice that they tent to reuse it and not create a new shadow variable.

So again... I do want this package to follow the best Go idioms out there, but I doubt this is really one of them. Any additional info reasoning/explaining/discussing this would be appreciated to learn why you think this is (as you say it) "the preferred style".

gitlab.go
svanharmelen: Yeah... I guess the `panic` would be fine and likely even cleaner for users of the package. After we added the `panic` I will add a test for this function making sure it will never panic (unless someone altered the code and did not run the tests :wink:

projects.go
svanharmelen: Good question! Will have a look at it when testing this (which will be somewhere the week)...

gitlab.go
mvdan: (y)

gitlab.go
mvdan: It's a nit anyway, so not too fussed either way. I'll undo the err changes.

mvdan:
Ha! So we followed the same path, only that I was late :)

Regarding the `[]` vs `[]*`, at least you agree that trying to help the user for the `range` use case isn't very useful. I was under the impression that `[]` was in general preferred over `[]*`, to the point where `[]*` bothers me unless I see a reason why it's useful.

You are right, about one tenth of the slices in `golang/go` are `[]*` instead of `[]`. That comes as a surprise. I guess the only explanation I can find for those is if the elements of the slice are allocated somewhere else, so an `append` of a pointer is much cheaper than copying over the element if the structure is big.

You managed to confuse me about this, to the point where I don't know what to do with funcs that return lists anymore. :(

I'll update the PR removing that too.

gitlab.go
mvdan: BTW, I caught this with `errcheck`. Nice little util.

svanharmelen: Ah... Will have a look at that one. Thx...

svanharmelen:
Again thanks for the updates! And no worries, you managed to confuse me as well so I had a chat with some (very senior) colleague programmers to hear their point of view as well.

To make a long chat short, in order to find out what the best solution is for a given program you would have to benchmark the same program with using both `[]*Foo` and `[]Foo` to see what would be the most efficient for the program you are writing.

And that's because there are so many tiny little things that can have an impact, that's it's to hard to just say one is better then the other outside the context of the program itself. But of course these kind of actions only make sense if you are really in need of every little bit of performance you can get. So it's stays a little unclear in that sense, but it is clear it's not bad or good (or preferred) to use either one or the other.

So I tent to use a pointer to a struct almost everywhere, and to me it seems that quite a lot of projects I'm involved in they do the same. Also the 10% your talking about is probably a little to low as from the remaining 90% there are probable also a lot of slices with interfaces and slices with build-in types and/or custom types wrapping a build-in type. So the real number should be related to check many slices have structs compared to how many slices have pointer to structs...

But enough for this (nice) technical discussion, I'l go ahead and merge this one now! Thanks again for your time and efforts!

gitlab.go
mvdan: Bonus points: Look at `gometalinter`. It runs all the linters available. Some, like `golint`, can be very verbose e.g. if documentation is missing. But doing `gometalinter | less` is something I always do when starting to contribute to a Go project. On one hand it gives me an idea on how clean the code is, and on the other it can catch small typos/errors for a first PR :)

mvdan:
Yes, when I said 10% I completely overlooked that interface slices are actually pointers underneath :) So forget that metric.

I'm in no way a senior programmer, but I thought I had this pointer thing figured out. Oh well!

Thanks for merging!

PR #9, opened 2015-08-25
Add projects to group

ThomasBS:
This will add the ability to fetch projects for when calling ``GetGroup`` (projects will be returned in an array if there's any projects assigned to the group)


svanharmelen:
LGTM!

Thanks! :smiley:

PR #13, opened 2015-09-22
Fix documentation about ListProjectIssues method

difro:
Current master has wrong docs(ListIssues instead of ListProjectIssues)


svanharmelen:
Thanks! LGTM!

PR #19, opened 2015-11-04
Added support for project permissions attribute

### LINKED ISSUE #17: Add support for Permissions attribute for get single project endpoint
**Permissions** attribute is not supported
http://doc.gitlab.com/ce/api/projects.html#get-single-project

martinsefcik:
Fixes #17


gitlab.go
```diff
@@ -52,6 +52,22 @@ const (
...
+// List of available notification levels
+//
+// GitLab API docs: http://doc.gitlab.com/ce/...?
+const (
+	DisabledNotifications NotificationLevel = 0
+	ParticipatingNotifications NotificationLevel = 1
+	WatchNotifications NotificationLevel = 2
+	GlobalNotifications NotificationLevel = 3
+	MentionNotifications NotificationLevel = 4
+)
```
svanharmelen: Probably cleaner to write this using `iota`:

```go
const (
	DisabledNotifications NotificationLevel = iota
	ParticipatingNotifications
	WatchNotifications
	GlobalNotifications
	MentionNotifications
)
```

projects_test.go
```diff
@@ -0,0 +1,32 @@
...
+	ts, client := Stub("stubs/projects/index.json")
+
+	opt := &ListProjectsOptions{}
+	projects, _, err := client.Projects.ListAllProjects(opt)
+
+	assert.Equal(t, err, nil)
+	assert.Equal(t, len(projects), 2)
+	assert.Equal(t, projects[0].Name, "project")
+	assert.Equal(t, projects[1].Name, "project2")
+	defer ts.Close()
```
svanharmelen: This `defer` should be on line 10 to have the desired effect.

projects_test.go
```diff
@@ -0,0 +1,32 @@
+package gitlab
+
+import (
+	"github.com/stretchr/testify/assert"
```
svanharmelen: Not sure if I like using `testify` over just using the standard `Go` approach, but at least there are some test now :wink:

projects_test.go
```diff
@@ -0,0 +1,32 @@
...
+	ts, client := Stub("stubs/projects/show.json")
+
+	project, _, err := client.Projects.GetProject(1)
+
+	assert.Equal(t, err, nil)
+	assert.Equal(t, project.Name, "project")
+	assert.Equal(t, project.Namespace.Name, "group")
+	assert.Equal(t, project.Permissions.ProjectAccess.AccessLevel, MasterPermissions)
+	assert.Nil(t, project.Permissions.GroupAccess)
+	defer ts.Close()
```
svanharmelen: The same goes for this `defer`... It should be on line 23.

projects_test.go
```diff
@@ -0,0 +1,32 @@
...
+}
```
svanharmelen: Please end the file with a newline to avoid diffs on this in later commits.

stubs/projects/index.json
```diff
@@ -0,0 +1,84 @@
...
+      "created_at": "2014-01-05T22:01:51.000+01:00",
+      "updated_at": "2014-01-05T22:51:08.000+01:00",
+      "description": "",
+      "avatar": null
+    },
+    "avatar_url": null,
+    "star_count": 0,
+    "forks_count": 0
+  }
+]
```
svanharmelen: Same thing, please end the file with a newline to avoid diffs on this in later commits.

stubs/projects/show.json
```diff
@@ -0,0 +1,51 @@
...
+  "star_count": 0,
+  "forks_count": 0,
+  "permissions": {
+    "project_access": {
+      "access_level": 40,
+      "notification_level": 3
+    },
+    "group_access": null
+  }
+}
```
svanharmelen: And again :wink:

svanharmelen:
@martinsefcik another great PR! Thanks a bunch! I do have a few small comments this time, but it are all very minor things... Otherwise looks great!

martinsefcik:
Thanks for code review, I will try to fix it.
I'm thankfull for any hints and comments because I just started with Go language 2 days ago :) And these two PRs are my first written in Go.

gitlab.go
martinsefcik: Yes, it's cleaner, but you don't know quickly say which value belongs to which notification level. You have to always count lines to get value. :) And maybe in the future Gitlab developers will add new notification level and it's value for some reason will not be 5 but maybe 10.

But anyway if you wish I can replace it. Let me know.

svanharmelen: Thanks for the update... Could you maybe point me to the related GitLab docs or code? As I fail to find some details about this one...

projects_test.go
```diff
@@ -0,0 +1,51 @@
...
+	if project.Namespace.Name != "group" {
+		t.Fail()
+	}
+	if project.Permissions.ProjectAccess.AccessLevel != MasterPermissions {
+		t.Fail()
+	}
+	if project.Permissions.GroupAccess != nil {
+		t.Fail()
+	}
+}
```
svanharmelen: Very nice! The only thing left is using the a [different func](https://golang.org/pkg/testing/#B.Errorf) to mark the test as failed, so we can pass along some output. Normally this is done in a style/format like this:

```go
t.Errorf("Expected project name %q, got %q", "project", project.Name)
```

The same goes for `t.Fatal(err)` which could be more expressive by typing:

```go
t.Fatalf("Expected error nil, got: %v", err)
```

gitlab.go
martinsefcik: I found it only in the code:
https://github.com/gitlabhq/gitlabhq/blob/master/app/models/notification.rb#L3

svanharmelen: Check... Well in this case I think it would be better/cleaner to use the `iota` solution instead. I looks like they are also using these values purely for enumeration, so I suggest we do the same.

The reasoning that it would be hard (counting the lines) to get the value is not applicable as that's the whole reason these constants exists right? So you don't have to memorise the numbers, but can use the written labels :wink:

projects_test.go
martinsefcik: Yes, that makes sense. I fixed it. 
I'm currently also looking at go-github client tests. 
https://github.com/google/go-github/blob/master/github/repos_test.go
And I like them :) Maybe we should implement tests the same way, but maybe not as part of this PR (?)
What do you think ?

svanharmelen: That looks very nice indeed... But I guess it would be better to add/change that in a separate PR as this PR is good to go now :wink:

svanharmelen:
Thanks for the additional updates/tweaks! LGTM!

PR #20, opened 2015-11-22
Rewritten tests

martinsefcik:
Inspired by google's go-github tests.
Added few new tests for 'projects' API endpoint.


gitlab.go
```diff
@@ -187,9 +187,6 @@ func (c *Client) SetBaseURL(urlStr string) error {
 		return err
 	}
 
-	// Set the encoded opaque data
-	c.baseURL.Opaque = c.baseURL.Path
-
```
svanharmelen: Why are you deleting this? It's in here for a very specific reason and without it I'm pretty sure stuff will break.

Take for example the [get-single-project](http://doc.gitlab.com/ee/api/projects.html#get-single-project) call. This one (as almost all other calls) requires an ID which can take the form of `NAMESPACE/PROJECT`. This value will be places into the path, but should be treated as a single value in the URL. So this needs to be URL encodes and then put into the path.

The problem is that once you did that and then create a request for the URL it will be URL encoded again which will break the API call(s). You prevent that second encode from happening by using the opaque stuff.

gitlab_test.go
```diff
@@ -0,0 +1,115 @@
...
+	"net/http/httptest"
+	"net/url"
+	"reflect"
+	"strings"
+	"testing"
+)
+
+var (
+// mux is the HTTP request multiplexer used with the test server.
+	mux *http.ServeMux
```
svanharmelen: Why is this a global var? I don't think `mux` is used anywhere outside `setup()` right? So I would not make this a separate var.

gitlab_test.go
```diff
@@ -0,0 +1,115 @@
...
+	client *Client
+
+// server is a test HTTP server used to provide mock API responses.
+	server *httptest.Server
+)
+
+// setup sets up a test HTTP server along with a gitlab.Client that is
+// configured to talk to that test server.  Tests should register handlers on
+// mux which provide mock responses for the API method being tested.
+func setup() {
```
svanharmelen: Instead of using a global `client` and `server` var, I would let this func return those values (as I believe that is much more idiomatic). So it would be something like:

```go
func setup() (*Client, *httptest.Server) {
  // test server
  server := httptest.NewServer(http.NewServeMux())

  // gitlab client configured to use test server
  client := NewClient(nil, "")
  client.SetBaseURL(server.URL)

  return client, server
}
```

gitlab_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+	want := &ErrorResponse{
+		Response: res,
+		Message:  "m",
+		Errors:   []Error{{Resource: "r", Field: "f", Code: "c"}},
+	}
+	if !reflect.DeepEqual(err, want) {
+		t.Errorf("Error = %#v, want %#v", err, want)
+	}
+}
```
svanharmelen: _nitting_ please end all files with a newline char

projects_test.go
```diff
@@ -1,52 +1,176 @@
...
-func TestListAllProjects(t *testing.T) {
-	ts, client := Stub("stubs/projects/index.json")
-	defer ts.Close()
+func TestListProjects(t *testing.T) {
+	setup()
+	defer teardown()
 
-	opt := &ListProjectsOptions{}
-	projects, _, err := client.Projects.ListAllProjects(opt)
+	mux.HandleFunc("/projects", func(w http.ResponseWriter, r *http.Request) {
```
svanharmelen: Ah... Here you are using the `mux` I commented about earlier :wink: In that case I would say this one also needs to be returned by the `setup()` func...

gitlab_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+func responseBody(w http.ResponseWriter, filename string) {
+	body, _ := ioutil.ReadFile(filename)
+	w.Write([]byte(body))
+}
+
+func TestNewClient(t *testing.T) {
+	c := NewClient(nil, "")
+
+	if got, want := c.BaseURL().String(), defaultBaseURL; got != want {
```
svanharmelen: I think you've gotten a little overboard with the `got`, `want` part :wink: This is just as clean if not cleaner (IMO):

 ```go
if c.BaseURL().String() != defaultBaseURL {
  t.Errorf("NewClient BaseURL is %s, want %s", c.BaseURL().String(), defaultBaseURL)
}
```

Also notice the `%s` instead of the `%v`. We get a string here so let's format it as one. `%v` has a different use case and can be a saver option when formatting and error (`err` is not always a standard error everywhere). But if we are getting a string please use the proper formatting option.

File: gitlab_test.go [hunk 7]
```diff
@@ -0,0 +1,115 @@
...
+		Request:    &http.Request{},
+		StatusCode: http.StatusBadRequest,
+		Body: ioutil.NopCloser(strings.NewReader(`{"message":"m",
+			"errors": [{"resource": "r", "field": "f", "code": "c"}]}`)),
+	}
+	err := CheckResponse(res).(*ErrorResponse)
+
+	if err == nil {
+		t.Errorf("Expected error response.")
+	}
```
svanharmelen: I would write this as:

```go
if err := CheckResponse(res).(*ErrorResponse); err == nil {
  t.Errorf("Expected error response.")
}
```

gitlab.go
martinsefcik: URL was changed in Go 1.5 and this workaround with Opaque is not needed. (See: https://github.com/golang/go/commit/874a605af0764a8f340c3de65406963f514e21bc).
I did small fix and added additional test for finding project by `NAMESPACE/PROJECT`. It's working in Go 1.5.
Now I realized that travis job is configured to use Go 1.4. This is the reason why test is failing. Do we want to stay at 1.4 or can we upgrade to 1.5?

gitlab_test.go
martinsefcik: Fixed

gitlab_test.go
martinsefcik: Fixed

gitlab_test.go
martinsefcik: I'm sorry. 
I will try to setup some "macro" during file saving in IntelliJ to do it automatically.

projects_test.go
martinsefcik: Fixed

gitlab_test.go
martinsefcik: Fixed

gitlab_test.go
martinsefcik: Not fixed.
err variable cannot be defined only in "if" scope, because is used also in code below.

gitlab.go
svanharmelen: Very nice, wasn't aware of that change/commit :smile: I'm fine with moving to Go 1.5, but we should mention this minimum requirement very clearly in the README.

gitlab_test.go
svanharmelen: Check! Missed that one I guess...

gitlab_test.go
svanharmelen: No worries :wink:

svanharmelen:
Again a great PR, so thanks for that! Will check the tests tomorrow with Go 1.5 and if all is good, merge this one in...

gitlab.go
svanharmelen: After a good night sleep, I decided I would like to hold of moving to Go 1.5 as a minimum requirement for this package. As we don't know who is using this package and what versions of Go they might be using, this could potentially break software using this package.

Next to that doesn't change or fix anything other then that to code is slightly cleaner. So let's wait with changing this to something that requires at least Go 1.5 for now...

projects_test.go
```diff
@@ -1,52 +1,198 @@
...
+
+	if err != nil {
+		t.Errorf("Projects.CreateProject returned error: %v", err)
+	}
+
+	want := &Project{ID: Int(1)}
+	if !reflect.DeepEqual(want, project) {
+		t.Errorf("Projects.CreateProject returned %+v, want %+v", project, want)
+	}
+}
```
svanharmelen: Missed one... :wink:

svanharmelen:
PR looks good now, but as commented let's not force people to use at least Go 1.5 for the time being...

gitlab.go
martinsefcik: As you wish, I can revert all changes related to 1.5.

But are there any known incompatibilities or big changes between 1.4 and 1.5?
1.5 is out for more than 3 months and some of my favorite projects written in go like docker and drone already use 1.5.x.
So I don't think It should be a big problem for another software using this package to upgrade to 1.5 if they want to use this package ... Additionally they will benefit from Go 1.5 improvements (garbage collector related, multicore support in scheduler external dependencies vendoring, ..)    :stuck_out_tongue_winking_eye: :sunglasses:

Let me know if you change your mind.  :smile:  I would like to work on it today evening (CET).

svanharmelen: I'm totally with you concerning the use of Go 1.5, but I don't think it's up to me or you to decide when people upgrade their software.

Next to that this change doesn't fix anything, so functionally there is no reason to change the current implementation. And with the current implementation the Go version you use doesn't matter. So changing the code doesn't fix or improve anything but it does limit the usability of the package in other packages...

Think it's clear that we should not change this now :wink:

martinsefcik: OK, I reverted back Go 1.5 related changes. I just made one small change in this Opaque related workaround.
I moved the following logic from SetBaseURL
```
c.baseURL.Opaque = c.baseURL.Path
```
to the newRequest method:
```
u.Opaque = c.baseURL.Path + path
```
So in such case baseUrl variable is not affected by this workaround and BaseURL method always return full url: 
```
https://gitlab.com/api/v3/ 
```
instead of: 
```
https:/api/v3/
```

svanharmelen:
LGTM! Thanks @martinsefcik!

martinsefcik:
Thanks!

PR #24, opened 2015-11-27
Test both 1.4 and 1.5 on Travis, remove unused code

mvdan:


mvdan:
I noticed that someone tried to add 1.5 to Travis, and you rejected it since it would stop testing 1.4. Why not test both? In fact, you could test any Go versions you wished. But I think the latest versions of 1.4 and 1.5 are a sane option.

svanharmelen:
@mvdan well I didn't reject it because it would mean we would stop testing 1.4.x :wink:

I rejected it because in that same PR he also changed the code so it would **only** work with Go 1.5 onwards. So if I would have accepted the PR in that state, it would mean people still using Go 1.4.x could no longer use the latest version of this package...

I'm fine with testing this code agains both Go 1.4.x and 1.5.x of course, that only improves reliability right :+1:

mvdan:
Right, I hadn't noticed the code bits that required 1.5.

martinsefcik:
I was using Url.RawPath which is Go 1.5 specific feature.

PR #26, opened 2015-12-02
Use json encoded request body for payload of POST and PUT

760379:


svanharmelen:
Sorry for the delay and big thank you for the PR @760379! :+1:

PR #31, opened 2016-02-09
Clarify usage in README

brasilikum:
- rename var client to git like in the other examples
- show that baseurl expects /api/v3


svanharmelen:
Thanks!

PR #32, opened 2016-02-10
make error message more usefuly by revealing type

brasilikum:
what is your thought on this? Is it worth introducing reflect to make the error more clear?

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -19,12 +19,12 @@ package gitlab
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"reflect"
 	"strconv"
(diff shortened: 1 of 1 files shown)


svanharmelen:
In general people using this package will never call `parseID()` directly, so I wonder how/when you encountered this error? I guess while making changes to the package itself?

So not sure if it has any real value, but if we do want to add more details (which is fine of course) I don't want to use reflect for something that can be done by normal formatting: http://play.golang.org/p/zJJgusMPGr

So I suggest to use either the `%#v` from the example, or `%T` to only show the type info. But if we're going to add more details I suggest the first options as then you have the full details in the error message.

brasilikum:
You are right of course, have not considered that but changed that now.
I actually encountered it when trying to create an issue (see #33).

svanharmelen:
:+1:

PR #38, opened 2016-03-12
Added support for Slack service

thecanadianbaker:
Added support for Slack service per http://doc.gitlab.com/ce/api/services.html#slack


svanharmelen:
Looks great! Thanks for the PR :+1:

PR #40, opened 2016-03-17
Support for user identities & fix for block/unblock

peterm0x:
User identities were added in V8.3 although the API docs does not reflect this yet (tested on V8.4.4 with ldap users).

Fix for block/unblock user - the response does not contain User object, so the unmarshaling fails. The API docs does not define the response body, V8.4.4 seems to return simple boolean.

### DIFF
--- a/users.go
+++ b/users.go
@@ -18,6 +18,7 @@ package gitlab
 
 import (
 	"fmt"
+	"io/ioutil"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


svanharmelen:
@peterm0x thanks for the PR!

As for the updates on the (un)block calls, would it be an idea to just return an error here instead of the response and an error?

Something like:

```go
func (s *UsersService) BlockUser(user int) error {
  u := fmt.Sprintf("users/%d/block", user)
  		  
  req, err := s.client.NewRequest("PUT", u, nil)
  if err != nil {
    return err
  }
  		  
  resp, err := s.client.Do(req, nil)
  if err != nil {
    return err
  }

  switch resp.StatusCode {
  case 200:
    return nil
  case 403:
    return errors.New("Cannot block a user that is already blocked by LDAP synchronization")
  case 404:
    return errors.New("User does not exists")
  default:
    return fmt.Errorf("Received unexpected result code: %d", resp.StatusCode)
  }  
}
```

That will save users from this package from having to check this themselves every time they need to use this func.

users.go
```diff
@@ -390,39 +397,37 @@ func (s *UsersService) DeleteSSHKeyForUser(user int, kid int) (*Response, error)
...
 
 	req, err := s.client.NewRequest("PUT", u, nil)
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 
-	usr := new(User)
-	resp, err := s.client.Do(req, usr)
+	resp, err := s.client.Do(req, ioutil.Discard)
```
svanharmelen: Please send `nil` instead of `ioutil.Discard`, as this will save reading the body for no use (there is a `nil` check in the `s.client.Do` func for this)

peterm0x:
Hmm, would it make sense to return nil when block/unblock fails with "already blocked/unblocked" ? Alternatively custom struct so that clients could ignore those errors.

svanharmelen:
According to the docs, these are the only valid (to be expected) return codes:

>Will return 200 OK on success, 404 User Not Found is user cannot be found or 403 Forbidden when >trying to block an already blocked user by LDAP synchronization.

So I don't see when/how you can get an error that block/unblock fails with "already blocked/unblocked"? How can you tell that based on the documented return codes?

peterm0x:
Ok, I have checked the gitlab sources, so the "already blocked by LDAP synchronisation" is related only to ActiveDirectory. Got a bit mixed up with it, forget that previous comment ;-)

svanharmelen:
LGTM! Thanks!

PR #41, opened 2016-03-17
Allow empty token

mishak87:
Useful when using oauth2 http client like [golang.org/x/oauth2](https://godoc.org/golang.org/x/oauth2#NewClient).


svanharmelen:
Thanks! LGTM :+1:

PR #46, opened 2016-05-11
Fixes an internal server error when no label available

matm:
Error occurs when trying to create an issue with no labels at all. In
this case, the POST body parameter will be `"labels":[""]` instead of
the expected `"labels":[]"`.


matm:
How to reproduce the error before this patch:
```shell
$ curl -XPOST https://server/api/v3/projects/:id/issues -H "Private-Token: token" \
  -H "Content-Type: application/json" \
  -d '{"title":"My issue","description":"desc","labels":[""]}'
```

svanharmelen:
Thanks for the fix @matm! LGTM!

PR #47, opened 2016-05-12
Fall back to passing raw query arguments with the request.

matm:
A bug was introduced by commmit 0e4fc277 in `gitlab.go`. Before that, request parameters were supplied as raw query arguments with `query.Values()`, like
```
description=This+is+a+body.&labels=aaa&title=My+issue
```
  
for all request methods (even POST/PUT), the same as documented in GitLab's official API doc.

But as of commit 0e4fc277 (2015-12-02), those arguments are now passed in the request body instead, which is fine but introduces a bug since labels are now sent as a `list` `{"labels":["aaa,bbb"]}` where GitLab still expects a comma-separated label names as a `string` parameter.

This produces an internal server error:
```
$ curl -XPOST https://server/api/v3/projects/:id/issues \
  -H "Private-Token: token" -H "Content-Type: application/json" \
  -d '{"title":"My issue","description":"This is a body.","labels":["aaa"]}'
```

The correct request should be instead:
```
  -d '{"title":"My issue","description":"This is a body.","labels":"aaa"}' 
```

So this block of code will marshal all options, rendering slices as JSON lists where GitLab expects a string of comma-separated values:
```
if method == "POST" || method == "PUT" {
...
}
```

Commenting this block of code makes everything work fine again, so I think this is safer to revert that change an continue passing raw query arguments, don't you think?

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -17,7 +17,6 @@
 package gitlab
 
 import (
-	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


svanharmelen:
@matm so I assume the change that was merged yesterday is not needed after all (as I understand that things still don't work out as expected)?

I must admit that I don't know if simply reverting the commit you reverenced is the right solution. I think I know a better/cleaner way to solve this, but am away for a long weekend so not behind a desktop until Wednesday.

So if you can spare a few day's, I'll get back on this as soon as I'm available again.

matm:
You are right. That change is not needed anymore indeed. There may be a better way to solve it (with maybe a few more lines of code than the original 0e4fc27 patch) but I need my program that depends on this lib to work ASAP :)

Let's talk about it in a few days then, have a nice week-end!

svanharmelen:
Closing this one in favour of PR #48

PR #48, opened 2016-05-13
Fix label marshalling

svanharmelen:


svanharmelen:
@matm I didn't have time to test all use cases, so please let me know if this PR properly fixes the `label` issue. It should as far as I can tell and IMO this is a more generic and clean solution...

matm:
Thanks for your quickc action. Should do the job indeed. Are we sure only labels are concerned? I mean no other list fields in GitLab's API doc that would require a custom marshaling action?

svanharmelen:
A first scan only shows the labels field. But if other pop up, the fix could be generalized by renaming `type Labels` to a more generic name and use it for the other fields as well. But for now I didn't spot them...

matm:
Okay, should be fine. I'll test it today. Thanks.

PR #49, opened 2016-05-13
labels: marshal error fix

matm:
After testing recent labels marshalling changes in #48, we still got a marshal error for labels from the `json` package. Here is a fix. Works perfectly now.


svanharmelen:
LGTM!

PR #50, opened 2016-05-16
Support the last_sign_in_at user field

Lujeni:
Hello,

This field could be useful to retrieve the "deprecated" users list.
http://docs.gitlab.com/ee/api/users.html#for-admins

Thanks


svanharmelen:
LGTM! Thanks for the PR @Lujeni

PR #56, opened 2016-07-01
support oauth token.

simpleelegant:


svanharmelen:
Looks great, thanks! If you could add just one test using the new OAuth client func, that would be great!

simpleelegant:
@svanharmelen Sorry, I have no idea for test, because I think the changes of this commit too little, and I'm unfamiliar with test in the past day.

svanharmelen:
OK... In that case I will add a test to verify that the code in this PR works correctly. No problem :wink:

dobegor:
That's what I was just going to implement. Thanks @simpleelegant !
Hope that gets approved ASAP.

svanharmelen:
I don't have enough time right now, but since there is other untested code in this package as well, I'll go ahead and merge it so you can start using and testing it. Will add some unit tests later on...

dobegor:
@svanharmelen thanks! I will report any errors if any.

Issue #29, opened 2015-12-19
Structs should use pointers for fields


martinsefcik:
I think we should change fields in structs to use pointers and so distinguish between nil and empty values.

The same approach is used in go-github repository.
Here you can find detailed blog why it is good to use pointers:
https://willnorris.com/2014/05/go-rest-apis-and-pointers

It's also problem in go-gitlab.
For example you cannot edit existing project's description and change it to empty string, because empty description value will never be sent to server (will be omitted).

What do you think ?

svanharmelen:
I think I agree, but want to think about it for another couple of days before we start changing stuff.

And when we do start changing this, I prefer to change it with one PR for the whole package. This way everything stays consistent and we don't have two different patterns at the same time.

So I'll come back on this one, but like I said I think I agree :grinning:

svanharmelen:
@martinsefcik after thinking about this for a little while I agree that we would need to change this. The problem is of course that a change like this would be a breaking change, but I think there is no way to get around that...

So let's go ahead and (when time permits) start the work on this one :grinning:

Since it's a breaking change I do prefer one big PR for the whole change and some minimal helper functions that can be used by users of this package to similar to what AWS has done https://github.com/aws/aws-sdk-go/blob/master/aws/convert_types.go

With that's in place I think the burden of updating code does not outweigh the benefits from having a better API on which we can continue to build for the future.

martinsefcik:
If I will have some free time I can start working on it.

svanharmelen:
Fixed by PR #66

martinsefcik:
👍

mvdan:
This is a rather surprising change, mostly because the API now seems very verbose and repetitive for no reason. Then again, I can't think of a better solution and it seems like `go-github` does the same.

svanharmelen:
@mvdan I understand your point and I partly agree... Yet I don't agree it is "for no reason" as certain use cases do not work without this change...

And yes, I too cannot think of a better/other way so following the `go-github` approach feels like the best way forward.

mvdan:
> Yet I don't agree it is "for no reason" as certain use cases do not work without this change...

I meant to say "at first glance" :) It seems like a non-idiomatic way to do things, but as said it seems like the only way.

svanharmelen:
[...]

PR #60, opened 2016-07-19
Add visibility_level parameter to CreateGroupOptions

### LINKED ISSUE #59: Visibility parameter for CreateGroup is missing
Check out [this](http://docs.gitlab.com/ce/api/groups.html#new-group).

dobegor:
Fixes #59.

### DIFF
--- a/groups.go
+++ b/groups.go
@@ -95,10 +95,10 @@ func (s *GroupsService) GetGroup(gid interface{}) (*Group, *Response, error) {
 //
 // GitLab API docs: http://doc.gitlab.com/ce/api/groups.html#new-group
 type CreateGroupOptions struct {
-	Name        string          `url:"name,omitempty" json:"name,omitempty"`
-	Path        string          `url:"path,omitempty" json:"path,omitempty"`
-	Description string          `url:"description,omitempty" json:"description,omitempty"`
-	Visibility  VisibilityLevel `url:"visibility_level" json:"visibility_level"`
+	Name            string          `url:"name,omitempty" json:"name,omitempty"`
+	Path            string          `url:"path,omitempty" json:"path,omitempty"`
+	Description     string          `url:"description,omitempty" json:"description,omitempty"`
+	VisibilityLevel VisibilityLevel `url:"visibility_level" json:"visibility_level"`
 }
 
 // CreateGroup creates a new project group. Available only for users who can
--- a/groups.go
+++ b/groups.go
@@ -95,10 +95,10 @@ func (s *GroupsService) GetGroup(gid interface{}) (*Group, *Response, error) {
 //
 // GitLab API docs: http://doc.gitlab.com/ce/api/groups.html#new-group
 type CreateGroupOptions struct {
-	Name        string `url:"name,omitempty" json:"name,omitempty"`
-	Path        string `url:"path,omitempty" json:"path,omitempty"`
-	Description string `url:"description,omitempty" json:"description,omitempty"`
-	Visibility  int    `url:"visibility_level" json:"visibility_level"`
+	Name        string          `url:"name,omitempty" json:"name,omitempty"`
+	Path        string          `url:"path,omitempty" json:"path,omitempty"`
+	Description string          `url:"description,omitempty" json:"description,omitempty"`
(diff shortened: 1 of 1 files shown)


groups.go
```diff
@@ -98,6 +98,7 @@ type CreateGroupOptions struct {
 	Name        string `url:"name,omitempty" json:"name,omitempty"`
 	Path        string `url:"path,omitempty" json:"path,omitempty"`
 	Description string `url:"description,omitempty" json:"description,omitempty"`
+	Visibility  int    `url:"visibility_level" json:"visibility_level"`
```
svanharmelen: There is a type for the visibility level that you should use instead: https://github.com/xanzy/go-gitlab/blob/master/gitlab.go#L87

Also in other places this is called `VisibilityLevel` instead of just `Visibility`. If you can change both these things we should be good to go!

svanharmelen:
Thanks for the PR @dobegor! Did leave a small comment, but if you can update accordingly all should be good to merge...

dobegor:
@svanharmelen I would like to make it called `Visibility` as it's called in API and to distinguish variable name and type.

svanharmelen:
@dobegor sorry, but I don't think that is a good idea. The field is called `visibility_level` in the API, so the correct/expected Go field name should be `VisibilityLevel` (IMO). Next to that there are already other types in this package that have a `VisibilityLevel` field, so also for consistency I do not want to introduce `Visibility` as a new field name for the same thing.

dobegor:
@svanharmelen yes, you're right, I haven't seen these fields earlier. 
If we set `omitempty` it will create Public groups instead of Private by default. So I left it to fill the default (0 as it's integer) value if not specified something different as a simple security concern to make everything Private by default.

svanharmelen:
@dobegor sorry to disagree with you again, but I don't think we should make these kind of decisions for users of this package.

This is an optional field so it should be optional in here too. Meaning that if not specified, the default behaviour of the API should be expected.

Now I do see an issue with this, as this would mean you cannot use the `private` level as the is `0`. In other places this is fixed by making this a pointer field:

```go
VisibilityLevel *VisibilityLevel `url:"visibility_level,omitempty" json:"visibility_level,omitempty"`
```

This would be the most clean solution which is the closed to how the API works. I will see if I can invest some time to implement/fix issue #29 as this would make this (using a pointer) a standard for all fields which fixes more of these use cases elsewhere in the package.

svanharmelen:
LGTM! Thanks again!

dobegor:
@svanharmelen wait

dobegor:
We can't take an address of i.e. `&gitlab.PrivateVisibility` as it's `const`.

dobegor:
@svanharmelen Please apply [this](https://github.com/dobegor/go-gitlab/commit/6981369fd5c5b5163436f44d58d97f58a3eacfd0) as a hot fix.
Sorry for troubles, was testing it in a bit different way :)

svanharmelen:
No problem @dobegor! I'm doing this all from my phone, so I guess I didn't look good enough myself too... :wink:

I deleted the last commit hoping that nobody pulled in the few minutes it was in master, so we have some time to properly add this change the correct way. Which maybe comes down to actually fixing #29 right away.

I will check things tomorrow when I'm behind a desk again. Maybe the quick fix solution can be merged first, but give me a day to make sure we are doing the right thing (would not like to make another mistake caused by not being able to check things properly from my phone).

Will be in tough tomorrow...

svanharmelen:
@dobegor could you create a new PR with the same line, except for the `omitempty` parts?

```go
VisibilityLevel VisibilityLevel `url:"visibility_level" json:"visibility_level"`
```

Of course I can add this myself as well, but it was your change right :wink:

I still don't really like that by this we change the default behaviour that you would expect from the API, but for now it seems like the most sensible way forward. At least until issue #29 is fully solved/implemented (which I don't have time for at this moment)...

Thanks!

dobegor:
@svanharmelen I am on business trip and away from my laptop for a day or two, so please, feel free to commit it yourself :)

Thanks for your work!

svanharmelen:
Will do... Thanks again and sorry for all the fuss...

PR #65, opened 2016-07-25
Use pointers in the *Options structs

### LINKED ISSUE #29: Structs should use pointers for fields

woopla:
Hi @svanharmelen,

Since I would very much like to be able to have 'false' being passed to some Gitlab requests, I've gone ahead and changed all the relevant *Options struct fields to pointers. Hopefully, this is enough to close #29.

Let me know if that works for you. I'm not completely satisfied with the new *LevelPtr funcs, but not adding them makes for very long lines in the code using it.

Cheers,

Clément

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -140,7 +140,7 @@ type Client struct {
 // support pagination.
 type ListOptions struct {
 	// For paginated result sets, page of results to retrieve.
-	Page *int `url:"page,omitempty" json:"page,omitempty"`
+	Page int `url:"page,omitempty" json:"page,omitempty"`
 
 	// For paginated result sets, the number of results to include per page.
 	PerPage int `url:"per_page,omitempty" json:"per_page,omitempty"`
--- a/gitlab.go
+++ b/gitlab.go
@@ -49,50 +49,50 @@ const (
 	oAuthToken
 )
 
-// AccessLevel represents a permission level within GitLab.
+// AccessLevelValue represents a permission level within GitLab.
 //
 // GitLab API docs: http://doc.gitlab.com/ce/permissions/permissions.html
-type AccessLevel int
+type AccessLevelValue int
 
 // List of available access levels
 //
 // GitLab API docs: http://doc.gitlab.com/ce/permissions/permissions.html
 const (
-	GuestPermissions     AccessLevel = 10
-	ReporterPermissions  AccessLevel = 20
-	DeveloperPermissions AccessLevel = 30
-	MasterPermissions    AccessLevel = 40
-	OwnerPermission      AccessLevel = 50
+	GuestPermissions     AccessLevelValue = 10
+	ReporterPermissions  AccessLevelValue = 20
+	DeveloperPermissions AccessLevelValue = 30
+	MasterPermissions    AccessLevelValue = 40
+	OwnerPermission      AccessLevelValue = 50
 )
 
-// NotificationLevel represents a notification level within Gitlab.
+// NotificationLevelValue represents a notification level within Gitlab.
 //
 // GitLab API docs: http://doc.gitlab.com/ce/api/
-type NotificationLevel int
+type NotificationLevelValue int
 
 // List of available notification levels
 //
 // GitLab API docs: http://doc.gitlab.com/ce/api/
 const (
-	DisabledNotifications NotificationLevel = iota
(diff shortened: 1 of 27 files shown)


gitlab.go
```diff
@@ -469,3 +469,19 @@ func String(v string) *string {
 	*p = v
 	return p
 }
+
+// AccessLevelPtr is a helper routine that allocates a new AccessLevel value
+// to store v and returns a pointer to it.
+func AccessLevelPtr(v AccessLevel) *AccessLevel {
```
svanharmelen: Maybe we can rename the type `AccessLevel` to `AccessLevelValue` and rename this func to just `AccessLevel`?

gitlab.go
```diff
@@ -469,3 +469,19 @@ func String(v string) *string {
...
+	p := new(AccessLevel)
+	*p = v
+	return p
+}
+
+// VisibilityLevelPtr is a helper routine that allocates a new AccessLevel value
+// to store v and returns a pointer to it.
+func VisibilityLevelPtr(v VisibilityLevel) *VisibilityLevel {
```
svanharmelen: And of course we can then do the same for `VisibilityLevel`...

gitlab.go
```diff
@@ -140,10 +140,10 @@ type Client struct {
 // support pagination.
 type ListOptions struct {
 	// For paginated result sets, page of results to retrieve.
-	Page int `url:"page,omitempty" json:"page,omitempty"`
+	Page *int `url:"page,omitempty" json:"page,omitempty"`
 
 	// For paginated result sets, the number of results to include per page.
-	PerPage int `url:"per_page,omitempty" json:"per_page,omitempty"`
+	PerPage *int `url:"per_page,omitempty" json:"per_page,omitempty"`
```
svanharmelen: We do not need to change these ones, so let's keep these as just int's

svanharmelen:
@woopla whoooot!! 🎉 

I noticed the examples still need to be updated to get Travis green and I had a few minor suggestions about `AccessLevelPtr` and `VisibilityLevelPtr`. But other then that I think this looks good and should indeed fix #29 

This will be a nasty update for users of this package, as it will be backwards incompatible, but I think it is needed to make sure we can support all use cases properly. So I will add some info about this to the README, hoping people understand why this is a needed step to take.

Big, big thanks for the efforts!!

woopla:
@svanharmelen  Yep I saw the Travis build fail and thought.... Ooooops, forgot something :smile_cat: 

I'll update the *Level stuff yes. It's going to be more changes to the API, but it makes the code slightly cleaner indeed. I'll also revert the PerPage stuff.

svanharmelen:
Cool :wink:

And I don't think it will change the API much... The constants remain the same, only their type changes. But I don't think people are using that type directly outside this package...

And well... If we break it, we might as well do it right now 😁

woopla:
Done. Hopefully, Travis won't complain this time around :stuck_out_tongue: 

I also changed `NotificationLevel` to `NotificationLevelValue` for consistency.

gitlab.go
```diff
@@ -143,7 +143,7 @@ type ListOptions struct {
 	Page *int `url:"page,omitempty" json:"page,omitempty"`
```
svanharmelen: Ah... You forgot this one...

svanharmelen:
Nice one... Consistency is a good thing 👍🏻

One tiny comment to go and we should be good! Will merge this tomorrow when behind a desk again.

Thanks again for your efforts! Really nice to have this in 😀

woopla:
Done, sorry for missing this. That's what happens when I pretend that I can multitask :stuck_out_tongue_closed_eyes:

svanharmelen:
Once more a big thank you :wink: Your commit is squashed and added to master (which is now tagged as being v0.2.0).

woopla:
Yay, awesome! Thanks a lot for the quick turn around time :fireworks:

PR #68, opened 2016-08-20
Removed log print from create issue

Joukehofman:
I noticed some output in my own script and traced it back to this log.Printf which isn't being used anywhere else in the project. (Great project btw, saved me a lot of time! :))


svanharmelen:
Thanks @Joukehofman! LGTM

PR #70, opened 2016-09-07
Add username parameter to user search

dobegor:
Check out http://docs.gitlab.com/ce/api/users.html

```
You can search for users by email or username with: /users?search=John

In addition, you can lookup users by username:

GET /users?username=:username
```


svanharmelen:
LGTM! Thanks for the PR @dobegor 😀

Issue #42, opened 2016-03-18
API /project/*/builds/*


mishak87:
Do you plan on adding `projects/<id>/{builds,repository}` API?
http://doc.gitlab.com/ce/api/builds.html

svanharmelen:
No immediate plans and personally quite busy at the moment... But of course a PR would be most welcome! :wink:

mishak87:
@svanharmelen PR it is then :-)

svanharmelen:
Fixed by PR #71

PR #71, opened 2016-09-08
Add support for /projects/*/builds/*

ArkBriar:
PR for issue #42

### DIFF
--- a/builds.go
+++ b/builds.go
@@ -150,7 +150,8 @@ type Build struct {
 
 // ListProjectBuilds gets a list of builds in a project.
 //
-// The scope of builds to show, one or array of: pending, running, failed, success, canceled; showing all builds if none provided.
+// The scope of builds to show, one or array of: pending, running,
+// failed, success, canceled; showing all builds if none provided.
 //
 // GitLab API docs:
 // http://docs.gitlab.com/ce/api/builds.html#list-project-builds
@@ -175,7 +176,8 @@ func (s *BuildsService) ListProjectBuilds(pid interface{}, opts *ListBuildsOptio
 	return builds, resp, err
 }
 
-// ListCommitBuilds gets a list of builds for specific commit in a project. If the commit SHA is not found, it will respond with 404.
+// ListCommitBuilds gets a list of builds for specific commit in a
+// project. If the commit SHA is not found, it will respond with 404.
 //
 // GitLab API docs:
 // http://docs.gitlab.com/ce/api/builds.html#list-commit-builds
@@ -250,7 +252,8 @@ func (s *BuildsService) GetBuildArtifacts(pid interface{}, buildID int) (io.Read
 	return artifactsBuf, resp, err
 }
 
-// DownloadArtifactsFile download the artifacts file from the given reference name and job provided the build finished successfully.
+// DownloadArtifactsFile download the artifacts file from the given
+// reference name and job provided the build finished successfully.
 //
 // GitLab API docs:
 // http://docs.gitlab.com/ce/api/builds.html#download-the-artifacts-file
@@ -350,7 +353,8 @@ func (s *BuildsService) RetryBuild(pid interface{}, buildID int) (*Build, *Respo
 	return build, resp, err
 }
 
-// EraseBuild erases a single build of a project, removes a build artifacts and a build trace.
+// EraseBuild erases a single build of a project, removes a build
+// artifacts and a build trace.
 //
 // GitLab API docs:
 // http://docs.gitlab.com/ce/api/builds.html#erase-a-build
@@ -375,7 +379,8 @@ func (s *BuildsService) EraseBuild(pid interface{}, buildID int) (*Build, *Respo
 	return build, resp, err
 }
 
-// KeepArtifacts prevents artifacts from being deleted when expiration is set.
+// KeepArtifacts prevents artifacts from being deleted when
+// expiration is set.
 //
 // GitLab API docs:
 // http://docs.gitlab.com/ce/api/builds.html#keep-artifacts
--- a/builds.go
+++ b/builds.go
@@ -24,65 +24,9 @@ import (
 	"time"
 )
 
-// BuildScope specifies five build status: pending, running, failed, success and canceled
-type BuildScope int
-
-// Build scopes: pending, running, failed, success, canceled
-const (
-	BuildPending BuildScope = 1 << iota
-	BuildRunning
-	BuildFailed
-	BuildSuccess
-	BuildCanceled
-)
-
-func buildScope(scope BuildScope) string {
-	switch scope {
-	case BuildPending:
-		return "pending"
-	case BuildRunning:
-		return "running"
-	case BuildFailed:
-		return "failed"
-	case BuildSuccess:
-		return "success"
-	case BuildCanceled:
-		return "canceled"
-	default:
-		return "null"
-	}
-}
-
 // ListBuildsOptions are options for two list apis
 type ListBuildsOptions struct {
-	Scope *[]string `url:"scope, omitempty", json:"scope, omitempty"`
-}
-
-// BuildListBuildsOptions builds ListBuildsOptions from scopes in the form of integer.
-//
-// For example, BuildListBuildsOptions(BuildPending | BuildRunning | BuildFailed) will build
-// an option with "pending", "running" and "failed" scopes.
-func BuildListBuildsOptions(scopes int) *ListBuildsOptions {
-	opts := &ListBuildsOptions{
-		Scope: &[]string{},
-	}
-
-	if scopes == 0 {
-		return nil
-	}
-
-	for i := BuildPending; i <= BuildCanceled; i = i << 1 {
-		if (scopes & int(i)) != 0 {
-			temp := append(*opts.Scope, buildScope(i))
-			opts.Scope = &temp
-		}
-	}
-
-	if len(*opts.Scope) == 0 {
-		return nil
-	}
-
-	return opts
+	Scope []string `url:"scope,omitempty", json:"scope,omitempty"`
 }
 
 // BuildsService handles communication with the ci builds related methods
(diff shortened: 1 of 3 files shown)


builds.go
```diff
@@ -0,0 +1,431 @@
...
+	case BuildCanceled:
+		return "canceled"
+	default:
+		return "null"
+	}
+}
+
+// ListBuildsOptions are options for two list apis
+type ListBuildsOptions struct {
+	Scope *[]string `url:"scope, omitempty", json:"scope, omitempty"`
```
svanharmelen: A slice is already a referenced type (with its zero value being `nil`), so please do not make this a pointer. 

Additionally please leave out the spaces between labels, so `"scope, omitempty"` should be `"scope,omitempty"`. Having the space here is of course not wrong, but I like to keep things consistent throughout the package :wink:

builds.go
```diff
@@ -0,0 +1,431 @@
...
+// GitLab API docs: http://docs.gitlab.com/ce/api/builds.html
+type BuildsService struct {
+	client *Client
+}
+
+// Build represents a ci build.
+//
+// GitLab API docs: http://docs.gitlab.com/ce/api/builds.html
+type Build struct {
+	Commit *struct {
```
svanharmelen: I think we should use the existing `Commit` struct here: https://github.com/xanzy/go-gitlab/blob/master/commits.go#L36

builds.go
```diff
@@ -0,0 +1,431 @@
...
+		Description string `json:"description"`
+		Active      bool   `json:"active"`
+		IsShared    bool   `json:"is_shared"`
+		Name        string `json:"Name"`
+	} `json:"runner, omitempty"`
+	Stage     string     `json:"stage"`
+	StartedAt *time.Time `json:"started_at"`
+	Status    string     `json:"status"`
+	Tag       bool       `json:"tag"`
+	User      *struct {
```
svanharmelen: I think we should use the existing `User` struct here: https://github.com/xanzy/go-gitlab/blob/master/users.go#L36

builds.go
```diff
@@ -0,0 +1,431 @@
...
+	Commit *struct {
+		AuthorEmail string     `json:"author_email"`
+		AuthorName  string     `json:"author_name"`
+		CreatedAt   *time.Time `json:"created_at"`
+		ID          string     `json:"id"`
+		Message     string     `json:"message"`
+		ShortID     string     `json:"short_id"`
+		Title       string     `json:"title"`
+	} `json:"commit"`
+	Coverage      *string    `json:"coverage, omitempty"`
```
svanharmelen: Please do not use a pointer here

builds.go
```diff
@@ -0,0 +1,431 @@
...
+	CreatedAt     *time.Time `json:"created_at"`
+	ArtifactsFile *struct {
```
svanharmelen: Please do not use a pointer here

builds.go
```diff
@@ -0,0 +1,431 @@
...
+		Filename string `json:"filename"`
+		Size     int    `json:"size"`
+	} `json:"artifacts_file"`
+	FinishedAt *time.Time `json:"finished_at"`
+	ID         int        `json:"id"`
+	Name       string     `json:"name"`
+	Ref        string     `json:"ref"`
+	Runner     *struct {
```
svanharmelen: Please do not use a pointer here

builds.go
```diff
@@ -0,0 +1,431 @@
...
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/builds", url.QueryEscape(project))
+
+	req, err := s.client.NewRequest("GET", u, opts)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	builds := []Build{}
```
svanharmelen: `var builds []Build` would be nicer here IMHO (goes for the other locations in this file as well of course)

builds.go
```diff
@@ -0,0 +1,431 @@
...
+//
+// GitLab API docs:
+// http://docs.gitlab.com/ce/api/builds.html#get-a-single-build
+func (s *BuildsService) GetSingleBuild(pid interface{}, buildID int) (*Build, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	u := fmt.Sprintf("projects/%s/builds/%d", project, buildID)
```
svanharmelen: _nitting_ There is a whiteline above this line, were in all other places the whiteline is below this line...

builds.go
```diff
@@ -0,0 +1,431 @@
...
+	req, err := s.client.NewRequest("GET", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	build := &Build{}
```
svanharmelen: In the rest of this package this is written as `build := new(Build)` (again for consistency)

builds.go
```diff
@@ -0,0 +1,431 @@
...
+//
+// GitLab API docs:
+// http://docs.gitlab.com/ce/api/builds.html#get-build-artifacts
+func (s *BuildsService) GetBuildArtifacts(pid interface{}, buildID int) (io.Reader, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	u := fmt.Sprintf("projects/%s/builds/%d/artifacts", project, buildID)
```
svanharmelen: Same here about the whiteline...

gitlab.go
```diff
@@ -134,6 +134,7 @@ type Client struct {
 	Settings        *SettingsService
 	SystemHooks     *SystemHooksService
 	Users           *UsersService
+	Builds          *BuildsService
```
svanharmelen: _nitting_ Could you please insert the new `Builds` service following alphabetic ordering?

gitlab.go
```diff
@@ -190,6 +191,7 @@ func newClient(httpClient *http.Client, tokenType tokenType, token string) *Clie
 	c.Settings = &SettingsService{client: c}
 	c.SystemHooks = &SystemHooksService{client: c}
 	c.Users = &UsersService{client: c}
+	c.Builds = &BuildsService{client: c}
```
svanharmelen: _nitting_ Could you please insert the new `c.Builds` service following alphabetic ordering?

builds.go
```diff
@@ -0,0 +1,431 @@
...
+			opts.Scope = &temp
+		}
+	}
+
+	if len(*opts.Scope) == 0 {
+		return nil
+	}
+
+	return opts
+}
```
svanharmelen: I must admit I don't really like this approach (sorry)... I would leave out the `buildScope ` and `BuildListBuildsOptions ` functions all together and change `ListBuildsOptions` to:

```go
type ListBuildsOptions struct {
  Scope []BuildState `url:"scope,omitempty", json:"scope,omitempty"`
}
```

As `BuildState` is already an existing constant in this package: https://github.com/xanzy/go-gitlab/blob/4f1f8d2b5b095a5e718e3658602598890f4ec746/commits.go#L319

So calling the `ListProjectBuilds` would then become something like:

```go
opts := &ListBuildsOptions{
  Scope: []BuildState{gitlab.Running, gitlab.Pending},
}

builds, resp, err := ListProjectBuilds("svanharmelen/project", opts)
<snip>
```

Which looks pretty user friendly and much cleaner to me...

svanharmelen:
@ArkBriar thanks you so much for this PR!

I left quite some comments, but most of them are very small nits to ensure consistency throughout this package and should just take a minute to update.

Next to that there is one comment (regarding the build scopes) that is about the chosen approach which I would like to change a little to make it more straightforward and (IMO) more logical. But again that should easy enough to update.

But overall it looks great so again thanks for adding this API 😀

ArkBriar:
Thanks for your comments, that makes me more clear about how to write effective go codes.
I have modified builds.go according to your advices and performed an `gofmt` on this package.

The `coverage` field is always null when i test with "gitlab.com" and project "gitlab-org/gitlab-ce".

builds.go
```diff
@@ -0,0 +1,359 @@
...
+	"bytes"
+	"fmt"
+	"io"
+	"net/url"
+	"time"
+)
+
+// ListBuildsOptions are options for two list apis
+type ListBuildsOptions struct {
+	Scope []string `url:"scope,omitempty", json:"scope,omitempty"`
```
svanharmelen: This should be `Scope []BuildState ...`

builds.go
```diff
@@ -0,0 +1,359 @@
...
+//
+// GitLab API docs: http://docs.gitlab.com/ce/api/builds.html
+type BuildsService struct {
+	client *Client
+}
+
+// Runner represents a ci runner
+//
+// GitLab API docs: http://docs.gitlab.com/ce/api/builds.html
+type Runner struct {
```
svanharmelen: I don't think we should define `Runner` here, as `Runner` has it's own API: https://docs.gitlab.com/ce/api/runners.html

So until we have implemented that API and it's type, I prefer to use this type as an anonymous embedded struct (like you had before) instead.

builds.go
```diff
@@ -0,0 +1,359 @@
...
+	Description string `json:"description"`
+	Active      bool   `json:"active"`
+	IsShared    bool   `json:"is_shared"`
+	Name        string `json:"name"`
+}
+
+// ArtifactsFile represents an artifacts file (.gitlab.yml)
+//
+// GitLab API docs: http://docs.gitlab.com/ce/api/builds.html
+type ArtifactsFile struct {
```
svanharmelen: As this is not a known entity I again prefer to also just keep this struct as an embedded struct in the Build struct.

builds.go
```diff
@@ -0,0 +1,359 @@
...
+	Filename string `json:"filename"`
+	Size     int    `json:"size"`
+}
+
+// Build represents a ci build.
+//
+// GitLab API docs: http://docs.gitlab.com/ce/api/builds.html
+type Build struct {
+	Commit        Commit        `json:"commit"`
```
svanharmelen: Please use a pointer here...

Let me explain a little why this should have a pointer and the previous ones don't. It is not possible to pass around an anonymous struct to another function or variable, so there is no use in it being a pointer. But defined types like `Commit` and `User` can be passed around. So they being a pointer, does make sense :wink:

builds.go
```diff
@@ -0,0 +1,359 @@
...
+	FinishedAt    *time.Time    `json:"finished_at"`
+	ID            int           `json:"id"`
+	Name          string        `json:"name"`
+	Ref           string        `json:"ref"`
+	Runner        Runner        `json:"runner"`
+	Stage         string        `json:"stage"`
+	StartedAt     *time.Time    `json:"started_at"`
+	Status        string        `json:"status"`
+	Tag           bool          `json:"tag"`
+	User          User          `json:"user"`
```
svanharmelen: Same goes for this one, should be a pointer.

svanharmelen:
@ArkBriar thanks for the updates, just a few additional comments left...

As for the `coverage` , I think it's going to expect a struct here as well. But I also cannot find any detail on that one, so I suggest to just delete the field for now.

Thx!

svanharmelen:
LGTM! Again thanks for the PR @ArkBriar!

PR #75, opened 2016-10-05
FIX Bug: Api projects/:id

wolfitem:
Add ops is equal to null judgment


svanharmelen:
Thanks @wolfitem! LGTM!

PR #79, opened 2016-10-08
closes #77 - add `web_url` field to `Issue`

### LINKED ISSUE #77: Add `web_url` field to `Issue`
The [`Issue`](https://godoc.org/github.com/xanzy/go-gitlab#Issue) struct (and the parsing stuff etc.) does not have the `web_url` field.

Would you be willing to accept a PR in case, I make the necessary changes to add the same?

the-shank:


issues.go
```diff
@@ -70,6 +70,7 @@ type Issue struct {
 	State     string     `json:"state"`
 	UpdatedAt *time.Time `json:"updated_at"`
 	CreatedAt *time.Time `json:"created_at"`
+	WebUrl    string     `json:"web_url"`
```
svanharmelen: This should be `WebURL` following the Go style guides...

svanharmelen:
@the-shank thanks for your PR! While we are adding missing fields to the `Issue` struct, could you maybe also add the other fields that were added recently:

```
 "subscribed" : false,
 "user_notes_count": 1,
 "due_date": "2016-07-22",
 "confidential": false
```

Would be great as we then are fully up-to-date again with this struct :wink:

the-shank:
Sure. I will update the PR.

the-shank:
@svanharmelen quick question. Should `due_date` be a `string` or `*time.Time`. As I see, the `created_at`, `updated_at` fields are of type `*time.Time`. But the `due_date` field (for example in `Milestone` struct) is of type `string`. What's the reason behind this?

svanharmelen:
I guess in this case it should be a `string`.

The reason that this is not consistent, is that the Gitlab API is not consistent. In most cases it uses date format RFC3339, but in some places it doesn't (e.g. https://gitlab.com/gitlab-org/gitlab-ce/issues/20881).

Looking at the output example, in this case it also doesn't use the RFC3339 format, so just use a `string` here.

issues.go
the-shank: sorry, will update accordingly!

the-shank:
@svanharmelen cool, thanks!

svanharmelen:
Thanks for your contribution @the-shank! LGTM!

Issue #81, opened 2016-10-13
Support for Accept Merge Request parameters


jereksel:
https://docs.gitlab.com/ce/api/merge_requests.html#accept-mr

svanharmelen:
Fixed by both PR #82 and PR #104

PR #82, opened 2016-10-13
Support for AcceptMergeRequest parameters

jereksel:
Implements #81

### DIFF
--- a/merge_requests.go
+++ b/merge_requests.go
@@ -18,6 +18,7 @@ package gitlab
 
 import (
 	"fmt"
+	"github.com/google/go-querystring/query"
 	"net/url"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


merge_requests.go
```diff
@@ -286,12 +299,21 @@ func (s *MergeRequestsService) UpdateMergeRequest(
...
-	u := fmt.Sprintf("projects/%s/merge_request/%d/merge", url.QueryEscape(project), mergeRequest)
+
+	v, err := query.Values(opt)
+
+	if err != nil {
+		return nil, nil, err
+	}
+
+	//For some reason putting struct to body doesn't work
+	u := fmt.Sprintf("projects/%s/merge_request/%d/merge?%s", url.QueryEscape(project), mergeRequest, v.Encode())
```
svanharmelen: Could you explain what is not working? As this is done for all other calls successfully, I wonder if it's a bug in the Gitlab API? And if so, maybe it's already fixed in the latest version(s)?

svanharmelen:
@jereksel thanks for the PR! Just have a question about the error you get, the code itself looks fine...

merge_requests.go
jereksel: I'm testing in on 8.11.6. Also I don't userstand why putting it in body would work in the first place - Gitlab docs say that parameters should be put into URL.

svanharmelen:
@jereksel I tested this in a few different ways just now, and I do not get any error when using the logic using in the rest of this package (so putting the options in the body instead of the URL). So I created a new PR (#104) in which I cherry picked your commit and updated the code to make in inline with the rest.

Thanks for your PR and sorry for the long delay in looking into this one!!

PR #92, opened 2016-11-08
add EMail endpoints from gitlab api

ulrichSchreiner:
hi,

i've implemented some REST-endpoints for the management of user emails. The functions should be identical to the `SSH-Key` functions

### DIFF
--- a/users.go
+++ b/users.go
@@ -438,58 +438,58 @@ func (s *UsersService) UnblockUser(user int) error {
 	}
 }
 
-// EMail represents an EMail.
+// Email represents an Email.
 //
-// GitLab API docs: https://doc.gitlab.com//help/api/users.md#list-emails
-type EMail struct {
+// GitLab API docs: https://doc.gitlab.com/ce/api/users.html#list-emails
+type Email struct {
 	ID    int    `json:"id"`
-	EMail string `json:"email"`
+	Email string `json:"email"`
 }
 
-// ListEMails gets a list of currently authenticated user's EMails.
+// ListEmails gets a list of currently authenticated user's Emails.
 //
 // GitLab API docs: http://doc.gitlab.com/ce/api/users.html#list-emails
-func (s *UsersService) ListEMails() ([]*EMail, *Response, error) {
+func (s *UsersService) ListEmails() ([]*Email, *Response, error) {
 	req, err := s.client.NewRequest("GET", "user/emails", nil)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	var k []*EMail
-	resp, err := s.client.Do(req, &k)
+	var e []*Email
+	resp, err := s.client.Do(req, &e)
 	if err != nil {
 		return nil, resp, err
 	}
 
-	return k, resp, err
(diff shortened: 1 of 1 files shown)


users.go
```diff
@@ -437,3 +437,152 @@ func (s *UsersService) UnblockUser(user int) error {
 		return fmt.Errorf("Received unexpected result code: %d", resp.StatusCode)
 	}
 }
+
+// EMail represents an EMail.
```
svanharmelen: Could you rename all instances of `EMail` to `Email` please?

users.go
```diff
@@ -437,3 +437,152 @@ func (s *UsersService) UnblockUser(user int) error {
...
+//
+// GitLab API docs: https://doc.gitlab.com//help/api/users.md#list-emails
```
svanharmelen: This URL is not correct, should be: `http://doc.gitlab.com/ce/api/users.html#list-emails`

users.go
```diff
@@ -437,3 +437,152 @@ func (s *UsersService) UnblockUser(user int) error {
...
+// ListEMails gets a list of currently authenticated user's EMails.
+//
+// GitLab API docs: http://doc.gitlab.com/ce/api/users.html#list-emails
+func (s *UsersService) ListEMails() ([]*EMail, *Response, error) {
+	req, err := s.client.NewRequest("GET", "user/emails", nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var k []*EMail
```
svanharmelen: The `k` stood for `key` in the `SSH` calls, so it should be `e` here to keep things consistent (applies to all instances of course)

users.go
```diff
@@ -437,3 +437,152 @@ func (s *UsersService) UnblockUser(user int) error {
...
+
+	return k, resp, err
+}
+
+// ListEMailsForUser gets a list of a specified user's EMails. Available
+// only for admin
+//
+// GitLab API docs:
+// http://doc.gitlab.com/ce/api/users.html#list-emails-for-user
+func (s *UsersService) ListEMailsForUser(user int) ([]*EMail, *Response, error) {
```
svanharmelen: I would prefer `uid` over `user` (applies to all instances of course)

users.go
```diff
@@ -437,3 +437,152 @@ func (s *UsersService) UnblockUser(user int) error {
...
+	return e, resp, err
+}
+
+// DeleteEMail deletes email owned by currently authenticated user. This is an
+// idempotent function and calling it on a key that is already deleted or not
+// available results in 200 OK.
+//
+// GitLab API docs:
+// http://doc.gitlab.com/ce/api/users.html#delete-email-for-current-owner
+func (s *UsersService) DeleteEMail(kid int) (*Response, error) {
```
svanharmelen: `kid` should be `eid` to keep things consistent (applies to all instances of course)

ulrichSchreiner:
hi, 

thanx for the comments. i implemented them. hope it's ok now.

usc

svanharmelen:
@ulrichSchreiner looks great now, thanks!

PR #93, opened 2016-11-18
Added missing GetSingleTag method to TagService

warmans:


svanharmelen:
Thanks @warmans, LGTM!

PR #94, opened 2016-11-29
Add missing milestone parameter gitlab.Milestone.StartDate

bjorand:
Add the optional `start_date` parameter to Milestone.

https://docs.gitlab.com/ee/api/milestones.html#create-new-milestone

PR #95, opened 2016-11-29
Add Milestone.StartDate to Issue

bjorand:
Following #94, Milestone.StartDate was missing from Issue struct.

### DIFF
--- a/issues.go
+++ b/issues.go
@@ -46,6 +46,7 @@ type Issue struct {
 		ID          int        `json:"id"`
 		Title       string     `json:"title"`
 		Description string     `json:"description"`
+		StartDate   string     `json:"start_date"`
 		DueDate     string     `json:"due_date"`
 		State       string     `json:"state"`
 		UpdatedAt   *time.Time `json:"updated_at"`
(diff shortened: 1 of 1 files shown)


issues.go
```diff
@@ -46,6 +46,7 @@ type Issue struct {
 		ID          int        `json:"id"`
 		Title       string     `json:"title"`
 		Description string     `json:"description"`
+		StartDate   string     `json:"start_date"`
```
svanharmelen: I think it better if we delete this inline struct and just use the `Milestone` struct here?

```
<snip>
  Milestone *Milestone `json:"milestone"`
<snip>
```

issues.go
```diff
@@ -36,23 +36,14 @@ type IssuesService struct {
...
-		CreatedAt   *time.Time `json:"created_at"`
-	} `json:"milestone"`
-	Assignee struct {
+	ID          int       `json:"id"`
+	IID         int       `json:"iid"`
+	ProjectID   int       `json:"project_id"`
+	Title       string    `json:"title"`
+	Description string    `json:"description"`
+	Labels      []string  `json:"labels"`
+	Milestone   Milestone `json:"milestone"`
```
svanharmelen: Please make this a pointer to the Milestone struct (`*Milestone`), after that I'll merge it right away... Thanks!

svanharmelen:
Thanks again for the PR!

Issue #98, opened 2016-11-30
PipelineEvent json unmarshal error


mustafayildirim:
When parsing gitlab pipeline event hook giving pipeline json unmarshal error. [https://docs.gitlab.com/ce/web_hooks/web_hooks.html#pipeline-events](url)

svanharmelen:
Fixed by PR #99

PR #99, opened 2016-11-30
PipelineEvent type changed

mustafayildirim:
#98 fixing


mustafayildirim:
#100 added

events.go
```diff
@@ -312,8 +314,98 @@ type PipelineEvent struct {
...
+		FinishedAt  string `json:"finished_at"`
+	} `json:"commit"`
+	Repository struct {
+		Name            string `json:"name"`
+		GitSSHURL       string `json:"git_ssh_url"`
+		Description     string `json:"description"`
+		Homepage        string `json:"homepage"`
+		GitHTTPURL      string `json:"git_http_url"`
+		VisibilityLevel int    `json:"visibility_level"`
+	} `json:"repository"`
```
svanharmelen: Same goes for this event. Please use the existing structs, unless there is a difference that requires a separate struct to be defined.

mustafayildirim:
You're right, I missed the point.
I have changed. 
Thanks

svanharmelen:
Smart move, adding test for the events 👍🏻

Noticed you reverted the refactoring again. Any reason for that?

Let me know when your done making changes so I can review it again.

Thanks so far!

mustafayildirim:
When refactored the code I could not unmarshal, because they have some different properties.
You can review it.
I finished now.
After a while I may add some test to events.

svanharmelen:
@mustafayildirim I don't think this is a good approach. Looking at the Gitlab API there is no specific type/entity for a `GroupMergeEvent`, there only is a `MergeEvent`.

The better solution is to make the `MergeEvent` type work with both responses. I tested this locally, and if I update the current `MergeEvent` so it looks like the `GroupMergeEvent` you added, it works with both responses.

So please undo the `GroupMergeEvent` related changes and update the `MergeEvent` struct and tests instead to fix this new issue.

Thanks for your work and making the event work better!

mustafayildirim:
You are right. 
I removed GroupMergeEvent and changed MergeEvent for Group Webhooks.
Thanks for guidance.

svanharmelen:
LGTM! Thanks again @mustafayildirim!

PR #112, opened 2016-12-28
add support for project variables

david-ds:
Add support to list / get / create / update or delete build variables for a given project.

Gitlab API doc : https://docs.gitlab.com/ce/api/build_variables.html


svanharmelen:
Thanks for the PR @david-ds! 👍

PR #115, opened 2017-01-19
add support for paging to ListGroupMembers

peterm0x:


svanharmelen:
Thanks for this @peterm0x! I do have one thought about this... To keep things consistent, I wonder if it would be better to create a `ListGroupMembersOptions` struct like this:

```go
type ListGroupMembersOptions {
    ListOptions
}
```

And add that one to the `ListGroupMembers` call. I know at first hand this may seem a little cumbersome, but this way using/setting `ListOptions` remains fully consistent within the package.

So unless you have a good reason not to do this, I would prefer this approach. But please let me know your thoughts! Thx...

peterm0x:
I have initially made just that, but then replaced it `ListOptions` since there were no additional options. Reverted.

svanharmelen:
@peterm0x yeah it's definitely debatable, but I'm just a sucker for consistency I guess :wink: Thanks again!

PR #119, opened 2017-02-02
Add Time Tracking API

hoop33:
GitLab recently released their Time Tracking API. It's documented on the issues page (https://docs.gitlab.com/ce/api/issues.html), but they're separate calls and responses so I kept it separate. I tried to follow the project's conventions, but let me know if you have questions/concerns.


svanharmelen:
Thanks @hoop33! LGTM 👍

PR #125, opened 2017-02-13
Hotfix/84 api error reponse

### LINKED ISSUE #84: Incorrect handling of API errors
I discovered a possible bug today related to how the client handles validation errors returned from the API. According to the [docs](https://docs.gitlab.com/ce/api/README.html#data-validation-and-error-reporting), validation errors will be formatted as following in the response:

``` json
HTTP/1.1 400 Bad Request
Content-Type: application/json
{
    "message": {
        "bio": [
            "is too long (maximum is 255 characters)"
        ]
    }
}

{
    "message": {
        "<property-name>": [
            "<error-message>",
            "<error-message>",
            ...
        ],
        "<embed-entity>": {
            "<property-name>": [
                "<error-message>",
                "<error-message>",
                ...
            ],
        }
    }
}
```

In the code, an error response body is always unmarshalled into a `ErrorResponse` struct:

``` go
type ErrorResponse struct {
    Response *http.Response // HTTP response that caused this error
    Message  string         `json:"message"` // error message
    Errors   []Error        `json:"errors"`  // more detail on individual errors
}
```

This fails because the struct does not correctly reflect the response, resulting in an empty `ErrorResponse` being returned.

I have no experience with the older versions of the Gitlab API, so don't know if the error response format has changed, or whether this always has been a bug.

sindrepm:
This fixes #84


svanharmelen:
@sindrepm it seems you have mixed up some commits from different PR's here. And also noticed you added `.vscode/launch.json`. If you could sort the commits and delete `.vscode/launch.json` that would be great.

I'll have a closer look at the actual changes in a few days...

sindrepm:
Oops, sorry about that. I must have accidentally merged my other PR branch to my master. I will fix this tomorrow.

sindrepm:
@svanharmelen It should be fixed now. Sorry for messing up the commit history.

svanharmelen:
Fixed by #131 

Thanks for your groundwork and help!

PR #129, opened 2017-02-15
add paging for group-projects

ulrichSchreiner:
Hi,

the current function `ListGroupProjects` does not support paging. this is a problem, if a group contains many projects. this PR add paging functionality, so a caller can query **all** projects of a group.

usc


svanharmelen:
Thanks @ulrichSchreiner! LGTM!

PR #138, opened 2017-02-27
Commit: Fix parent_ids field (typo)

meeee:
This also renames the Go field, which may break things. Probably not a
lot though - ParentIds never had any content.

This typo probably is a result from invalid Gitlab API documentation
on tags, which also wrongly uses `parents_ids`.

I created a merge request there as well:

https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/9550


svanharmelen:
Thanks @meeee!

PR #141, opened 2017-03-15
Add statistics information and flag to query options

ulrichSchreiner:
Gitlab returns statistics for the `projects/all`, `projects/owned`, `groups` and `groups/owned` endpoints. Unfortunately the structure of such a statistics is not very well documented on their site.

This PR implements the statistics structure as it is implemented in current Gitlab code.

### DIFF
--- a/groups.go
+++ b/groups.go
@@ -33,12 +33,12 @@ type GroupsService struct {
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/groups.html
 type Group struct {
-	ID          int               `json:"id"`
-	Name        string            `json:"name"`
-	Path        string            `json:"path"`
-	Description string            `json:"description"`
-	Projects    *[]Project        `json:"projects,omitempty"`
-	Statistics  StorageStatistics `json:"statistics"`
+	ID          int                `json:"id"`
+	Name        string             `json:"name"`
+	Path        string             `json:"path"`
+	Description string             `json:"description"`
+	Projects    *[]Project         `json:"projects,omitempty"`
+	Statistics  *StorageStatistics `json:"statistics"`
 }
 
 // ListGroupsOptions represents the available ListGroups() options.
--- a/groups.go
+++ b/groups.go
@@ -37,7 +37,7 @@ type Group struct {
 	Name        string             `json:"name"`
 	Path        string             `json:"path"`
 	Description string             `json:"description"`
-	Projects    *[]Project         `json:"projects,omitempty"`
+	Projects    *[]Project         `json:"projects"`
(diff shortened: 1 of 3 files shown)


projects.go
```diff
@@ -108,6 +109,20 @@ type ProjectNamespace struct {
 	UpdatedAt   *time.Time `json:"updated_at"`
 }
 
+// StorageStatistics represents a statistics record for a group or project.
+type StorageStatistics struct {
+	StorageSize        uint64 `json:"storage_size"`
+	RepositorySize     uint64 `json:"repository_size"`
+	LfsObjectsSize     uint64 `json:"lfs_objects_size"`
+	BuildArtifactsSize uint64 `json:"build_artifacts_size"`
```
svanharmelen: You are right that this will never be a negative number, but a signed `int64` is more then big enough (it can represent sizes up to 8388608 Terabyte)

projects.go
```diff
@@ -108,6 +109,20 @@ type ProjectNamespace struct {
...
+}
+
+// ProjectStatistics represents a statistics record for a project.
+type ProjectStatistics struct {
+	StorageStatistics
+	CommitCount uint64 `json:"commit_count"`
```
svanharmelen: No reason to use this big a value for this field. Please just use an `int` here (represents a commit count up until 2.1 billion commits).

projects.go
```diff
@@ -77,6 +77,7 @@ type Project struct {
 		GroupName        string `json:"group_name"`
 		GroupAccessLevel int    `json:"group_access_level"`
 	} `json:"shared_with_groups"`
+	Statistics ProjectStatistics `json:"statistics"`
```
svanharmelen: Please make this a pointer: `Statistics *ProjectStatistics`

groups.go
```diff
@@ -33,19 +33,21 @@ type GroupsService struct {
...
-	Name        string     `json:"name"`
-	Path        string     `json:"path"`
-	Description string     `json:"description"`
-	Projects    *[]Project `json:"projects,omitempty"`
+	ID          int               `json:"id"`
+	Name        string            `json:"name"`
+	Path        string            `json:"path"`
+	Description string            `json:"description"`
+	Projects    *[]Project        `json:"projects,omitempty"`
+	Statistics  StorageStatistics `json:"statistics"`
```
svanharmelen: Please make this a pointer: `Statistics  *StorageStatistics`

And while you are editing this file, would you also want to fix the `Projects` line so it reads: 
`Projects []Project        'json:"projects"'`

Wonder how that one slipped through... Thx!

svanharmelen:
Great stuff 👍  Thx!

PR #144, opened 2017-03-22
Add a WithContext function

### LINKED ISSUE #143: Make all HTTP calls context-aware
With the `context` being made part of the standard library in 1.7, it should be possible to use a `context` for tracking requests, maybe at a client level and/or at a specific request level. Perhaps it could be an Option (`WithContext`)? The context should be used with the `http.Client` when performing requests, like so:

```go
resp, err := client.Do(req.WithContext(ctx))
```

https://github.com/golang/net/blob/master/context/ctxhttp/ctxhttp.go is a nice example of how this is done.

EDIT: I just had a look at the `gitlab.OptionFunc` and this seems almost serendipitously well suited for something like this:

```go
// WithContext runs the request with the provided context
func WithContext(ctx context.Context) OptionFunc {
	return func(req *http.Request) error {
			req = req.WithContext(ctx)
			return nil
		}
	}
}
```

johanbrandhorst:
Fixes #143 

This enables the use of requests with a specific context attached.

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -546,6 +546,15 @@ func WithSudo(uid interface{}) OptionFunc {
 	}
 }
 
+// WithContext runs the request with the provided context
+func WithContext(ctx context.Context) OptionFunc {
+	return func(req *http.Request) error {
+			req = req.WithContext(ctx)
+			return nil
+		}
+	}
+}
+
 // Bool is a helper routine that allocates a new bool value
 // to store v and returns a pointer to it.
 func Bool(v bool) *bool {
--- a/gitlab.go
+++ b/gitlab.go
@@ -18,6 +18,7 @@ package gitlab
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


johanbrandhorst:
What with this forcing users of `go-gitlab` to use 1.7, maybe something to consider? The standard library has dropped support for 1.6 since 1.8 was released.

johanbrandhorst:
Alternatively I could implement this using the `golang.org/x/net/context/` package if you prefer to keep backwards compatibility?

svanharmelen:
@johanbrandhorst seems like a nice addition to the package, thanks! I'll remove the Travis build against Go 1.6.x and add 1.8.x instead and trigger a rebuild by Travis.

I'm OK with using the `context` package from the standard lib, I'll make sure to give it a proper tag so people can choose to use this newer version.

svanharmelen:
@johanbrandhorst can you rebase this PR so Travis can run with the update `.travis.yml` file?

johanbrandhorst:
Rebased

svanharmelen:
Thanks!

PR #149, opened 2017-03-22
Fix the WithContext func

svanharmelen:
Please see https://golang.org/src/net/http/request.go?#L307 for more details about why this is needed.


svanharmelen:
@johanbrandhorst your test succeeded because the default return value of `req.Context()` is `context.Background()` if the context is `nil`.

The updated test fails when not dereferencing the pointers.

johanbrandhorst:
Ah, terribly sorry. Good spot!

svanharmelen:
No worries 👍

PR #150, opened 2017-03-22
Add Build Trigger endpoints

### LINKED ISSUE #148: Add support for project triggers
Doesn't look like [project pipeline triggers](https://docs.gitlab.com/ce/api/pipeline_triggers.html) are supported yet?

johanbrandhorst:
Fixes #148.

This implements build triggers for API v3, with the comments linking to relevant v3 API documents.

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -915,13 +915,13 @@ func (s *ProjectsService) ListBuildTriggers(pid interface{}, opt *ListBuildTrigg
 		return nil, nil, err
 	}
 
-	var ph []*BuildTrigger
-	resp, err := s.client.Do(req, &ph)
+	var bt []*BuildTrigger
+	resp, err := s.client.Do(req, &bt)
 	if err != nil {
 		return nil, resp, err
 	}
 
-	return ph, resp, err
+	return bt, resp, err
 }
 
 // GetBuildTrigger gets a specific build trigger for a project.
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
I've tested all 4 functions against my local gitlab v3 API and they all work. I thought about implementing tests but they don't really make sense without a real Gitlab server to test against.

projects.go
```diff
@@ -879,6 +879,120 @@ func (s *ProjectsService) DeleteProjectHook(pid interface{}, hook int, options .
...
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/triggers", url.QueryEscape(project))
+
+	req, err := s.client.NewRequest("GET", u, opt, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var ph []*BuildTrigger
```
svanharmelen: I would suggest using `bt` everywhere instead of `ph` which refers to `ProjectHook`

svanharmelen:
Just one small comment, looks good otherwise 👍

projects.go
johanbrandhorst: Can you tell where I got the inspiration from ;)?

svanharmelen:
Nice stuff @johanbrandhorst 😀  Your on a roll today 🎉 

Thanks!

PR #152, opened 2017-03-23
Change Build Triggers to PipelinTriggers.

johanbrandhorst:
This is in line with the updated v4 API. See https://docs.gitlab.com/ce/api/pipeline_triggers.html and https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/9713 for more information.

Also add StatusNoContent as successful status and update hardcoded status codes to `http.Status*` codes.


johanbrandhorst:
All changed functions tested against a v4 GitLab API.

svanharmelen:
@johanbrandhorst the code looks fine, but I would suggest creating a dedicated service for the PipelineTriggers (`PipelineTriggerService`) and put the code in it's own file `pipeline_triggers.go`. I know the name is a bit long, but it's consistent with the rest of the code.

FYI I try to follow (more or less) the segmentation Gitlab uses to describe different parts of the API: https://docs.gitlab.com/ce/api/README.html#resources

johanbrandhorst:
OK I will update it. I thought since it was under `/projects/` it would go under the `ProjectService`. No problem!

svanharmelen:
That also makes sense, I just choose to follow the docs and by that also limit the file sizes a bit.

Thanks!

svanharmelen:
Looks great! Thanks :+1:

PR #156, opened 2017-03-24
Update BranchesService Methods to Conform to v3 API

johanbrandhorst:
This change breaks backwards compatibility.

The `BranchesService` methods were almost all inconsistent with other Services and not supplying all available options. Also updated all documentation links.

The Branch struct now includes all properties exported by the v3 API.
`ListBranches` takes a new `ListBranchesOptions` struct for pagination.
Added the ability to configure the branch protection in line with API documentation.
Added a new `DeleteMergedBranches` which is part of the branches page.

### DIFF
--- a/branches.go
+++ b/branches.go
@@ -24,37 +24,48 @@ import (
 // BranchesService handles communication with the branch related methods
 // of the GitLab API.
 //
-// GitLab API docs: https://docs.gitlab.com/ce/api/branches.html
+// GitLab API docs: https://gitlab.com/gitlab-org/gitlab-ce/blob/8-16-stable/doc/api/branches.md
 type BranchesService struct {
 	client *Client
 }
 
 // Branch represents a GitLab branch.
 //
-// GitLab API docs: https://docs.gitlab.com/ce/api/branches.html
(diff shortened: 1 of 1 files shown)


svanharmelen:
Thanks @johanbrandhorst LGTM! :tada: 👍

johanbrandhorst:
I'm a bit taken aback - I guess you're not upset with breaking backwards compatibility at this point then?

svanharmelen:
Well... In general I most certainly are, but in this case I think it's OK to merge this one because the backwards incompatibility is limited to 2 specific API endpoints. So I hope and suspect the impact is very low and for those that are impacted the fix is very easy and fast.

Next to that I would like to work up to one more release in master which will receive tag v0.6.0 indicating the possible breaking changes. Additionally that release should come with some additional info in the README explaining the changes and the approach to start supporting the V4 API. From that point on I will only accept bugfixes for the V3 code, so changes that add value (like this one) should be added before that moment.

Finally I hope most people do vendoring by now, so if they stick to the latest tag they should be good and will not be impacted.

That all being said, I fully agree backwards compatibility should not be taken lightly...

PR #163, opened 2017-04-01
Return basic info about pipeline in GET /projects/:id/pipelines

ivaravko:
This is updated v4 API. See https://docs.gitlab.com/ce/api/pipelines.html#list-project-pipelines and https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/8875 for more information.

### DIFF
--- a/pipelines.go
+++ b/pipelines.go
@@ -58,24 +58,28 @@ type Pipeline struct {
 	Coverage    string     `json:"coverage"`
 }
 
-// RootPipeline represents a GitLab list project pipelines
+func (i Pipeline) String() string {
+	return Stringify(i)
+}
+
+// PipelineList represents a GitLab list project pipelines
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/pipelines.html#list-project-pipelines
-type RootPipeline struct {
+type PipelineList []struct {
 	ID     int    `json:"id"`
 	Status string `json:"status"`
 	Ref    string `json:"ref"`
 	Sha    string `json:"sha"`
 }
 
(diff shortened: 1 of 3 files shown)


pipelines.go
```diff
@@ -58,14 +58,24 @@ type Pipeline struct {
...
+// RootPipeline represents a GitLab list project pipelines
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/pipelines.html#list-project-pipelines
+type RootPipeline struct {
+	ID     int    `json:"id"`
+	Status string `json:"status"`
+	Ref    string `json:"ref"`
+	Sha    string `json:"sha"`
+}
+
```
svanharmelen: I'm not sure what to make of this. I see what you are trying to do and I agree it's maybe somewhat cleaner to have a smaller struct matching the field returned from a list call. But I don't think I like the suggested solution.

The name `RootPipeline` seems to imply something that doesn't exist (there is no such thing as a root of a pipeline). And also please do not put the new type in between the `Pipeline` type and it's `String` method. I like to keep those close together. Additionally please also add the `String` method for the new struct.

As for a possible solution I would like to suggest making a `PipelineList` type:

```go
type PipelineList []struct {
    ID     int    `json:"id"`
    Status string `json:"status"`
    Ref    string `json:"ref"`
    Sha    string `json:"sha"`
}
```

I think that would (both in term of it's name and it's structure) fit the context a bit better. Thoughts?

svanharmelen:
@ivaravko that looks much better (to me at least :wink:) Thanks!

PR #165, opened 2017-04-06
Adding Todos API resource

eguevara:
Adding GitLab Todo API resource to support ListTodos, MarkTodoRead and MarkTodosRead
https://docs.gitlab.com/ce/api/todos.html

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -186,6 +186,7 @@ type Client struct {
 	SystemHooks          *SystemHooksService
 	Tags                 *TagsService
 	TimeStats            *TimeStatsService
+	Todos                *TodosService
 	Users                *UsersService
 }
 
@@ -249,6 +250,7 @@ func newClient(httpClient *http.Client, tokenType tokenType, token string) *Clie
 	c.SystemHooks = &SystemHooksService{client: c}
 	c.Tags = &TagsService{client: c}
 	c.TimeStats = &TimeStatsService{client: c}
+	c.Todos = &TodosService{client: c}
 	c.Users = &UsersService{client: c}
 
 	return c
--- a/todos.go
+++ b/todos.go
@@ -1,0 +1,120 @@
+package gitlab
+
+import "time"
+import "fmt"
+
+// TodosService handles communication with the todos related methods of
+// the Gitlab API.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/todos.html
+type TodosService struct {
+	client *Client
+}
+
+// TodoAction represents the available actions that can be performed on a todo.
+// Todo represents a GitLab todo.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/todos.html
+type TodoAction string
+
+// The available todo actions.
+const (
+	TodoAssigned          TodoAction = "assigned"
+	TodoMentioned         TodoAction = "mentioned"
+	TodoBuildFailed       TodoAction = "build_failed"
+	TodoMarked            TodoAction = "marked"
+	TodoApprovalRequired  TodoAction = "approval_required"
+	TodoDirectlyAddressed TodoAction = "directly_addressed"
+)
+
+// Todo represents a GitLab todo.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/todos.html
+type Todo struct {
+	ID      int `json:"id"`
+	Project struct {
+		ID                int    `json:"id"`
+		HTTPURLToRepo     string `json:"http_url_to_repo"`
+		WebURL            string `json:"web_url"`
+		Name              string `json:"name"`
+		NameWithNamespace string `json:"name_with_namespace"`
+		Path              string `json:"path"`
+		PathWithNamespace string `json:"path_with_namespace"`
+	} `json:"project"`
+	Author struct {
+		ID        int    `json:"id"`
+		Name      string `json:"name"`
+		Username  string `json:"username"`
+		State     string `json:"state"`
+		AvatarURL string `json:"avatar_url"`
+		WebURL    string `json:"web_url"`
+	} `json:"author"`
+	ActionName TodoAction  `json:"action_name"`
+	TargetType string      `json:"target_type"`
+	Target     interface{} `json:"target"`
+	TargetURL  string      `json:"target_url"`
+	Body       string      `json:"body"`
+	State      string      `json:"state"`
+	CreatedAt  *time.Time  `json:"created_at"`
+}
+
+func (t Todo) String() string {
+	return Stringify(t)
+}
+
+// ListTodosOptions represents the available ListTodos() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/todos.html#get-a-list-of-todos
+type ListTodosOptions struct {
+	Action    *TodoAction `url:"action,omitempty" json:"action,omitempty"`
+	AuthorID  *int        `url:"author_id,omitempty" json:"author_id,omitempty"`
+	ProjectID *int        `url:"project_id,omitempty" json:"project_id,omitempty"`
+	State     *string     `url:"state,omitempty" json:"state,omitempty"`
+	Type      *string     `url:"type,omitempty" json:"type,omitempty"`
+}
+
+// ListTodos lists all todos created by authenticated user.
+// When no filter is applied, it returns all pending todos for the current user.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/todos.html#get-a-list-of-todos
+func (s *TodosService) ListTodos(opt *ListTodosOptions, options ...OptionFunc) ([]*Todo, *Response, error) {
+	req, err := s.client.NewRequest("GET", "todos", opt, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var t []*Todo
+	resp, err := s.client.Do(req, &t)
+	if err != nil {
+		return nil, resp, err
+	}
+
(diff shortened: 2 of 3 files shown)


todos.go
```diff
@@ -0,0 +1,109 @@
...
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/todos.html#get-a-list-of-todos
+type ListTodosOptions struct {
+	Action    string `url:"action,omitempty" json:"action,omitempty"`
+	AuthorID  int    `url:"author_id,omitempty" json:"author_id,omitempty"`
+	ProjectID int    `url:"project_id,omitempty" json:"project_id,omitempty"`
+	State     string `url:"state,omitempty" json:"state,omitempty"`
+	Type      string `url:"type,omitempty" json:"type,omitempty"`
+}
```
svanharmelen: These should all be pointer values. See all other option structs in this package, they all have pointer fields. This is needed to be able to make a distinction between a types zero value and no value so that the `omitempty` works as expected.

gitlab.go
```diff
@@ -192,6 +192,7 @@ type Client struct {
 	Tags                 *TagsService
 	TimeStats            *TimeStatsService
 	Users                *UsersService
+	Todos                *TodosService
```
svanharmelen: Could you please order these alphabetically? Thx!

gitlab.go
```diff
@@ -254,6 +255,7 @@ func newClient(httpClient *http.Client, tokenType tokenType, token string) *Clie
 	c.Tags = &TagsService{client: c}
 	c.TimeStats = &TimeStatsService{client: c}
 	c.Users = &UsersService{client: c}
+	c.Todos = &TodosService{client: c}
```
svanharmelen: Could you please order these alphabetically? Thx!

todos.go
```diff
@@ -0,0 +1,109 @@
+package gitlab
+
+import "time"
+import "fmt"
+
+// TodosService handles communication with the todos related methods of
+// the Gitlab API.
+//
+// GitLab API docs:
+// https://gitlab.com/gitlab-org/gitlab-ce/blob/8-16-stable/doc/api/todos.md
```
svanharmelen: I don't know if there are any differences, but it would be much appreciated if you could use the api V4 docs and target this PR against the `f-api-v4` branch instead.

todos.go
```diff
@@ -0,0 +1,109 @@
...
+		ID        int    `json:"id"`
+		Name      string `json:"name"`
+		Username  string `json:"username"`
+		State     string `json:"state"`
+		AvatarURL string `json:"avatar_url"`
+		WebURL    string `json:"web_url"`
+	} `json:"author"`
+	ActionName string      `json:"action_name"`
+	TargetType string      `json:"target_type"`
+	Target     interface{} `json:"target"`
```
svanharmelen: `Target` should not be an `interface{}`. Please create an inline struct with the correct fields instead.

todos.go
```diff
@@ -0,0 +1,109 @@
...
+	} `json:"project"`
+	Author struct {
+		ID        int    `json:"id"`
+		Name      string `json:"name"`
+		Username  string `json:"username"`
+		State     string `json:"state"`
+		AvatarURL string `json:"avatar_url"`
+		WebURL    string `json:"web_url"`
+	} `json:"author"`
+	ActionName string      `json:"action_name"`
```
svanharmelen: I think we should we make a specific type and some constants that represent all the available actions. What do you think?

todos.go
```diff
@@ -0,0 +1,109 @@
...
+// Todo represents a GitLab todo.
+//
+// GitLab API docs:
+// https://gitlab.com/gitlab-org/gitlab-ce/blob/8-16-stable/doc/api/todos.md
+type Todo struct {
+	ID      int `json:"id"`
+	Project struct {
+		ID                int    `json:"id"`
+		HTTPURLToRepo     string `json:"http_url_to_repo"`
+		WebURL            string `json:"web_url"`
```
svanharmelen: I cannot see both the `HTTPURLToRepo` and `WebURL` fields in the docs. Are you sure they are returned by the API?

todos.go
```diff
@@ -0,0 +1,109 @@
...
+
+func (t Todo) String() string {
+	return Stringify(t)
+}
+
+// ListTodosOptions represents the available ListTodos() options.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/todos.html#get-a-list-of-todos
+type ListTodosOptions struct {
```
svanharmelen: Ah, it seems you switched to the V4 API docs here. Please make sure the links are consistent and preferable take the V4 docs as leading and target this PR against the V4 branch.

todos.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// MarkTodoDone marks a single pending todo given by its ID for the current user as done.
+//
+// GitLab API docs:
+// https://gitlab.com/gitlab-org/gitlab-ce/blob/8-16-stable/doc/api/todos.md#mark-a-todo-as-done
+func (s *TodosService) MarkTodoDone(id int, options ...OptionFunc) (*Response, error) {
+	u := fmt.Sprintf("todos/%d", id)
+
+	req, err := s.client.NewRequest("DELETE", u, nil, options)
```
svanharmelen: This should be `POST` instead of `DELETE` and the given URL should be `todos/%d/mark_as_done` instead as just `todos/%d`.

_EDIT:_ Ah, I see... This is a difference between the V3 and V4 API's. Please consider making this PR target the V4 API.

todos.go
```diff
@@ -0,0 +1,109 @@
...
+
+	return s.client.Do(req, nil)
+}
+
+// MarkTodosDone marks all pending todos for the current user as done.
+//
+// GitLab API docs:
+// https://gitlab.com/gitlab-org/gitlab-ce/blob/8-16-stable/doc/api/todos.md#mark-all-todos-as-done
+func (s *TodosService) MarkTodosDone(options ...OptionFunc) (*Response, error) {
+	req, err := s.client.NewRequest("DELETE", "todos", nil, options)
```
svanharmelen: This should also be `POST` instead of `DELETE` and the given URL should be `todos/mark_as_done` instead as just `todos`.

_EDIT:_ Ah, I see... This is a difference between the V3 and V4 API's. Please consider making this PR target the V4 API.

todos.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	return s.client.Do(req, nil)
+}
+
+// MarkTodosDone marks all pending todos for the current user as done.
+//
+// GitLab API docs:
+// https://gitlab.com/gitlab-org/gitlab-ce/blob/8-16-stable/doc/api/todos.md#mark-all-todos-as-done
+func (s *TodosService) MarkTodosDone(options ...OptionFunc) (*Response, error) {
```
svanharmelen: I would name this function `MarkAllTodosAsDone` for clarity.

todos.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	return t, resp, err
+}
+
+// MarkTodoDone marks a single pending todo given by its ID for the current user as done.
+//
+// GitLab API docs:
+// https://gitlab.com/gitlab-org/gitlab-ce/blob/8-16-stable/doc/api/todos.md#mark-a-todo-as-done
+func (s *TodosService) MarkTodoDone(id int, options ...OptionFunc) (*Response, error) {
```
svanharmelen: I would call this function `MarkTodoAsDone` instead.

todos.go
eguevara: @svanharmelen yep. will update to target f-api-v4 target.

todos.go
eguevara: correct, the example response from the [docs](https://docs.gitlab.com/ce/api/todos.html#mark-a-todo-as-done) does not have them however, if I curl the v4 resource, I can see them listed.  

curl --header "PRIVATE-TOKEN: $PRIVATE_TOKEN" https://gitlab.com/api/v4/todos 

"project": {
      "id": 3048683,
      "http_url_to_repo": "https://gitlab.com/eguevara/helloworld.git",
      "web_url": "https://gitlab.com/eguevara/helloworld",
      "name": "helloworld",
      "name_with_namespace": "Erick Guevara / helloworld",
      "path": "helloworld",
      "path_with_namespace": "eguevara/helloworld"
    },

todos.go
eguevara: cool.. same can be said about the state field?  I'll list the available options based on the docs.

eguevara:
thanks @svanharmelen for your time reviewing.  I've amended all the suggested changes except the interface{} on the Target field.  

I'll take a look at that in bit and add as a separate commit to create the full struct for the Target field.

eguevara:
hey @svanharmelen, I made the suggested change to remove the interface{} on the Target field.   Since some of the fields are specific to the target type, i made some as pointers with omitempty so that they can be omitted?? Let me know if this works or if there is a better way to handle this case.

Thanks again.

svanharmelen:
Hi @eguevara! It's starting to look very nice 👍 

As for the fields that are specific to a certain todo type, please do not make them pointers and do not add the `omitempty` tag. I would say keep the comment saying the fields are only used for specific types, but just thread the fields like all other fields.

Once that is done, we should be good to merge this one. Thx!

eguevara:
Thanks @svanharmelen.  I amended the comit to remove the pointers.

svanharmelen:
Thanks for your efforts @eguevara! LGTM 😀 🎉

PR #166, opened 2017-04-13
Changing Builds service to Jobs service (v4)

bgetsug:
This change will replace the Builds service with the Jobs service. The `builds` endpoints are no longer available in GitLab 9.


svanharmelen:
Thanks @bgetsug! LGTM 🎉

PR #167, opened 2017-04-13
Fix groupvisibility and add sharewithgroup component 

nilsjung:
I fixed the api call for group visibility. it has not the same scheme like the project visibility in version 9. furthermore I added the project share with groups option.

### DIFF
--- a/groups.go
+++ b/groups.go
@@ -21,22 +21,6 @@ import (
 	"time"
 )
 
-// defines visibility level for groups
-type GroupVisibilityLevelValue string
-
-const (
-	PrivateGroupVisibility  GroupVisibilityLevelValue = "private"
-	PublicGroupVisibility   GroupVisibilityLevelValue = "public"
-	InternalGroupVisibility GroupVisibilityLevelValue = "internal"
-)
-
-// GroupVisibilityLevel as helper for setting visibility in options, same as like VisibilityLevel
-func GroupVisibilityLevel(v GroupVisibilityLevelValue) *GroupVisibilityLevelValue {
-	p := new(GroupVisibilityLevelValue)
-	*p = v
-	return p
-}
-
 // GroupsService handles communication with the group related methods of
 // the GitLab API.
 //
@@ -113,10 +97,10 @@ func (s *GroupsService) GetGroup(gid interface{}, options ...OptionFunc) (*Group
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/groups.html#new-group
 type CreateGroupOptions struct {
-	Name                 *string                    `url:"name,omitempty" json:"name,omitempty"`
-	Path                 *string                    `url:"path,omitempty" json:"path,omitempty"`
-	Description          *string                    `url:"description,omitempty" json:"description,omitempty"`
-	GroupVisibilityLevel *GroupVisibilityLevelValue `url:"visibility" json:"visibility,omitempty"`
+	Name            *string               `url:"name,omitempty" json:"name,omitempty"`
+	Path            *string               `url:"path,omitempty" json:"path,omitempty"`
+	Description     *string               `url:"description,omitempty" json:"description,omitempty"`
+	VisibilityLevel *VisibilityLevelValue `url:"visibility" json:"visibility,omitempty"`
 }
 
 // CreateGroup creates a new project group. Available only for users who can
--- a/groups.go
+++ b/groups.go
@@ -21,6 +21,22 @@ import (
 	"time"
 )
 
+// defines visibility level for groups
+type GroupVisibilityLevelValue string
+
+const (
+	PrivateGroupVisibility  GroupVisibilityLevelValue = "private"
+	PublicGroupVisibility   GroupVisibilityLevelValue = "public"
+	InternalGroupVisibility GroupVisibilityLevelValue = "internal"
+)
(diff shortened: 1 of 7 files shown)


groups.go
```diff
@@ -97,10 +113,10 @@ func (s *GroupsService) GetGroup(gid interface{}, options ...OptionFunc) (*Group
...
 // GitLab API docs: https://docs.gitlab.com/ce/api/groups.html#new-group
 type CreateGroupOptions struct {
-	Name            *string               `url:"name,omitempty" json:"name,omitempty"`
-	Path            *string               `url:"path,omitempty" json:"path,omitempty"`
-	Description     *string               `url:"description,omitempty" json:"description,omitempty"`
-	VisibilityLevel *VisibilityLevelValue `url:"visibility_level" json:"visibility_level,omitempty"`
+	Name                 *string                    `url:"name,omitempty" json:"name,omitempty"`
+	Path                 *string                    `url:"path,omitempty" json:"path,omitempty"`
+	Description          *string                    `url:"description,omitempty" json:"description,omitempty"`
+	GroupVisibilityLevel *GroupVisibilityLevelValue `url:"visibility" json:"visibility,omitempty"`
```
svanharmelen: @nilsjung all the changes in this file are not needed. Instead the currently existing `VisibilityLevelValue` type and constants should be updates to these new string values. See this item from the V3 to V4 docs:

image

projects.go
```diff
@@ -367,6 +367,27 @@ func (s *ProjectsService) CreateProject(opt *CreateProjectOptions, options ...Op
 	return p, resp, err
 }
 
+// ShareWithGroupOptions represents options to share project with groups
+//
+type ShareWithGroupOptions struct {
```
svanharmelen: Please add a URL to the docs here...

projects.go
```diff
@@ -367,6 +367,27 @@ func (s *ProjectsService) CreateProject(opt *CreateProjectOptions, options ...Op
...
+	ID          *int              `url:"id" json:"id"`
+	GroupID     *int              `url:"group_id" json:"group_id"`
+	GroupAccess *AccessLevelValue `url:"group_access" json:"group_access"`
+	ExpiresAt   *string           `url:"expires_at" json:"expires_at"`
+}
+
+func (s *ProjectsService) ShareProjectWithGroup(opt *ShareWithGroupOptions, options ...OptionFunc) (*Response, error) {
```
svanharmelen: Please add the few lines of comments all functions have in this package.

projects.go
```diff
@@ -367,6 +367,27 @@ func (s *ProjectsService) CreateProject(opt *CreateProjectOptions, options ...Op
...
+	u := fmt.Sprintf("projects/%d/share", *opt.ID)
+
+	req, err := s.client.NewRequest("POST", u, opt, options)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := s.client.Do(req, nil)
+	return resp, err
```
svanharmelen: Please do just `return s.client.Do(req, nil)` instead of these 2 lines.

projects.go
```diff
@@ -367,6 +367,27 @@ func (s *ProjectsService) CreateProject(opt *CreateProjectOptions, options ...Op
 	return p, resp, err
 }
 
+// ShareWithGroupOptions represents options to share project with groups
+//
+type ShareWithGroupOptions struct {
+	ID          *int              `url:"id" json:"id"`
```
svanharmelen: `ID` is not an option, so it should not be in the options struct, but in the function signature.

projects.go
```diff
@@ -367,6 +367,27 @@ func (s *ProjectsService) CreateProject(opt *CreateProjectOptions, options ...Op
...
+	GroupID     *int              `url:"group_id" json:"group_id"`
+	GroupAccess *AccessLevelValue `url:"group_access" json:"group_access"`
+	ExpiresAt   *string           `url:"expires_at" json:"expires_at"`
+}
+
+func (s *ProjectsService) ShareProjectWithGroup(opt *ShareWithGroupOptions, options ...OptionFunc) (*Response, error) {
+	u := fmt.Sprintf("projects/%d/share", *opt.ID)
```
svanharmelen: Don't use the `ID` from the options struct, but please put in the signature. And then before using it pass it through the `parseID` func so it becomes something like this:

```
	project, err := parseID(pid)
	if err != nil {
		return nil, err
	}
	u := fmt.Sprintf("projects/%s/share", url.QueryEscape(project))
```

projects.go
```diff
@@ -82,20 +82,20 @@ type Project struct {
...
-	SSHURL            string `json:"ssh_url"`
-	HTTPURL           string `json:"http_url"`
+	Name              string                `json:"name"`
+	Description       string                `json:"description"`
+	WebURL            string                `json:"web_url"`
+	AvatarURL         string                `json:"avatar_url"`
+	GitSSHURL         string                `json:"git_ssh_url"`
+	GitHTTPURL        string                `json:"git_http_url"`
+	Namespace         string                `json:"namespace"`
+	VisibilityLevel   *VisibilityLevelValue `json:"visibility"`
```
svanharmelen: This shouldn't be a pointer...

svanharmelen:
Thanks @nilsjung! Looks great now 🎉

Issue #151, opened 2017-03-22
Support the v4 API


johanbrandhorst:
Lets track this in an issue. The GitLab API v3 is officially deprecated and will be supported until at least 9.3. Given the current release schedule (once per month), that means it will be unsupported 3 months from now.

I would like to be able to start consuming v4 functionality so I might start work on a v4 branch.

I think all current documentation links need to be updated to the v3 documentation as currently the links all point to the v4 documentation which causes confusion. Separately some effort should be made to start supporting v4 specific functionality (probably in a feature branch). Once both those things are completed, v3 could be spun out to a separate branch and v4 can be merged to master.

Below are the changes made between V3 and V4. A tickbox indicates that work has been completed on that change.

### Completed
- [x] Rename Build Triggers to be Pipeline Triggers API [!9713](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/9713). Fixed in #152.
  - `POST /projects/:id/trigger/builds` to `POST /projects/:id/trigger/pipeline`
  - Require description when creating a new trigger `POST /projects/:id/triggers`
- [x] Removed `GET /projects/:search` (use: `GET /projects?search=x`) [!8877](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/8877). Fixed in #161.
- [x] `iid` filter has been removed from `GET /projects/:id/issues` [!8967](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/8967). Fixed in #161.
- [x] `GET /projects/:id/merge_requests?iid[]=x&iid[]=y` array filter has been renamed to `iids` [!8793](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/8793). Fixed in #161.
- [x] Endpoints under `GET /projects/merge_request/:id` have been removed (use: `GET /projects/merge_requests/:id`) [!8793](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/8793). Fixed in #161.
- [x] Project snippets do not return deprecated field `expires_at` [!8723](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/8723). Fixed in #161.
- [x] Endpoints under `GET /projects/:id/keys` have been removed (use `GET /projects/:id/deploy_keys`) [!8716](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/8716). Fixed in #161.
- [x] Update v3 documentation to point to archived v3 docs. Fixed in #160.
[...]

PR #170, opened 2017-05-11
Refactor ListProject calls to support the v4 API

CHERTS:
Hi

I fixed GitLab APIv4 support

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -33,8 +33,8 @@ import (
 )
 
 const (
-	libraryVersion = "0.1.1"
-	defaultBaseURL = "https://gitlab.com/api/v3/"
+	libraryVersion = "0.1.2"
+	defaultBaseURL = "https://gitlab.com/api/v4/"
 	userAgent      = "go-gitlab/" + libraryVersion
 )
 
--- a/gitlab.go
+++ b/gitlab.go
@@ -34,7 +34,7 @@ import (
 
 const (
 	libraryVersion = "0.1.2"
-	defaultBaseURL = "https://gitlab.com/api/v4/"
+	defaultBaseURL = "https://gitlab.com/api/v3/"
 	userAgent      = "go-gitlab/" + libraryVersion
 )
 
--- a/gitlab.go
+++ b/gitlab.go
@@ -33,7 +33,7 @@ import (
 )
 
 const (
-	libraryVersion = "0.1.2"
+	libraryVersion = "0.2.0"
 	defaultBaseURL = "https://gitlab.com/api/v3/"
 	userAgent      = "go-gitlab/" + libraryVersion
 )
--- a/projects.go
+++ b/projects.go
@@ -168,6 +168,7 @@ type ListProjectsOptions struct {
 	Search     *string               `url:"search,omitempty" json:"search,omitempty"`
 	Simple     *bool                 `url:"simple,omitempty" json:"simple,omitempty"`
 	Visibility *VisibilityLevelValue `url:"visibility,omitempty" json:"visibility,omitempty"`
+	Statistics *bool                 `url:"statistics,omitempty" json:"statistics,omitempty"`
 	Owned      *bool                 `url:"owned,omitempty" json:"owned,omitempty"`
 	Membership *bool                 `url:"membership,omitempty" json:"membership,omitempty"`
 	Starred    *bool                 `url:"starred,omitempty" json:"starred,omitempty"`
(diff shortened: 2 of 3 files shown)


svanharmelen:
@CHERTS besides that the Travis checks are failing, this PR does not provide a proper fix.

Have a look at this [page](https://docs.gitlab.com/ce/api/v3_to_v4.html) to see what changed. If you look at the changes related to the endpoints you deleted, you'll notice they should not be deleted, but adjusted:

https://docs.gitlab.com/ce/api/v3_to_v4.html
image

We keep track of the V4 support using issue #151 which contains the same list, but with tickboxes to tick off items that are done.

Are you willing to update this PR so it properly fixes the changes? If not I'll close this PR. Thx!

CHERTS:
I fixed the projects_test, please run a check in Travis

svanharmelen:
@CHERTS please read my comment again... The failing tests don't matter much since the changes themselves just don't make sense. You're not fixing anything, you're just deleting things.

The endpoints that you deleted still exist in the V4 API, but they are used slightly different. So the code should be updated/changed, not deleted.

CHERTS:
In APIv4, there are no endpoints "/project/owned" "/project/all" "/projects/starred" and this can easily be seen in this query:

curl --request GET -H "PRIVATE-TOKEN: XXXXX" -H "Accept: application/json" -H "Content-Type:application/json" https://gitlab.xxxxx.ru/api/v4/projects/all | jq .

{
  "message": "404 Project Not Found"
}

This endpoints only exist in APIv3

For example, APIv3 had:
curl --request GET -H "PRIVATE-TOKEN: XXXXX" -H "Accept: application/json" -H "Content-Type:application/json" https://gitlab.xxxxx.ru/api/v4/projects/owned | jq .

In APIv4 was:
curl --request GET -H "PRIVATE-TOKEN:  XXXXX" -H "Accept: application/json" -H "Content-Type:application/json" https://gitlab.xxxxx.ru/api/v4/projects?owned=true | jq .

I fix in the code "type ListProjectsOptions struct" and removed the func ListOwnedProjects, ListStarredProjects, ListAllProjects. These functions are no longer needed for APIv4

CHERTS:
If we do not understand each other, then please close the my pull request.

svanharmelen:
@CHERTS I understand, but I don't agree that we should just remove these functions...

The functions should remain in the code, but the function bodies should be changed so they no longer call the old endpoints, but use the new format (e.g. `/projects?owned=true`).

As the endpoint is maybe gone, but the functionality isn't. You can still list all owned or starred projects, but the API just expects a different call for that...

Do you see my point?

CHERTS:
I understand that deleting these functions (ListOwnedProjects, ListStarredProjects, ListAllProjects) will violate the compatibility, but this is normal if the API changes a lot, which is the case.

Why duplicate the functions of ListOwnedProjects, ListStarredProjects, ListAllProjects if the use of the function ListProjects does the same?

gitlab.go
```diff
@@ -33,7 +33,7 @@ import (
 )
 
 const (
-	libraryVersion = "0.1.1"
+	libraryVersion = "0.1.2"
```
svanharmelen: Please make this `0.2.0`

gitlab.go
```diff
@@ -33,7 +33,7 @@ import (
...
 	defaultBaseURL = "https://gitlab.com/api/v3/"
```
svanharmelen: I noticed you first changed this `v4`, but changed it back again. I guess it's fine to change this to `v4` as this branch is for the v4 API.

projects.go
```diff
@@ -168,7 +168,9 @@ type ListProjectsOptions struct {
 	Search     *string               `url:"search,omitempty" json:"search,omitempty"`
 	Simple     *bool                 `url:"simple,omitempty" json:"simple,omitempty"`
 	Visibility *VisibilityLevelValue `url:"visibility,omitempty" json:"visibility,omitempty"`
-	Statistics *bool                 `url:"statistics,omitempty" json:"statistics,omitempty"`
```
svanharmelen: Looking at the docs, I notice that `Statistics` is still a valid option. So could you please add back `Statistics to the bottom of this list? And to satisfy my OCD, could you also relocate `Visibility` to be the last entry? Thx!

```
...
Owned
Membership
Starred
Statistics
Visibility
```

projects_test.go
```diff
@@ -22,12 +22,14 @@ func TestListProjects(t *testing.T) {
 			"search":     "query",
 			"simple":     "true",
 			"visibility": "public",
-			"statistics": "true",
+			"owned": "false",
+			"membership": "false",
+			"starred": "false",
```
svanharmelen: Please update the tests as well so they include `statistics` again...

CHERTS:
Ok, fixed libraryVersion and back Statistics

svanharmelen:
Thanks @CHERTS!

PR #175, opened 2017-05-17
Added missing parameters to ProjectCreate

dihedron:
Hallo, this pull request contains an update to the CreateProjectOptions struct (added "default_branch" and "only_allow_merge_requests_if_all_discussions_are_resolved") and to the CreateProjectForUserOptions struct, which once modified is identical to the "general purpose" version and thus was redefined as a type alias for the sake of conciseness.

The struct fields were reordered according to the API docs, so it's easier to scan through the list.

During my tests I discovered that the "default_branch" does not seem to work with API V3, neither at creation time nor via EditProject(). A bug in GitLab maybe? Anyway I left the the option there because it does no harm and the docs state it should be there at least for "CreateForUser" and "Edit", so I'd rather keep it. 

Please let me know if you prefer to drop it.
Cheers!

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -378,24 +378,26 @@ func (s *ProjectsService) GetProjectEvents(pid interface{}, opt *GetProjectEvent
 // GitLab API docs:
 // https://gitlab.com/gitlab-org/gitlab-ce/blob/8-16-stable/doc/api/projects.md#create-project
 type CreateProjectOptions struct {
-	Name                          *string               `url:"name,omitempty" json:"name,omitempty"`
-	Path                          *string               `url:"path,omitempty" json:"path,omitempty"`
-	NamespaceID                   *int                  `url:"namespace_id,omitempty" json:"namespace_id,omitempty"`
-	Description                   *string               `url:"description,omitempty" json:"description,omitempty"`
-	IssuesEnabled                 *bool                 `url:"issues_enabled,omitempty" json:"issues_enabled,omitempty"`
-	MergeRequestsEnabled          *bool                 `url:"merge_requests_enabled,omitempty" json:"merge_requests_enabled,omitempty"`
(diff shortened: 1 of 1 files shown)


svanharmelen:
Thanks again @dihedron! LGTM!

I do however wonder if it makes sense to put your efforts in improving master (e.g. API V3 support) while that will be deprecated by Gitlab in not too long. So maybe your time is better spent on fixing these issues in the `f-api-v4` branch, which will be merged into master when all issues in #151 are done.

dihedron:
Hi @svanharmelen,
yes I would like to work on V4 too and I would like to become a regular contributor to this library, which is a good piece of engineering and is serving me well.

The point is, I have to prepare a working demo for a project I'm working on at work involving the automatic, declarative provisioning of resources by means of a tool which at the moment uses go-gitlab vendored at version 0.40, so no V4 support there. Moreover, this tool's upstream gitlab plugin only supports project creation with the bare minimum of parameters, so there's a lot of ground to cover still. 

I tried to explain my use case in a comment on issue #151 which has received no attention so far; the way we implement V4 support in go-gitlab is crucial for the adoption of the library in the aforementioned tool; i would like to see both APIs supported so I can implement a Terraform plugin which handles **both** older and newer GitLab instances. That would be contributed back to the community too, there's nothing proprietary in what I'm doing so...

Cheers!

svanharmelen:
@dihedron check, makes sense... I'll respond to your comment on issue #151 in a bit...

_FYI:_ I have a commit bit on Terraform as well, so if you need any help there, you can ping me for that as well...

dihedron:
@svanharmelen thanks a lot. 

At the moment I'm working on an external (plugin) provider to avoid the cost of re-linking the whole terraform (120 MB) each time. It really saved my life when I realised just how huge a beast Terraform is.

You can take a look at my code [here](https://github.com/dihedron/terraform-provider-gitlabx); I just finished testing at it looks quite stable so far. Projects and Groups are completely implemented, now I'll go on adding more resources.

PR #177, opened 2017-05-20
Bring List Repository Tree in line with current API docs

ArdaXi:
Heya,

Noticed a couple parameters missing from the ListTree API.

Also noticed the efforts for v4, this should translate over 1:1 with the exception of `RefName` having been renamed to `Ref`. I'm absolutely willing to update that branch as well but I kinda needed this functionality now. :smile:


svanharmelen:
Hi @ArdaXi! Looks good, thx!

PR #178, opened 2017-05-23
Add ListVisibleProjects to projects.go

ArdaXi:
This one's a little poorly documented, but `/projects` doesn't actually do what the Godoc says. This adds the `/projects/visible` endpoint which does, and updates the Godoc to be accurate again.


svanharmelen:
LGTM! Thanks again!

PR #179, opened 2017-05-24
Add Impersonation Tokens Support and Add missing field to CreateGroupOptions

ralphg6:
Add missing field to CreateGroupOptions for SubGroup Support

### DIFF
--- a/examples/impersonation.go
+++ b/examples/impersonation.go
@@ -1,0 +1,46 @@
+package main
+
+import (
+	"log"
+
+	"github.com/xanzy/go-gitlab"
+)
+
+func projectExample() {
+	git := gitlab.NewClient(nil, "yourtokengoeshere")
+
+	// Create new project
+	p := &gitlab.CreateProjectOptions{
+		Name:                 gitlab.String("My Project"),
+		Description:          gitlab.String("Just a test project to play with"),
+		MergeRequestsEnabled: gitlab.Bool(true),
+		SnippetsEnabled:      gitlab.Bool(true),
+		VisibilityLevel:      gitlab.VisibilityLevel(gitlab.PublicVisibility),
+	}
+	project, _, err := git.Projects.CreateProject(p)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Add a new snippet
+	s := &gitlab.CreateSnippetOptions{
+		Title:           gitlab.String("Dummy Snippet"),
+		FileName:        gitlab.String("snippet.go"),
+		Code:            gitlab.String("package main...."),
+		VisibilityLevel: gitlab.VisibilityLevel(gitlab.PublicVisibility),
+	}
+	_, _, err = git.ProjectSnippets.CreateSnippet(project.ID, s)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// List all project snippets
+	snippets, _, err := git.ProjectSnippets.ListSnippets(project.PathWithNamespace, &gitlab.ListSnippetsOptions{})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	for _, snippet := range snippets {
+		log.Printf("Found snippet: %s", snippet.Title)
+	}
+}
--- a/examples/impersonation.go
+++ b/examples/impersonation.go
@@ -6,41 +6,30 @@ import (
 	"github.com/xanzy/go-gitlab"
 )
 
-func projectExample() {
+func impersonationExample() {
 	git := gitlab.NewClient(nil, "yourtokengoeshere")
 
-	// Create new project
-	p := &gitlab.CreateProjectOptions{
-		Name:                 gitlab.String("My Project"),
-		Description:          gitlab.String("Just a test project to play with"),
-		MergeRequestsEnabled: gitlab.Bool(true),
-		SnippetsEnabled:      gitlab.Bool(true),
-		VisibilityLevel:      gitlab.VisibilityLevel(gitlab.PublicVisibility),
-	}
-	project, _, err := git.Projects.CreateProject(p)
+	userID := 1
+
+	//list impersonation token from an user
+	impersonationTList, _, err := git.Users.GetAllImpersonationTokens(userID, &gitlab.GetAllImpersonationTokensOptions{
+		State: gitlab.String("active"),
+	})
 	if err != nil {
-		log.Fatal(err)
+		panic(err)
 	}
 
-	// Add a new snippet
-	s := &gitlab.CreateSnippetOptions{
-		Title:           gitlab.String("Dummy Snippet"),
-		FileName:        gitlab.String("snippet.go"),
-		Code:            gitlab.String("package main...."),
-		VisibilityLevel: gitlab.VisibilityLevel(gitlab.PublicVisibility),
-	}
-	_, _, err = git.ProjectSnippets.CreateSnippet(project.ID, s)
-	if err != nil {
-		log.Fatal(err)
+	for _, token := range impersonationTList {
+		log.Println(token.Token)
 	}
 
-	// List all project snippets
-	snippets, _, err := git.ProjectSnippets.ListSnippets(project.PathWithNamespace, &gitlab.ListSnippetsOptions{})
+	//create an impersonation token of an user
+	impersonationT, _, err := git.Users.CreateImpersonationToken(userID, &gitlab.CreateImpersonationTokenOptions{
+		Scopes: &[]string{"api"},
+	})
 	if err != nil {
-		log.Fatal(err)
+		panic(err)
 	}
 
-	for _, snippet := range snippets {
-		log.Printf("Found snippet: %s", snippet.Title)
-	}
+	log.Println(impersonationT.Token)
 }
--- a/examples/impersonation.go
+++ b/examples/impersonation.go
@@ -12,24 +12,24 @@ func impersonationExample() {
 	userID := 1
 
 	//list impersonation token from an user
-	impersonationTList, _, err := git.Users.GetAllImpersonationTokens(userID, &gitlab.GetAllImpersonationTokensOptions{
+	tokens, _, err := git.Users.GetAllImpersonationTokens(userID, &gitlab.GetAllImpersonationTokensOptions{
 		State: gitlab.String("active"),
 	})
 	if err != nil {
 		panic(err)
 	}
 
-	for _, token := range impersonationTList {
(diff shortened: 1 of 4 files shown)


groups.go
```diff
@@ -54,12 +54,13 @@ type Group struct {
...
-	Search       *string `url:"search,omitempty" json:"search,omitempty"`
-	Sort         *string `url:"sort,omitempty" json:"sort,omitempty"`
-	Statistics   *bool   `url:"statistics,omitempty" json:"statistics,omitempty"`
+	AllAvailable *bool            `url:"all_available,omitempty" json:"all_available,omitempty"`
+	OrderBy      *string          `url:"order_by,omitempty" json:"order_by,omitempty"`
+	Owned        *bool            `url:"owned,omitempty" json:"owned,omitempty"`
+	Search       *string          `url:"search,omitempty" json:"search,omitempty"`
+	Sort         *string          `url:"sort,omitempty" json:"sort,omitempty"`
+	Statistics   *bool            `url:"statistics,omitempty" json:"statistics,omitempty"`
+	Visibility   *VisibilityValue `url:"visibility,omitempty" json:"visibility_level,omitempty"`
```
svanharmelen: `Visibility` is not part of the `ListGroupsOptions`.

users.go
```diff
@@ -582,3 +582,113 @@ func (s *UsersService) DeleteEmailForUser(uid int, eid int, options ...OptionFun
...
+	Revoked   *bool      `json:"revoked"`
+	Name      *string    `json:"name"`
+	CreatedAt *time.Time `json:"created_at"`
+	ExpiresAt *time.Time `json:"expires_at"`
+}
+
+// GetAllImpersonationTokensOptions represents the available GetAllImpersonationTokens() options.
+//
+// GitLab API docs:
+// https://gitlab.com/gitlab-org/gitlab-ce/blob/9-0-stable/doc/api/users.md#get-all-impersonation-tokens-of-a-user
```
svanharmelen: Please use these kind of links instead: `https://docs.gitlab.com/ce/api/users.html#get-all-impersonation-tokens-of-a-user`

users.go
```diff
@@ -582,3 +582,113 @@ func (s *UsersService) DeleteEmailForUser(uid int, eid int, options ...OptionFun
...
+// GitLab API docs: https://docs.gitlab.com/ce/api/users.html#get-all-impersonation-tokens-of-a-user
+type ImpersonationToken struct {
+	ID        int        `json:"id"`
+	Active    *bool      `json:"active"`
+	Token     *string    `json:"token"`
+	Scopes    *[]string  `json:"scopes"`
+	Revoked   *bool      `json:"revoked"`
+	Name      *string    `json:"name"`
+	CreatedAt *time.Time `json:"created_at"`
+	ExpiresAt *time.Time `json:"expires_at"`
```
svanharmelen: Please do not use pointers to builtin types (with the exception of `*time.Time`), unless it's in a options struct.

users.go
```diff
@@ -582,3 +582,113 @@ func (s *UsersService) DeleteEmailForUser(uid int, eid int, options ...OptionFun
...
+// https://gitlab.com/gitlab-org/gitlab-ce/blob/9-0-stable/doc/api/users.md#get-all-impersonation-tokens-of-a-user
+type GetAllImpersonationTokensOptions struct {
+	State *string `url:"state,omitempty" json:"state,omitempty"`
+}
+
+// Get all impersonation tokens of a user.
+//
+// GitLab API docs:
+// https://gitlab.com/gitlab-org/gitlab-ce/blob/9-0-stable/doc/api/users.md#get-all-impersonation-tokens-of-a-user
+func (s *UsersService) GetAllImpersonationTokens(uid int, opt *GetAllImpersonationTokensOptions, options ...OptionFunc) ([]*ImpersonationToken, *Response, error) {
```
svanharmelen: Please call `uid` just `user` for consistency.

users.go
```diff
@@ -582,3 +582,113 @@ func (s *UsersService) DeleteEmailForUser(uid int, eid int, options ...OptionFun
...
+	}
+
+	return i, resp, err
+}
+
+// Get an impersonation token of a user.
+//
+// GitLab API docs:
+// https://gitlab.com/gitlab-org/gitlab-ce/blob/9-0-stable/doc/api/users.md#get-an-impersonation-token-of-a-user
+func (s *UsersService) GetImpersonationToken(uid int, itid int, options ...OptionFunc) (*ImpersonationToken, *Response, error) {
```
svanharmelen: Please call `itid`, `token` as that this much clearer IMO

examples/impersonation.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/xanzy/go-gitlab"
+)
+
+func impersonationExample() {
+	git := gitlab.NewClient(nil, "yourtokengoeshere")
+
+	userID := 1
+
+	//list impersonation token from an user
+	impersonationTList, _, err := git.Users.GetAllImpersonationTokens(userID, &gitlab.GetAllImpersonationTokensOptions{
```
svanharmelen: I would call `impersonationTList` just `tokens`

examples/impersonation.go
```diff
@@ -0,0 +1,35 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	for _, token := range impersonationTList {
+		log.Println(token.Token)
+	}
+
+	//create an impersonation token of an user
+	impersonationT, _, err := git.Users.CreateImpersonationToken(userID, &gitlab.CreateImpersonationTokenOptions{
```
svanharmelen: I would call `impersonationT` just `token`

users.go
```diff
@@ -582,3 +582,113 @@ func (s *UsersService) DeleteEmailForUser(uid int, eid int, options ...OptionFun
...
+}
+
+// CreateImpersonationTokenOptions represents the available CreateImpersonationToken() options.
+//
+// GitLab API docs:
+// https://gitlab.com/gitlab-org/gitlab-ce/blob/9-0-stable/doc/api/users.md#create-an-impersonation-token
+type CreateImpersonationTokenOptions struct {
+	Name      *string   `url:"name" json:"name"`
+	Scopes    *[]string `url:"scopes" json:"scopes"`
+	ExpiresAt *time.Time `url:"expires_at,omitempty" json:"expires_at,omitempty"`
```
svanharmelen: Please add `omitempty` everywhere.

ralphg6: Ok i willl add... I not added because they are mandatory parameters..

users.go
ralphg6: 'itid' is simplifyed name of 'impersonation_token_id'... I think that 'token' will confuse the developer, doing think that referes content.

groups.go
ralphg6: OK

users.go
ralphg6: LGTM

users.go
svanharmelen: Yeah, I understand... But if you think about it, you would like to get an error back if you didn't specify it right? If you do not `omitempty`, then it will get a default value (the zero value of the type) instead of nothing at all. This could give unexpected behaviour as well.

users.go
ralphg6: I followed the pattern 'func (s *UsersService) DeleteEmailForUser(uid int,...', because 'uid' is a acronym of 'user_id'. Why you judge that 'user' has more consistency?

users.go
ralphg6: LTGM

users.go
svanharmelen: I understand what it stands for :wink: But if you look at https://github.com/xanzy/go-gitlab/blob/master/groups.go and see the fucnction signatures, then maybe you see my point.

I use the `gid` and `pid` which are always an `interface{}` when the API expects either a `string` or an `int`. When it only expects an `int` I write the thing it expects the ID from. So either `project` of `user`. So to keep things consistent within this package, please use `user` instead of `uid` and `token` instead of `itid`.

Thanks!

users.go
svanharmelen: I understand what it stands for... But please see my previous reaction on the `uid` comment.

groups.go
```diff
@@ -54,12 +54,12 @@ type Group struct {
...
-	Owned        *bool   `url:"owned,omitempty" json:"owned,omitempty"`
-	Search       *string `url:"search,omitempty" json:"search,omitempty"`
-	Sort         *string `url:"sort,omitempty" json:"sort,omitempty"`
-	Statistics   *bool   `url:"statistics,omitempty" json:"statistics,omitempty"`
+	AllAvailable *bool            `url:"all_available,omitempty" json:"all_available,omitempty"`
+	OrderBy      *string          `url:"order_by,omitempty" json:"order_by,omitempty"`
+	Owned        *bool            `url:"owned,omitempty" json:"owned,omitempty"`
+	Search       *string          `url:"search,omitempty" json:"search,omitempty"`
+	Sort         *string          `url:"sort,omitempty" json:"sort,omitempty"`
+	Statistics   *bool            `url:"statistics,omitempty" json:"statistics,omitempty"`
```
svanharmelen: Is this formatted using `gofmt`?

svanharmelen: Is this formatted using `gofmt`?

users.go
```diff
@@ -582,3 +582,113 @@ func (s *UsersService) DeleteEmailForUser(uid int, eid int, options ...OptionFun
...
+// GitLab API docs: https://docs.gitlab.com/ce/api/users.html#get-all-impersonation-tokens-of-a-user
+type ImpersonationToken struct {
+	ID        int        `json:"id"`
+	Active    bool      `json:"active"`
+	Token     string    `json:"token"`
+	Scopes    []string  `json:"scopes"`
+	Revoked   bool      `json:"revoked"`
+	Name      string    `json:"name"`
+	CreatedAt *time.Time `json:"created_at"`
+	ExpiresAt *time.Time `json:"expires_at"`
```
svanharmelen: Please format your gode using `gofmt`.

users.go
svanharmelen: Please change this URL as well...

svanharmelen:
@ralphg6 I also noticed you added this one:

image

Please remove it and squash you commits as well to keep things nice and tidy a bit :wink:

users.go
ralphg6: Hehe.. Sorry, i edited directly on GitHub!

svanharmelen:
@ralphg6 looks good now. Thanks for your efforts!

PR #183, opened 2017-07-11
Supports v4 get repository files

neodooth:


svanharmelen:
@neodooth thanks for the PR! Two small questions...

1. Could you make sure the tests are also updated so they are passing?
2. Could you update the related endpoints as well so this PR fixes these as well:

image

Thanks!

neodooth:
Hi,
The test is updated now.
As for other endpoints it may take more time to update since actually I needed to use get files so I made this quick fix. If it's ok you can merge it for now in case I can't find enough time to update other endpoints.

svanharmelen:
Fair enough.. Thanks!

Issue #191, opened 2017-08-20
Add support for feature flags


pcarranza:
As per https://docs.gitlab.com/ce/api/features.html

I plan to implement this myself as I need it

pcarranza:
Closing because the branch is merged in

PR #192, opened 2017-08-20
Add Feature Flags baseline

pcarranza:
This commit adds the service, ensures that the service gets instanciated
with the rest and provides the required types and method to list
features flags.

### DIFF
--- a/feature_flags.go
+++ b/feature_flags.go
@@ -3,14 +3,13 @@ package gitlab
 import (
 	"fmt"
 	"net/url"
-	"strings"
 )
 
 // FeaturesService handles the communication with the application FeaturesService
 // related methods of the GitLab API.
 //
 // GitLab API docs:
-// https://gitlab.com/gitlab-org/gitlab-ce/blob/9-3-stable/doc/api/features.md
+// https://docs.gitlab.com/ce/api/features.html
 type FeaturesService struct {
 	client *Client
 }
@@ -18,7 +17,7 @@ type FeaturesService struct {
 // Feature represents a GitLab feature flag.
 //
 // GitLab API docs:
-// https://gitlab.com/gitlab-org/gitlab-ce/blob/9-3-stable/doc/api/features.md
+// https://docs.gitlab.com/ce/api/features.html
 type Feature struct {
 	Name  string `json:"name"`
 	State string `json:"state"`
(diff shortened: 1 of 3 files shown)


pcarranza:
This PR closes https://github.com/xanzy/go-gitlab/issues/191

feature_flags.go
```diff
@@ -0,0 +1,80 @@
...
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+// FeaturesService handles the communication with the application FeaturesService
+// related methods of the GitLab API.
+//
+// GitLab API docs:
+// https://gitlab.com/gitlab-org/gitlab-ce/blob/9-3-stable/doc/api/features.md
```
svanharmelen: Could you please update all the links so they point to `https://docs.gitlab.com/ce/api/...` instead of to a specific version of the docs.

feature_flags.go
```diff
@@ -0,0 +1,80 @@
...
+		return nil, resp, err
+	}
+	return f, resp, err
+}
+
+// SetFeatureFlag sets or creates a feature flag gate
+//
+// GitLab API docs:
+// https://gitlab.com/gitlab-org/gitlab-ce/blob/9-3-stable/doc/api/features.md
+func (s *FeaturesService) SetFeatureFlag(name, key, value string, options ...OptionFunc) (*Feature, *Response, error) {
```
svanharmelen: If I read the docs, it seems like the `key` is inferred by the API depending on the given `value`. So I would leave out the `key` param and make sure the `value` is of type `interface{}`.

feature_flags.go
```diff
@@ -0,0 +1,80 @@
...
+	u := fmt.Sprintf("features/%s", url.QueryEscape(name))
+
+	values := url.Values{}
+	values.Add(key, value)
```
svanharmelen: If we only have the value, we should use an anonymous struct here. So something like this:

```go
opt := struct {
  Value interface{} `url:"value" json:"value"`
}

req, err := s.client.NewRequest("POST", u, opt, options)
```

feature_flags.go
pcarranza: Hey! @svanharmelen there's only one thing here, the value gets posted through an HTTP form, that means that it gets always treated as a string anyway.

https://golang.org/pkg/net/url/#Values.Add

In fact, we are doing it here: https://github.com/xanzy/go-gitlab/blob/master/gitlab_test.go#L49-L55

Would it make sense to get `interface{}` and just convert it to a string inside the method?

feature_flags.go
pcarranza: Sure 👍

feature_flags.go
pcarranza: Ah, disregard, with the next comment it all makes sense. Thanks!

feature_flags.go
pcarranza: This makes a lot of sense. 👍

pcarranza:
Now, I had to do some cherry picking to bring it up.

svanharmelen:
Looks great now, thanks!

PR #193, opened 2017-08-21
Moved TimeStats / Added fields to different structs / Renaming

antony360:
Moved TimeStats / Added fields to different structs / Renaming

### Issues
The TimeStats functions are related to an issue, that's why we moved and adapted the functions to the issues.go file.
We actually kept timestats.go file for backward compatibility.

### Merge-Requests
We created the same functions as the issues's timestats functions for the merge-requests in the merge-requests.go file

### Groups
We added a `ParentID` field in the `CreateGroupOptions` struct in the groups.go file.

### Issues
We added a `Search` field in the `ListProjectIssuesOptions` struct in the issues.go file.

## Merge Requests
- We added a `Search` and a `Labels` field in the `ListMergeRequestsOptions` struct in the merge-requests.go file.
- We added a `MilestoneID`, a `Labels` and a `TargetBranch` field in the `UpdateMergeRequestOptions` struct in the merge-requests.go file.

## Renaming
We changed the `ListMergeRequests` function name and the `ListMergeRequestsOptions` struct name to `ListProjectMergeRequests` and `ListProjectMergeRequestsOptions`, indeed, it was inconsistent compared to the issues.go file using the same function (equivalent route) with `ListProjectIssues` function and `ListProjectIssuesOptions` struct.
We actually kept `ListMergeRequests` function and `ListMergeRequestsOptions` struct for backward compatibility.

### DIFF
--- a/issues.go
+++ b/issues.go
@@ -275,3 +275,128 @@ func (s *IssuesService) DeleteIssue(pid interface{}, issue int, options ...Optio
 
 	return s.client.Do(req, nil)
 }
+
+// SetTimeEstimate sets the time estimate for a single project issue.
+//
+// GitLab API docs:
+// https://gitlab.com/gitlab-org/gitlab-ce/blob/8-16-stable/doc/api/issues.md#set-a-time-estimate-for-an-issue
+func (s *IssuesService) SetTimeEstimate(pid interface{}, issue int, opt *SetTimeEstimateOptions, options ...OptionFunc) (*TimeStats, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/issues/%d/time_estimate", url.QueryEscape(project), issue)
+
+	req, err := s.client.NewRequest("POST", u, opt, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	t := new(TimeStats)
+	resp, err := s.client.Do(req, t)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return t, resp, err
+}
+
+// ResetTimeEstimate resets the time estimate for a single project issue.
+//
+// GitLab API docs:
+// https://gitlab.com/gitlab-org/gitlab-ce/blob/8-16-stable/doc/api/issues.md#reset-the-time-estimate-for-an-issue
+func (s *IssuesService) ResetTimeEstimate(pid interface{}, issue int, options ...OptionFunc) (*TimeStats, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/issues/%d/reset_time_estimate", url.QueryEscape(project), issue)
+
+	req, err := s.client.NewRequest("POST", u, nil, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	t := new(TimeStats)
+	resp, err := s.client.Do(req, t)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return t, resp, err
+}
+
+// AddSpentTime adds spent time for a single project issue.
+//
+// GitLab API docs:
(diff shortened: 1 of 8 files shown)


merge_requests.go
```diff
@@ -141,6 +143,49 @@ func (s *MergeRequestsService) ListMergeRequests(pid interface{}, opt *ListMerge
...
+
+	var m []*MergeRequest
+	resp, err := s.client.Do(req, &m)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return m, resp, err
+}
+
```
svanharmelen: Please remove `ListProjectMergeRequestsOptions` and `ListProjectMergeRequests` again. I understand your reasoning, but the function names are created inline with the docs. In the docs this API is called (or described as) `List merge requests` which makes the function name in here `ListMergeRequests`.

The issues API you refer to is called (or described as) `List project issues` so that makes `ListProjectIssues`. So there is no inconsistency here, at least not between the function names and the docs.

merge_requests.go
```diff
@@ -100,7 +100,7 @@ func (m MergeRequest) String() string {
 	return Stringify(m)
 }
 
-// ListMergeRequestsOptions represents the available ListMergeRequests()
+// ListMergeRequestsOptions represents the available ListProjectMergeRequests()
```
svanharmelen: Please revert this change.

merge_requests.go
```diff
@@ -111,6 +111,8 @@ type ListMergeRequestsOptions struct {
 	State   *string `url:"state,omitempty" json:"state,omitempty"`
 	OrderBy *string `url:"order_by,omitempty" json:"order_by,omitempty"`
 	Sort    *string `url:"sort,omitempty" json:"sort,omitempty"`
+	Search  *string `url:"search,omitempty" json:"search,omitempty"`
```
svanharmelen: I don't see `Search` as a valid option in the docs, are you sure it's an option here? Can you point me to the docs or code showing that?

merge_requests.go
```diff
@@ -111,6 +111,8 @@ type ListMergeRequestsOptions struct {
...
+	Labels  Labels  `url:"labels,comma,omitempty" json:"labels,omitempty"`
```
svanharmelen: Again this field is only supported by the V4 API, so you should really rebase and point this PR to the `f-api-v4` branch.

merge_requests.go
```diff
@@ -264,8 +309,10 @@ func (s *MergeRequestsService) CreateMergeRequest(pid interface{}, opt *CreateMe
 type UpdateMergeRequestOptions struct {
 	Title        *string `url:"title,omitempty" json:"title,omitempty"`
 	Description  *string `url:"description,omitempty" json:"description,omitempty"`
-	TargetBranch *string `url:"target_branch,omitemtpy" json:"target_branch,omitemtpy"`
 	AssigneeID   *int    `url:"assignee_id,omitempty" json:"assignee_id,omitempty"`
+	MilestoneID  *int    `url:"milestone_id,omitempty" json:"milestone_id,omitempty"`
+	Labels       Labels  `url:"labels,comma,omitempty" json:"labels,omitempty"`
+	TargetBranch *string `url:"target_branch,omitemtpy" json:"target_branch,omitemtpy"`
```
svanharmelen: Why move the `TargetBranch`? In the beginning we tried to order fields alphabetically, but now we try to keep these (more or less) inline with the ordering used in the docs. So in that respect, `TargetBranch` was in the correct spot and doesn't have to change (and `MilestoneID` and `Labels` should be switched).

svanharmelen: And again these fields only exist in the V4 API...

issues.go
```diff
@@ -274,3 +275,128 @@ func (s *IssuesService) DeleteIssue(pid interface{}, issue int, options ...Optio
...
+	}
+
+	t := new(TimeStats)
+	resp, err := s.client.Do(req, t)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return t, resp, err
+}
```
svanharmelen: I get what you're trying to do with these changes, but I think we can make this better and a bit more DRY. But I'll do that after we merge this PR...

svanharmelen:
Closed in favour of PR #198

PR #207, opened 2017-09-13
Updates projects and MR types to be compatible with V4 API

sindrepm:
In Gitlab v4 API, the 'build' term is replaced with 'jobs'. This PR updates some types that still were referencing the old 'builds' term, breaking compatibility with V4.


svanharmelen:
Cool... Thanks @sindrepm!

PR #222, opened 2017-10-03
Adding GET request for Slack Integration on a Project

rawkode:

PR #223, opened 2017-10-04
Tweak PR #222

svanharmelen:
Fixup PR #222 just a little. Most important fix that was needed, was the return type used for the Get call.


rawkode:
Thanks for this, @svanharmelen; I'll keep future PR's inline :+1:

svanharmelen:
Cool! And thank you too for the initial PR!

PR #228, opened 2017-10-19
add environments support

gawel:
I have to admit that only the ListEnvironments is tested (that the only feature I need)

Since the service is simple, other methods should be ok too... :)


svanharmelen:
Thanks @gawel! LGTM!

PR #234, opened 2017-10-31
# 204 add support for GET /users/:user_id/projects

dongmx:
This PR is to add support for "GET /users/:user_id/projects"

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -197,6 +197,26 @@ func (s *ProjectsService) ListProjects(opt *ListProjectsOptions, options ...Opti
 	return p, resp, err
 }
 
+// ListUserProjects gets a list of projects for the given user.
+//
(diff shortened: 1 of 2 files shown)


svanharmelen:
@dongmx I'll merge this as is, but there is a small issue with this code. I'll fix that in a separate commit after merging this one. Thx!

dongmx:
I'd like to know what issue is :)

svanharmelen:
See this commit: 7ef3a0e1bc22af568c2f332a4d6b781005531086 The main thing is that the function (according to the API docs) accepts either a user ID or a user name, where the first implementation only accepted an ID.

dongmx:
That's my mistake. Misguided by `:user_id`. Thanks for correct it.

PR #235, opened 2017-10-31
add support for "GET /groups/:id/members/:user_id"

dongmx:
This PR add a function implements `GET /groups/:id/members/:user_id`. The document is here : `https://docs.gitlab.com/ce/api/members.html#get-a-member-of-a-group-or-project`.

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -168,6 +168,7 @@ type Client struct {
 	Environments         *EnvironmentsService
 	Features             *FeaturesService
 	Groups               *GroupsService
+	GroupMembers         *GroupMembersService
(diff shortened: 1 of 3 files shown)


svanharmelen:
Thanks for the PR @dongmx! But this one should be implemented in it's own file. See https://github.com/xanzy/go-gitlab/blob/master/project_members.go as a reference. I guess we need to copy/paste and adjust this file as `group_members.go` to keep things consistent...

dongmx:
Yes, you're right. I'll fix that.

svanharmelen:
Cool... Thanks!

PR #236, opened 2017-10-31
add GroupMembersService

dongmx:
This PR add `GroupMembersService` which handles communication with the group members related methods of the GitLab API.
But it delete several functions used to belong `GroupsService`. There might be compatibility issues.


svanharmelen:
That looks perfect to me 👍  Thanks!

PR #237, opened 2017-11-02
groups api change from group to groups

etsrepo:


svanharmelen:
Thanks @etsrepo! LGTM!

PR #244, opened 2017-11-15
Rename Iid fields to IID to match golang convention

lkysow:
This is inconsistent right now, there are some fields that are named `IID` correctly.

> Words in names that are initialisms or acronyms (e.g. "URL" or "NATO") have a consistent case. ? > For example, "URL" should appear as "URL" or "url" (as in "urlPony", or "URLPony"), never as "Url". Here's an example: ServeHTTP not ServeHttp.

From https://github.com/golang/go/wiki/CodeReviewComments#initialisms

### DIFF
--- a/events.go
+++ b/events.go
@@ -126,7 +126,7 @@ type IssueEvent struct {
 		Description string `json:"description"`
 		MilestoneID int    `json:"milestone_id"`
 		State       string `json:"state"`
-		Iid         int    `json:"iid"`
+		IID         int    `json:"iid"`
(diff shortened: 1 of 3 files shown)


svanharmelen:
I'm closing this PR as the changes it contains are already merged/added by your other PR (#245) and this commit: 7c28a6b84d4ea52affd96b054c9d93805ae8f577

Thanks again for your efforts!

lkysow:
Oops, totally forgot that I was pushing to this branch :D Thanks!

Issue #239, opened 2017-11-07
MergeRequest.UpdatedAt is not a *time.Time anymore


seblegall:
Hello,

I just noticed that `MergeRequest.UpdatedAt` is now type `string` since the upgrade to API v4. Is that a choice or a mistake?

As an example : `Issue.UpdateAt` is still type `time.Time`.

I can work on a PR if this issue is an actual one. What do you think?

svanharmelen:
@seblegall unfortunately this is intentional. Please see issue #209 for more details...

seblegall:
@svanharmelen Wired.

I investigated a little bit. Here is what I found :

* Gitlab API v4 documentation does not mention any crated_at or updated_at field for the `MergeRequest` json object.
* When actually calling the `/merge_requests` endpoint those fields are present
* The date format is ok to be typed as time.Time

Here is an example of what I have : 

```json
[
    {
        "id": 17286,
        "iid": 565,
        "project_id": 341,
        "title": "",
        "description": "",
        "state": "merged",
        "created_at": "2017-11-06T13:17:37.239Z",
        "updated_at": "2017-11-06T13:36:26.063Z",
```

I understand that datetime in general are not consistent in the gitlab API v4. But for this particular example, it seems that the format is ok.

Am I right?

svanharmelen:
I agree it seems like it, but the `MergeRequest` struct is also used in [here](https://github.com/xanzy/go-gitlab/blob/master/events.go#L233) for this API: https://docs.gitlab.com/ce/user/project/integrations/webhooks.html#comment-on-merge-request

The output received from that API call contains a `merge_request` field which is 99% identical to the `MergeRequest` struct, but here the timestamps are in a different format.

So in order to be able to use the struct in both these calls, it seems these fields must be of type `string`. At least until the API uses a consistent timestamp in both API responses.

seblegall:
So sad. :-( I understand your dilemna... Maybe the solution is indeed to duplication the MergeRequest struct. But it's kind a ugly. I'll try to think about a better solution. Maybe there is a way to format the date when Unmarshaling JSON using a custom decoder...

svanharmelen:
Yeah, that should be possible and would be an easy fix for this one. But then the other issue (#227) still remains unsolved. So I think duplicating the struct is (very unfortunately) the better solution in this particular case...

svanharmelen:
Fixed by commit ee48e93a8e3b79669daebe87c77e7b1d40e876f1

Issue #227, opened 2017-10-18
Invalid struct for MergeCommentEvent.


afking:
Does not correctly unmarshal to *MergeRequest for MergeCommentEvent struct: https://godoc.org/github.com/xanzy/go-gitlab#MergeCommentEvent, as the API returns a structure similar to MergeRequestEvent: https://docs.gitlab.com/ce/user/project/integrations/webhooks.html#comment-on-merge-request.

svanharmelen:
Can you share some more details @afking? I'm not sure I understand what you mean.

Do you have an example of the error you get? And/or some example code that shows the problem?

afking:
@svanharmelen the struct `MergeCommentEvent` has a field `MergeRequest` of type `*MergeRequest`, however I think it should be the same as `MergeEvent.ObjectAttributes` instead.

```go
var e gitlab.MergeCommentEvent
err := json.Unmarshal(data, &e); err != nil {
    panic(err)
}
if e.MergeRequest.SHA == "" {
    panic("missing sha!")   //  data.merge_request.last_commit.id
}
```
Maybe I'm unmarhsalling to the wrong struct for the webhook events?

svanharmelen:
I seems the problem is that the `MergeRequest` struct is used in two API calls (here and here)

The outputs received are 99% identical, but it looks like just a few fields are not populated in the `MergeCommentEvent`.

Not sure how to proceed with this other that making two separate structs. Maybe that would also be the correct solution for issue #239

svanharmelen:
This issue is solved by PR #245

afking:
@svanharmelen thanks for the fix! This library has saved me a lot of time.

svanharmelen:
Thanks, that's good to hear 👍

PR #245, opened 2017-11-15
Use correct struct for MergeCommentEvent.MergeRequest

lkysow:
I copied the fields of `MergeEvent.ObjectAttributes` as suggested in #227.
Also referenced in #239.

I understand that you're thinking about creating separate structs, however if this definition is only used here and in `MergeEvent.ObjectAttributes` then maybe it's okay to just have this duplication?

I also haven't validated that this has all the fields expected from the API, it just seems to be working for my use-case.

### DIFF
--- a/events.go
+++ b/events.go
@@ -230,7 +230,58 @@ type MergeCommentEvent struct {
 		StDiff       *Diff  `json:"st_diff"`
 		URL          string `json:"url"`
 	} `json:"object_attributes"`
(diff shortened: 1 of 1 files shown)


svanharmelen:
Thanks @lkysow! I have the complete output of a `MergeCommentEvent` so will merge this one and the update it just a bit to make it complete...

lkysow:
Awesome!

PR #247, opened 2017-11-16
Support WebURL on project snippets

zaquestion:
Project Snippets return `web_url` in the fields. Believe adding it to the struct here should allow us to pick it up.
Ref: https://docs.gitlab.com/ce/api/project_snippets.html


svanharmelen:
Thanks @zaquestion! LGTM!

PR #258, opened 2017-11-21
add `expires_at` field to `GroupMember`

michaellihs:
add `expires_at` field to `GroupMember`

### DIFF
--- a/group_members.go
+++ b/group_members.go
@@ -41,6 +41,7 @@ type GroupMember struct {
 	State       string           `json:"state"`
 	CreatedAt   *time.Time       `json:"created_at"`
 	AccessLevel AccessLevelValue `json:"access_level"`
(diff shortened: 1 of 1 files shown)


group_members.go
```diff
@@ -41,6 +41,7 @@ type GroupMember struct {
 	State       string           `json:"state"`
 	CreatedAt   *time.Time       `json:"created_at"`
 	AccessLevel AccessLevelValue `json:"access_level"`
+	ExpiresAt   ISOTime          `json:"expires_at"`
```
svanharmelen: I think this should be `*ISOTime`... Maybe that is not needed in this case, but I prefer to keep things consistent.

michaellihs: hi @svanharmelen - I changed it accordingly - sorry, my bad...

svanharmelen:
Thanks, LGTM!

PR #260, opened 2017-11-26
Add fields

rekby:
CreateIssueOptions + field CreatedAt
UpdateIssueOptions + field UpdatedAt

### DIFF
--- a/issues.go
+++ b/issues.go
@@ -235,11 +235,12 @@ func (s *IssuesService) GetIssue(pid interface{}, issue int, options ...OptionFu
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/issues.html#new-issues
 type CreateIssueOptions struct {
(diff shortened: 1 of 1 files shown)


svanharmelen:
Thanks @rekby! One question through... I noticed there are 2 or 3 more missing params. Shouldn't those be added as well while we're at it?

rekby:
It will not fast.
Add parameter need not only change code and add fields to struct, but some test for it.
I have not time for additional work and test now.

I think will be better when sombody add fields if them need to work. And test it on self task (as minimum).
It will be better if I add fields, but I can do typo and fields will not work.

rekby:
I think it will be better to commit these fields and add other in other pull request - when it will need to somebody.

svanharmelen:
Ok that's fine. I'll add them myself... Thx!

PR #261, opened 2017-11-27
+ ClosesIssues

rekby:


merge_requests.go
```diff
@@ -328,6 +328,27 @@ func (s *MergeRequestsService) GetMergeRequestChanges(pid interface{}, mergeRequ
 	return m, resp, err
 }
 
+// CreateMergeRequest Get all the issues that would be closed by merging the provided merge request.
```
svanharmelen: This comment starts with the wrong function name.

merge_requests.go
```diff
@@ -328,6 +328,27 @@ func (s *MergeRequestsService) GetMergeRequestChanges(pid interface{}, mergeRequ
...
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/merge_requests.html#list-issues-that-will-close-on-merge
+func (s *MergeRequestsService) ClosesIssues(pid interface{}, mr_iid int, options ...OptionFunc) ([]*Issue, *Response, error) {
```
svanharmelen: I don't think `ClosesIssues` is a good (not consistent) name. Maybe make it `GetIssuesClosedOnMerge`? It's a bit long, but very descriptive and more inline with the other function names.

svanharmelen: And please rename `mr_iid` to mergeRequest` as it's called in other functions.

merge_requests.go
```diff
@@ -328,6 +328,27 @@ func (s *MergeRequestsService) GetMergeRequestChanges(pid interface{}, mergeRequ
...
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	u := fmt.Sprintf("/projects/%s/merge_requests/%v/closes_issues", url.QueryEscape(project), mr_iid)
```
svanharmelen: Please swap this line with the blank line above it.

merge_requests.go
```diff
@@ -328,6 +328,27 @@ func (s *MergeRequestsService) GetMergeRequestChanges(pid interface{}, mergeRequ
...
+	req, err := s.client.NewRequest("GET", u, nil, options)
```
svanharmelen: So don't check the error here... You should always check and handle errors.

svanharmelen:
Thanks @rekby! 🎉

PR #262, opened 2017-11-28
+ TimeStats to issue

rekby:


svanharmelen:
@rekby there is a `TimeStats` struct containing these fields, so I guess we should just use that one here instead right? (https://github.com/xanzy/go-gitlab/blob/master/time_stats.go)

rekby:
Yes, thanks.
First time I did same as near struct Author.

svanharmelen:
Thanks @rekby!

PR #263, opened 2017-11-28
Fix panic when []OptionFunc contains nil

opalmer:
From commit 42c4e6d:
```
NewRequest() can potentially be called with a
nil value present in []OptionFunc. For example
this code would trigger a panic prior to this
commit:

  _, _, err := git.Users.ModifyUser(event.UserId, options, nil)
```

Technically speaking the above is user error but I figured someone else could make the same mistake down the line. The correct way to call the above function and avoid the panic is:
```golang
  _, _, err := git.Users.ModifyUser(event.UserId, options)
```

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -356,6 +356,10 @@ func (c *Client) NewRequest(method, path string, opt interface{}, options []Opti
 	}
 
 	for _, fn := range options {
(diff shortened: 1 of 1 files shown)


svanharmelen:
Nice one! Thanks @opalmer!

PR #264, opened 2017-11-29
add missing properties for CreateProjetOptions

michaellihs:
There have been several options missing in the `CreateProjectOptions` struct. Added them according to current documentation: https://docs.gitlab.com/ce/api/projects.html#create-project

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -338,6 +338,7 @@ type CreateProjectOptions struct {
 	JobsEnabled                               *bool            `url:"jobs_enabled,omitempty" json:"jobs_enabled,omitempty"`
 	WikiEnabled                               *bool            `url:"wiki_enabled,omitempty" json:"wiki_enabled,omitempty"`
 	SnippetsEnabled                           *bool            `url:"snippets_enabled,omitempty" json:"snippets_enabled,omitempty"`
+	ResolveOutdatedDiffDiscussions            *bool            `url:"resolve_outdated_diff_discussions,omitempty" json:"resolve_outdated_diff_discussions,omitempty"`
 	ContainerRegistryEnabled                  *bool            `url:"container_registry_enabled,omitempty" json:"container_registry_enabled,omitempty"`
 	SharedRunnersEnabled                      *bool            `url:"shared_runners_enabled,omitempty" json:"shared_runners_enabled,omitempty"`
 	Visibility                                *VisibilityValue `url:"visibility,omitempty" json:"visibility,omitempty"`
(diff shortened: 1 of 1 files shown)


projects.go
```diff
@@ -328,6 +328,7 @@ func (s *ProjectsService) GetProjectEvents(pid interface{}, opt *GetProjectEvent
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/projects.html#create-project
 type CreateProjectOptions struct {
+	UserID                                    *int             `url:"user_id,omitempty" json:"omitempty,omitempty"`
```
svanharmelen: `UserID` should not be in here. It's a required field that is in the function signature already.

projects.go
```diff
@@ -347,6 +349,9 @@ type CreateProjectOptions struct {
 	OnlyAllowMergeIfAllDiscussionsAreResolved *bool            `url:"only_allow_merge_if_all_discussions_are_resolved,omitempty" json:"only_allow_merge_if_all_discussions_are_resolved,omitempty"`
 	LFSEnabled                                *bool            `url:"lfs_enabled,omitempty" json:"lfs_enabled,omitempty"`
 	RequestAccessEnabled                      *bool            `url:"request_access_enabled,omitempty" json:"request_access_enabled,omitempty"`
+	TagList                                   *[]string        `url:"tag_list,omitempty" json:"tag_list,omitempty"`
```
svanharmelen: Did you test if this works as expected? As I know that labels are also an Array, but still need some special handling. So I made a specific type for that. Not sure if that is also the case for tags?

projects.go
michaellihs: hm - I think it's not. You can only use it, if you want to set the owning user of the project. But to be honest, it had no effect on the project, when I tried it yesterday. Thought I still add it for being "feature-complete" regarding the API. The method signature has no such field:

    func (s *ProjectsService) CreateProject(opt *CreateProjectOptions, options ...OptionFunc) (*Project, *Response, error)

michaellihs: you are right - I used the attributes for "Create project for user". I'll update the PR

projects.go
michaellihs: `*[]string` works for `TagList` - tested it with my CLI:

    golab project create --name NAME --namespace_id ID --tag_list tag1,tag2,tag3

resulted in

      "tag_list": [
        "tag1",
        "tag2",
        "tag3"
      ],

in the created project

michaellihs:
I removed the `UserID` field. Tag should work as expected

svanharmelen:
Thanks @michaellihs! Looks good now 👍

PR #266, opened 2017-12-05
(mr,fix) changes_count is a string not int

zaquestion:
Ref: https://docs.gitlab.com/ce/api/merge_requests.html#list-project-merge-requests

> Note: the changes_count value in the response is a string, not an integer. This is because when an MR has too many changes to display and store, it will be capped at 1,000. In that case, the API will return the string "1000+" for the changes count.


svanharmelen:
Ah, nice one @zaquestion! LGTM! 🎉

Issue #267, opened 2017-12-08
Can i get private-token with my password?


LiangXugang:
I can get my username&password only, and how can i create the go-gitlab client??
In my app,  i cannot use OAUTH2.0 token anymore.

svanharmelen:
@LiangXugang you should be able to use your a private or personal token without any issues: https://docs.gitlab.com/ce/api/#personal-access-tokens

clns:
I think his requirement is to use the password to retrieve the token, as the token is not available at the time. This was achieved before GitLab 10.2.0 (2017-11-22) using the Session API which [was recently removed](https://gitlab.com/gitlab-org/gitlab-ce/blob/master/CHANGELOG.md#removed-5-changes).

At the moment I don't think this is possible.

LiangXugang:
@clns  yep, i use the Session API in v3-Gitlab. But it's removed in v4-Gitlab.

As my user is just a general user(not admin), I cannot create a impersonation token too.

clns:
It’s removed in v3 as well, I think it’s removed in all api versions.

svanharmelen:
I'm really sorry, but I fail to understand what the actual problem is here. If I login to `gitlab.com`and go here:

![image](https://user-images.githubusercontent.com/4171547/33824147-b7d725c6-de5d-11e7-936c-1a3f8b3d8162.png)

I can create `personal access tokens` that I can use as `private token` in the `go-gitlab` package. Isn't that all you need? Or am I missing something here?

vitalyisaev2:
@svanharmelen there are a lot of use-cases when it's not possible to obtain `personal access token` via web UI. But `Gitlab` provides no other way to achieve this. 

Suppose you're writing a service which interacts with `Gitlab` API and you have to write integration tests for it. On your CI system you deploy clean environment, run containers with `Gitlab` and your service. Than you need to get `personal access token` in order to provide access to `Gitlab` API for your service. How would you achieve this on a virtual machine without X server? In my opinion, possible options are:
1. Hack `Gitlab` database and `INSERT` several rows about `personal access tokens` inside it;
2. Deploy something like `Selenium` and imitate real user behavior in order to get token;
3. Write a function that performs several  `GET` and `POST` HTTP requests and parses resulting HTML to get token.

PR #269 adopts the third approach. I agree that it doesn't fit good to the purpose of this library (to support only API specification), but, as we can see, it can solve problems of many users of this library.

svanharmelen:
[...]

Issue #265, opened 2017-12-02
Automating work with Personal Access Token API


vitalyisaev2:
Hello, thanks for a perfect library! Would you like to accept contribution that will provide a way to get Personal Access Token API?

The interface would be something like:

```
// url - Gitlab API Token
// login, password - Gitlab user credentials
// name, expiresAt - name and expiration date for new token
NewPersonalAccessToken(url, login, password, name string, expiresAt *time.Time) (string, error) { //... }
```

antony360:
Interesting contribution, but how would you do that ? I don't think the API provides this option ...

svanharmelen:
Hi @vitalyisaev2! PR's are always welcome. Not sure what the signature of the function should be, but I would say just make a PR that you think should work and fits in with the rest of the code. And then we'll take it from there 👍 

Thanks!

LiangXugang:
I need this too!!! But i don't find a good way to do that. 
Please @@@@me if PR about this  made.
thx!

svanharmelen:
@LiangXugang this should now be possible using the `NewBasicAuthClient`!

PR #269, opened 2017-12-10
Function to get personal access token for Gitlab API

vitalyisaev2:
Hello, please review this patch (relates to #265, #267). The introduced solution is based on HTML parsing and may seem dirty. However, as far as I could check it, it works with the most recent versions of `Gitlab`: 10.1.4 and 10.2.2.

Also, if the proposed solution is acceptable for you, I would like to discuss the way of testing this. Should it be integration tests based on `Travis CI`? Or shall we confine ourselves to unit-testing only?

@LiangXugang I will appreciate it if you give some feedback on this function (whether it meets your needs).

Kind regards,
Vitaly

### DIFF
--- a/personal_access_token.go
+++ b/personal_access_token.go
@@ -1,0 +1,221 @@
+package gitlab
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/cookiejar"
+	"net/url"
+	"path"
+	"strings"
+	"time"
+
+	"github.com/yhat/scrape"
+	"golang.org/x/net/html"
+	"golang.org/x/net/html/atom"
+)
+
+const (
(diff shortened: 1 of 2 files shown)


svanharmelen:
Hi @vitalyisaev2, I'm sorry but I don't think I would like to merge this PR.

The more I read up on it, the more I get the feeling you are trying to do something that isn't supported (anymore). Please join the discussion in issue #267 so we have one location were we can talk about this feature and why (and if) it is needed...

svanharmelen:
Sorry @vitalyisaev2, but I'm not going to merge this. As you mentioned yourself this approach feels very dirty and hacky. I do understand your challenges so let me see if I can come up with another solution to your problem...

PR #270, opened 2017-12-10
fix expected return code for block / unblock user

michaellihs:
According to the API documentation, a `201 OK` will be returned
upon success for the block / unblock user call.

See https://docs.gitlab.com/ce/api/users.html#block-user


svanharmelen:
Thanks @michaellihs! LGTM!

PR #273, opened 2017-12-11
Fix GetRawFile

xuzhenglun:
`GetRawFile` is trying to decode `response.Body` as a JSON just like `GetFile` do.
But Gitlab should return a raw file rather than a JSON. so I think that we should return a `[]byte`.

And I add a max size of file in `GetRawFileOptions `. Because it's possible that some file is very large, it may be a a hidden danger of performance issuse. 

Please review the PR, if there is someting wrong, let me known. Thanks.

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -137,7 +137,7 @@ func (l *NotificationLevelValue) UnmarshalJSON(data []byte) error {
 
 // List of valid notification levels.
 const (
-	DisabledNotificationLevel NotificationLevelValue = iota
+	DisabledNotificationLevel      NotificationLevelValue = iota
 	ParticipatingNotificationLevel
 	WatchNotificationLevel
 	GlobalNotificationLevel
@@ -436,7 +436,7 @@ func (r *Response) populatePageValues() {
 			}
 
 			// try to pull out page parameter
-			url, err := url.Parse(segments[0][1 : len(segments[0])-1])
+			url, err := url.Parse(segments[0][1: len(segments[0])-1])
 			if err != nil {
 				continue
 			}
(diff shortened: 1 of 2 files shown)


svanharmelen:
Hi @xuzhenglun, thanks for the fix. I agree that `GetRawFile` should indeed return the raw bytes as  `[]byte`.

But I don't really see why you would want to limit the maximum response? You already made the API call, so the complete file is already send over the wire into the response body.

If you care about writing big files to disk you could add a check for that in you own code that calls this function. You just check the length of the `[]byte` you get returned and then decide if you want to continue processing the result.

As this is not part of the Gitlab API, I don't think we should add it in here...

svanharmelen:
Oke... So I thought about this and tested a few things, and decided I am not going to merge this PR when it contains the additional limited reader.

I think the better approach/solution for your use case, is to first use the `GetFile` API call to retrieve the file details. Among these details is the file size, which you can then use to determine if you really want to go ahead and retrieve the file or that the file is too big to be retrieved in which case you can give a nice error and save some CPU cycles and network bandwidth by starting a download which may end up to be too big...

Does that sound like a good solution?

svanharmelen:
I noticed you didn't `gofmt` the PR and that you initialized the buffer as `var buffer bytesBuffer = bytes.NewBuffer(nil)`. This can be done a bit cleaner, so I just fixed the return type of `GetRawFile` in this commit: d2ef590f12bb040a64f5169cf5457113beea09ec

So this PR can be closed now as I think you should be able to continue now. If not feel free to open an issue to discuss things further! Thanks again!

PR #274, opened 2017-12-12
fix event structs

xuzhenglun:
I find that some struct of Event payload has some flaws. and I fix it.

the source code has been formatted, so the patch seems a little mess, sorry for that.

### DIFF
--- a/events.go
+++ b/events.go
@@ -49,9 +49,21 @@ type PushEvent struct {
 		WebURL            string          `json:"web_url"`
 		Visibility        VisibilityValue `json:"visibility"`
 	} `json:"project"`
-	Repository        *Repository `json:"repository"`
-	Commits           []*Commit   `json:"commits"`
-	TotalCommitsCount int         `json:"total_commits_count"`
+	Repository *Repository `json:"repository"`
+	Commits    []*struct {
+		ID        string     `json:"id"`
+		Message   string     `json:"message"`
+		Timestamp *time.Time `json:"timestamp"`
+		URL       string     `json:"url"`
+		Author    struct {
+			Name  string `json:"name"`
+			Email string `json:"email"`
+		} `json:"author"`
+		Added    []string `json:"added"`
+		Modified []string `json:"modified"`
+		Removed  []string `json:"removed"`
+	} `json:"commits"`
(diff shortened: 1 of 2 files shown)


svanharmelen:
Thanks @xuzhenglun! Can you please use `gofmt` to format the file?

events.go
```diff
@@ -49,9 +49,21 @@ type PushEvent struct {
...
+		Message   string     `json:"message"`
+		Timestamp *time.Time `json:"timestamp"`
+		URL       string     `json:"url"`
+		Added     []string   `json:"added"`
+		Modified  []string   `json:"modified"`
+		Removed   []string   `json:"removed"`
+		Author struct {
+			Name  string `json:"name"`
+			Email string `json:"email"`
+		} `json:"author"`
```
svanharmelen: Would be great if you can keep the same order as in the docs. So that would place `Author` between `URL` and `Added`.

events.go
```diff
@@ -84,9 +96,21 @@ type TagEvent struct {
...
+	} `json:"commits"`
```
svanharmelen: Same here of course...

events.go
```diff
@@ -186,7 +210,13 @@ type CommitCommentEvent struct {
 			DeletedFile bool   `json:"deleted_file"`
 		} `json:"st_diff"`
 	} `json:"object_attributes"`
-	Commit *Commit `json:"commit"`
+	Commit *struct {
+		ID        string     `json:"id"`
+		Message   string     `json:"message"`
+		Timestamp *time.Time `json:"timestamp"`
+		URL       string     `json:"url"`
+		Author    *Author    `json:"author"`
```
svanharmelen: Is using the `Author` struct good here? As you didn't use that in the other locations?

xuzhenglun:
@svanharmelen Thanks for reviewing!

1. Code format issues have fixed 
2. Order issues have fixed 
3. I just check the doc and found all `author` are different with `Author`, so I changed them all.
ref:  https://docs.gitlab.com/ce/user/project/integrations/webhooks.html#comment-on-commit

svanharmelen:
Looks great now! Thansk @xuzhenglun!

PR #278, opened 2017-12-13
fix projectNamespace struct

xuzhenglun:
Hi: 
I found that the `namespace` in `project` is different with doc when I trying to fetch subgroup's fullpath.

ref: https://docs.gitlab.com/ce/api/projects.html#list-all-projects


svanharmelen:
LGTM @xuzhenglun! Thanks 🎉

PR #281, opened 2017-12-13
Add support for merge requests: subscribe / unsubscribe

michaellihs:
see https://docs.gitlab.com/ce/api/merge_requests.html#subscribe-to-a-merge-request
and https://docs.gitlab.com/ce/api/merge_requests.html#unsubscribe-from-a-merge-request

### DIFF
--- a/merge_requests.go
+++ b/merge_requests.go
@@ -511,6 +511,60 @@ func (s *MergeRequestsService) CancelMergeWhenPipelineSucceeds(pid interface{}, 
 	return m, resp, err
 }
 
+// Subscribe subscribes the authenticated user to the given merge request
+// to receive notifications. If the user is already subscribed to the
+// merge request, the status code 304 is returned.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/merge_requests.html#subscribe-to-a-merge-request
+func (s *MergeRequestsService) SubscribeToMergeRequest(pid interface{}, mergeRequest int, options ...OptionFunc) (*MergeRequest, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/merge_requests/%d/subscribe", url.QueryEscape(project), mergeRequest)
+
+	req, err := s.client.NewRequest("POST", u, nil, options)
+	if err != nil {
+	    return nil, nil, err
+	}
+
+	m := new(MergeRequest)
+	resp, err := s.client.Do(req, m)
+	if err != nil {
+		return nil, resp, err
+	}
+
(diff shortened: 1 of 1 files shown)


merge_requests.go
```diff
@@ -511,6 +511,58 @@ func (s *MergeRequestsService) CancelMergeWhenPipelineSucceeds(pid interface{},
 	return m, resp, err
 }
 
+// Subscribe subscribes the authenticated user to the given merge request to receive notifications.
+// If the user is already subscribed to the merge request, the status code 304 is returned.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/merge_requests.html#subscribe-to-a-merge-request
```
svanharmelen: Could you please keep the length of comments to ~80 chars? The URL is sometimes longer, but I try to keep the sentences to max ~80 chars.

FYI, while I'm sharing coding style preferences, for lines of code I try to limit the length to ~100 chars :wink:

merge_requests.go
```diff
@@ -511,6 +511,58 @@ func (s *MergeRequestsService) CancelMergeWhenPipelineSucceeds(pid interface{},
...
+func (s *MergeRequestsService) Subscribe(pid interface{}, mergeRequest int, options ...OptionFunc) (*MergeRequest, *Response, error) {
```
svanharmelen: I know it's a bit longer, but it feels more consistent to make this `SubscribeToMergeRequest`.

I understand that a lot of these methods could maybe be shorter because they are already defined on the `MergeRequestService`, but since I once started with names that mimic how the API call is named in the docs, I would like to keep doing that.

merge_requests.go
```diff
@@ -511,6 +511,58 @@ func (s *MergeRequestsService) CancelMergeWhenPipelineSucceeds(pid interface{},
...
+
+	return m, resp, err
+}
+
+// Unsubscribe unsubscribes the authenticated user from the given merge request to not receive notifications from that merge request.
+// If the user is not subscribed to the merge request, the status code 304 is returned.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/merge_requests.html#unsubscribe-from-a-merge-request
+func (s *MergeRequestsService) Unsubscribe(pid interface{}, mergeRequest int, options ...OptionFunc) (*MergeRequest, *Response, error) {
```
svanharmelen: So this should then be `UnsubscribeFromMergeRequest`.

Maybe it's a good thing to refactor all the method names, but that should be a package wide refactor and would need a major version update... So for now let's follow the existing convention.

michaellihs:
Hi @svanharmelen - thanks for your recommendations - I'll try to stick to them in future PRs. I overwrote the last commit so your change requests are in a single commit with my changes.

PR #282, opened 2017-12-13
Add support for merge request: create todo and merge request: get diff versions

michaellihs:
see https://docs.gitlab.com/ce/api/merge_requests.html#create-a-todo

### DIFF
--- a/merge_requests.go
+++ b/merge_requests.go
@@ -566,6 +566,33 @@ func (s *MergeRequestsService) UnsubscribeFromMergeRequest(pid interface{}, merg
 	return m, resp, err
 }
 
+// CreateTodo manually creates a todo for the current user on a merge request.
+// If there already exists a todo for the user on that merge request,
+// status code 304 is returned.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/merge_requests.html#create-a-todo
+func (s *MergeRequestsService) CreateTodo(pid interface{}, mergeRequest int, options ...OptionFunc) (*Todo, *Response, error) {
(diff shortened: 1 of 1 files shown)


merge_requests.go
```diff
@@ -511,6 +511,31 @@ func (s *MergeRequestsService) CancelMergeWhenPipelineSucceeds(pid interface{},
 	return m, resp, err
 }
 
+// CreateTodo manually creates a todo for the current user on a merge request. If there already exists a todo for the user on that merge request, status code 304 is returned.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/merge_requests.html#create-a-todo
```
svanharmelen: Same thing about the comments here... Try to keep them to ~80 chars. Thx!

michaellihs:
Fixed issues with comment and overwrote last commit.

svanharmelen:
LGTM, but needs to rebased now after merging PR #281...

michaellihs:
Done. If in the future I open up several PRs in at once, should I always add the changes from "previous" PRs to "following" PRs to avoid this re-work?

svanharmelen:
One way to do this is by creating the branch for the 2nd PR from the branch of the 1st PR instead of from master. And then the 3rd should be a branch of the 3rd. But this still is not a 100% solution as other PR's might be merged in between which would again wreck your other PR's.

In this case were all changes are pretty related and all take place on the same service and in the same file, I would just make 1 PR with 3 separate commits. One for each feature. That way you will never have this kind of problem...

But thanks for fixing, all looks good now!

PR #287, opened 2017-12-18
Add support for labels: subscribeTo / unsubscribeFrom

michaellihs:
this PR adds support for subscribing / unsubscribing to / from a label. See 
* https://docs.gitlab.com/ce/api/labels.html#subscribe-to-a-label
* https://docs.gitlab.com/ce/api/labels.html#unsubscribe-from-a-label

### DIFF
--- a/labels.go
+++ b/labels.go
@@ -162,3 +162,59 @@ func (s *LabelsService) UpdateLabel(pid interface{}, opt *UpdateLabelOptions, op
 
 	return l, resp, err
 }
+
+// SubscribeToLabel subscribes the authenticated user to a label to receive
+// notifications. If the user is already subscribed to the label,
+// the status code 304 is returned.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/labels.html#subscribe-to-a-label
+func (s *LabelsService) SubscribeToLabel(pid interface{}, labelID interface{}, options ...OptionFunc) (*Label, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+	label, err := parseID(labelID)
+	if err != nil {
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/labels/%s/subscribe", url.QueryEscape(project), label)
+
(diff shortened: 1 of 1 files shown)


labels.go
```diff
@@ -162,3 +162,60 @@ func (s *LabelsService) UpdateLabel(pid interface{}, opt *UpdateLabelOptions, op
...
+	}
+	u := fmt.Sprintf("projects/%s/labels/%s/unsubscribe", url.QueryEscape(project), url.QueryEscape(label))
+
+	req, err := s.client.NewRequest("POST", u, nil, options)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := s.client.Do(req, nil)
+	return resp, err
```
svanharmelen: This should just be `return s.client.Do(req, nil)`. No need to first capture the returned values in this case...

labels.go
```diff
@@ -162,3 +162,60 @@ func (s *LabelsService) UpdateLabel(pid interface{}, opt *UpdateLabelOptions, op
...
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, err
+	}
+	label, err := parseID(labelID)
+	if err != nil {
+		return nil, err
+	}
+	u := fmt.Sprintf("projects/%s/labels/%s/unsubscribe", url.QueryEscape(project), url.QueryEscape(label))
+
```
svanharmelen: According to the docs, `label` should not be url encoded...

labels.go
```diff
@@ -162,3 +162,60 @@ func (s *LabelsService) UpdateLabel(pid interface{}, opt *UpdateLabelOptions, op
...
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+	label, err := parseID(labelID)
+	if err != nil {
+	    return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/labels/%s/subscribe", url.QueryEscape(project), url.QueryEscape(label))
+
```
svanharmelen: According to the docs, `label` should not be url encoded...

michaellihs:
changed according to recommendations and amended to last commit

svanharmelen:
LGTM! Thanks!

PR #288, opened 2017-12-18
Add support for protected branches

michaellihs:
This is just a first draft of a PR for protected branches. I would like to know whether you agree with the basic implementation and naming of new types (so that I don't have to change everything if not). Would be great if you could give me some feedback, then I'll go on implementing missing features.

Please ignore `labels` implementation, this PR is based on #287

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -62,6 +62,7 @@ type AccessLevelValue int
 //
 // GitLab API docs: https://docs.gitlab.com/ce/permissions/permissions.html
 const (
+	NoPermissions        AccessLevelValue = 0
 	GuestPermissions     AccessLevelValue = 10
 	ReporterPermissions  AccessLevelValue = 20
 	DeveloperPermissions AccessLevelValue = 30
@@ -216,6 +217,7 @@ type Client struct {
 	Projects             *ProjectsService
 	ProjectMembers       *ProjectMembersService
 	ProjectSnippets      *ProjectSnippetsService
+	ProtectedBranches    *ProtectedBranchesService
 	Pipelines            *PipelinesService
 	PipelineTriggers     *PipelineTriggersService
 	Repositories         *RepositoriesService
@@ -292,6 +294,7 @@ func newClient(httpClient *http.Client, tokenType tokenType, token string) *Clie
 	c.ProjectSnippets = &ProjectSnippetsService{client: c}
 	c.Pipelines = &PipelinesService{client: c}
 	c.PipelineTriggers = &PipelineTriggersService{client: c}
+	c.ProtectedBranches = &ProtectedBranchesService{client: c}
 	c.Repositories = &RepositoriesService{client: c}
 	c.RepositoryFiles = &RepositoryFilesService{client: c}
 	c.Services = &ServicesService{client: c}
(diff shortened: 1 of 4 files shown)


gitlab.go
```diff
@@ -137,7 +153,7 @@ func (l *NotificationLevelValue) UnmarshalJSON(data []byte) error {
 
 // List of valid notification levels.
 const (
-	DisabledNotificationLevel NotificationLevelValue = iota
+	DisabledNotificationLevel      NotificationLevelValue = iota
```
svanharmelen: You didn't run `gofmt` or `goimports`, I think you should really tweak your editor so that this always runs on save. Then you can just forget about it...

gitlab.go
```diff
@@ -69,6 +69,22 @@ const (
...
+// https://docs.gitlab.com/ce/api/protected_branches.html#protected-branches-api
+type ProtectedBranchAccessLevel int
+
+// List of available access levels for protected branches
+//
+// GitLab API: https://docs.gitlab.com/ce/api/protected_branches.html#protected-branches-api
+const (
+	NoAccess        ProtectedBranchAccessLevel = 0
+	DeveloperAccess ProtectedBranchAccessLevel = 30
+	MasterAccess    ProtectedBranchAccessLevel = 40
```
svanharmelen: These access levels are actually the same as the `AccessLevelValue` types. So I would only add the missing `NoPermissions` one and then reuse those instead of:

```
  NoPermissions     AccessLevelValue = 0
```

That would eliminate the use of these custom dedicated types...

protected_branches.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// ProtectedBranch represents a protected branch
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/protected_branches.html#list-protected-branches
+type ProtectedBranch struct {
+	Name              string                    `json:"name"`
+	PushAccessLevels  []BranchAccessDescription `json:"push_access_levels"`
+	MergeAccessLevels []BranchAccessDescription `json:"merge_access_levels"`
```
svanharmelen: I prefer to use pointers here. So like `[]*BranchAccessDescription`

protected_branches.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var p []*ProtectedBranch
+	resp, err := s.client.Do(req, &p)
+	if err != nil {
+		return nil, resp, err
+	}
+	return p, resp, err
```
svanharmelen: Please add a blank line above the `return` statement.

gitlab.go
michaellihs: hm - strange... for me `go fmt` removes this whitespace, since I didn't touch this line at all manually. Besides: thx for the hint with the auto-save, I configured that now

svanharmelen: But that is correct... The whitespace shouldn't be there and was added in your commit.

svanharmelen:
Looks good now! Do you want to add more to this PR? Or is it ready to be merged?

michaellihs:
I will add all methods for protected branches, but will take some time. I‘ll add a comment, once I‘ve finished.

svanharmelen:
Cool 👍

michaellihs:
@svanharmelen I added all missing changes, so the PR is "finished"

svanharmelen:
🎉  LGTM @michaellihs! Thanks! 👍

PR #290, opened 2017-12-20
Add support for namespaces: get and update `Namespace` struct

michaellihs:
* implements all fields of `Namespace` object
* implement `GetNamespace`

for both see https://docs.gitlab.com/ce/api/namespaces.html#namespaces-api

### DIFF
--- a/namespaces.go
+++ b/namespaces.go
@@ -16,6 +16,10 @@
 
 package gitlab
 
+import (
+	"fmt"
+)
+
 // NamespacesService handles communication with the namespace related methods
 // of the GitLab API.
 //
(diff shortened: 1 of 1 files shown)


namespaces.go
```diff
@@ -28,9 +32,13 @@ type NamespacesService struct {
...
 type Namespace struct {
-	ID   int    `json:"id"`
-	Path string `json:"path"`
-	Kind string `json:"kind"`
+	ID                          int    `json:"id"`
+	Name                        string `json:"name"`
+	Path                        string `json:"path"`
+	Kind                        string `json:"kind"`
+	FullPath                    string `json:"full_path"`
+	ParentID                    int    `json:"parent_id"`
```
svanharmelen: It looks like `ParentID` should be a string:

image

michaellihs:
Fixed type of `ParentID` - it's `string` now.

svanharmelen:
LGTM!

PR #292, opened 2017-12-20
 BREAKING: add support for `from` in `GetUserActivities`

michaellihs:
Adds support for the `from` parameter in the `GetUserActivities` method.

This is a breaking change, since the interface of `UsersService.GetUserActivities()` changes.

See https://docs.gitlab.com/ce/api/users.html#get-user-activities-admin-only

Based on #291 and #290

### DIFF
--- a/users.go
+++ b/users.go
@@ -715,12 +715,20 @@ type UserActivity struct {
 	LastActivityOn *ISOTime `json:"last_activity_on"`
 }
 
+// GetUserActivitiesOptions represents the options for GetUserActivities
+//
+// GitLap API docs:
+// https://docs.gitlab.com/ce/api/users.html#get-user-activities-admin-only
+type GetUserActivitiesOptions struct {
+	From *ISOTime `url:"from,omitempty" json:"from,omitempty"`
+}
+
 // GetUserActivities retrieves user activities (admin only)
 //
 // GitLab API docs:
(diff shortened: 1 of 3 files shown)


svanharmelen:
Just a small FYI... Since all these 3 PR's do not actually depend on each other (the code changes are done in different files and are not dependent on each other), it would make more sense to base all these PR's on master, instead of on each other.

Only if they make changes in the same location of the same file, or the second PR uses code from the first PR then you would need to base the PR's of each others branches... 👍

users.go
```diff
@@ -715,12 +715,20 @@ type UserActivity struct {
 	LastActivityOn *ISOTime `json:"last_activity_on"`
 }
 
+// GetUserActivitiesOptions represents the options for GetUserActivities
+//
+// GitLap API docs:
+// https://docs.gitlab.com/ce/api/users.html#get-user-activities-admin-only
+type GetUserActivitiesOptions struct {
+	From *ISOTime `url:"from" json:"from"`
```
svanharmelen: You need to add `omitempty` to both these tags...

michaellihs:
Added `omitempty` to the `GetUserActivitiesOptions`

svanharmelen:
Thank @michaellihs! Merged them 👍

PR #293, opened 2017-12-21
(ci/lint) expose ci lint API

zaquestion:


gitlab.go
```diff
@@ -497,6 +497,34 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
...
+	}
+	req, err := c.NewRequest("POST", "ci/lint", opts, options)
+	if err != nil {
+		return false, err
+	}
+
+	lint := struct {
+		Status string   `json:"status"`
+		Errors []string `json:"errors"`
+	}{}
```
svanharmelen: The convention in this package is to define a named type for this, instead of an anonymous inline struct.

gitlab.go
```diff
@@ -497,6 +497,34 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
...
+	_, err = c.Do(req, &lint)
+	if err != nil {
+		return false, err
+	}
+	return lint.Status == "valid", nil
```
svanharmelen: I think we should just return the whole `lint` struct here, so people have access to both the status and any possible errors.

gitlab.go
```diff
@@ -497,6 +497,34 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
...
+// GitLab API docs: https://docs.gitlab.com/ee/api/lint.html
+func (c *Client) Lint(content string, options ...OptionFunc) (valid bool, err error) {
+	if content == "" {
+		return false, errors.New("content is missing")
+	}
+	opts := struct {
+		Content string `json:"content"`
+	}{
+		Content: content,
+	}
```
svanharmelen: The convention in this package is to define a named type for this, instead of an anonymous inline struct.

gitlab.go
```diff
@@ -497,6 +497,34 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
 	return response, err
 }
 
+// Lint validates .gitlab-ci content
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/lint.html
```
svanharmelen: Please only refer to the `ce` docs...

gitlab.go
```diff
@@ -497,6 +497,34 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
...
+func (c *Client) Lint(content string, options ...OptionFunc) (valid bool, err error) {
```
svanharmelen: It's convention that only the variables that are part of the API path are directly added in the function signature. Any other params (such as `content` in this case) are part of an options struct.

I know in this case that's maybe a bit much, but I guess I'm just a sucker for consistency :wink:

svanharmelen:
_EDIT:_ I removed 2 comments, which I think in this case do not make much sense :wink: So please only address the comments that are still here on Github 👍

gitlab.go
zaquestion: oops :+1:

gitlab.go
zaquestion: It looks like `client.Do` takes care of marshaling `error []string`  into an error type and would get returned above. I'll change it, just trying to check my understanding of the package.

zaquestion: IIRC, I was actually able to observe this in testing. It seems like the anonymous struct above really only needs the status and the caller only needs to know if its valid or not since if their are errors `client.Do` will return them.

zaquestion:
Should be able to finish this up after work, will get it moved to a separate service.

gitlab.go
svanharmelen: No, I don't think so... The errors in this struct are errors in your `ci` file. The errors returned by `client.Do` are errors caused by functions that failed to execute as expected.

svanharmelen:
Cool 👍

gitlab.go
zaquestion: @svanharmelen Perhaps I'm reading this wrong, it looks like https://github.com/xanzy/go-gitlab/blob/master/gitlab.go#L529 is doing just that?

svanharmelen: Yes, I think your reading it wrong :wink: The errors in the `Lint` response struct are syntax errors in your CI file. The errors the `client` processes are HTTP or server errors (https://github.com/xanzy/go-gitlab/blob/master/gitlab.go#L530-L533).

When your CI file contains errors, I expect that the HTTP call will still return a 200 as the HTTP call was successful. Didn't test that last part, but that should be easy to so with a curl...

zaquestion:
@svanharmelen can you give it another look?

svanharmelen:
Thanks @zaquestion! LGTM!

PR #295, opened 2017-12-23
Fixes #230: get users by username

### LINKED ISSUE #230: Support get user by username
There is an api to find users by username[1] which is not currently implemented but would be very useful downstream[2]

[1] https://docs.gitlab.com/ee/api/users.html#list-users documents
/users?username=:username.
[2] https://github.com/terraform-providers/terraform-provider-gitlab/pull/39

michaellihs:
This PR fixes #230 

Not sure though, whether we could provide a more "generic" method that handles external users as well... on the other hand, adding `/users?extern_uid=:extern_uid&provider=:provider` and `/users?external=true` would be no big deal.

### DIFF
--- a/users.go
+++ b/users.go
@@ -118,6 +118,39 @@ func (s *UsersService) GetUser(user int, options ...OptionFunc) (*User, *Respons
 	return usr, resp, err
 }
 
+// GetUsersAsAdminOptions represents the search options for GetUsersAsAdmin
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/users.html#for-admins
+type GetUsersAsAdminOptions struct {
+	Username     *string    `json:"username,omitempty" url:"username,omitempty"`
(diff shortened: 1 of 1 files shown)


svanharmelen:
Wouldn't it be better to just as an option struct here? That could maybe be a way to make all the variations work with this one function?

michaellihs:
@svanharmelen picked up your suggestion and added options here. Now we have all options but the custom attributes, since I don't know how to properly implement this.

michaellihs:
Hi @svanharmelen - just a short reminder that this PR is still open... would be great if you could merge it. If I should change anything, please let me know.

svanharmelen:
@michaellihs, since it was christmas and new year I was spending some time with my family 😏 

As for the PR, I wonder why we shouldn't just add the admin options to the `ListUsersOptions` struct? The only question is what kind of error you get when you use an admin option as a user. If the error is clear about it, I would say just add it instead of creating a new function for it...

svanharmelen:
Merged and updated your commit from the PR in PR #342. Thanks @michaellihs!

PR #298, opened 2018-01-05
(branches) fix #233, urlencode branch name

### LINKED ISSUE #233: branch need url encode
https://github.com/xanzy/go-gitlab/blob/e6f4b9d42e48b5dd1201aa16949b4034213ad50e/branches.go#L91

Recently, I used git flow in the development process, and there was a branch that like this format: feature/test, which resulted in the  query branch information api 404 not found.

I read the offical document and found that the branch and project name need url encode. Specific documents are as follows:
https://docs.gitlab.com/ce/api/branches.html#get-single-repository-branch

zaquestion:


svanharmelen:
Thanks @zaquestion! LGTM!

Issue #300, opened 2018-01-20
Cannot get coverage of individual jobs


rhbvkleef:
The GitLab API can expose a coverage number for each individual job. This number should be included in the Jobs struct, and all methods that create it should assign it.

zaquestion:
Can't seem to find any docs on the format of the coverage blob. Except that it can be `null`.
https://docs.gitlab.com/ce/api/jobs.html
https://docs.gitlab.com/ce/api/builds.html

I'll get something going in gitlab ci, hopefully have something up shortly.

zaquestion:
Looks likes its `null` or `int`

```
    "coverage": 34.6,
    "coverage": null,
```

svanharmelen:
Fixed by PR #300

PR #301, opened 2018-01-20
[#300] (jobs) return "coverage" in Jobs response

zaquestion:


jobs.go
```diff
@@ -44,6 +44,7 @@ type JobsService struct {
 type Job struct {
 	Commit        *Commit    `json:"commit"`
 	CreatedAt     *time.Time `json:"created_at"`
+	Coverage      *float64   `json:"coverage"`
```
svanharmelen: Please do not use a pointer to builtin types. Looks fine otherwise 👍

zaquestion: Ideally I'd like the API to not work this way, it's valid values are `null` and `float`. I've seen most API just go for the pointer approach -- here's some alternatives

```
type Coverage int
// And *Coverage
// or implement UnmarshalJSON on the type instead and ditch the pointer altogether 
```

LMK, otherwise I'll check for an example elsewhere in the code. I'll have have an update in the next few hours

svanharmelen:
@zaquestion there is no need for all that :wink: The zero value of a `float64` is `0`, so this wil just work without errors: https://play.golang.org/p/tV7o4wgTxyS

zaquestion:
I'm amazed, had no idea that go's json unmarshler would use the zero value in that case.

rhbvkleef:
There is a semantic difference between `null` and `0`, namely that there is a coverage number available versus the fact that there is no (or barely any) coverage at all. I don't suppose it could marshal to `NaN`?

svanharmelen:
No, that’s indeed not possible...

PR #303, opened 2018-01-23
Snippet: add Explore API call

tbruyelle:
Hey there, from what I saw, this API call is missing. 
See https://docs.gitlab.com/ce/api/snippets.html#explore-all-public-snippets

I've also added the `Snipper.RawURL` property.

Thanks for this lib!

### DIFF
--- a/snippets.go
+++ b/snippets.go
@@ -49,6 +49,7 @@ type Snippet struct {
 	UpdatedAt *time.Time `json:"updated_at"`
 	CreatedAt *time.Time `json:"created_at"`
 	WebURL    string     `json:"web_url"`
+	RawURL    string     `json:"raw_url"`
 }
 
 func (s Snippet) String() string {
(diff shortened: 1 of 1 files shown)


snippets.go
```diff
@@ -204,3 +205,22 @@ func (s *SnippetsService) SnippetContent(snippet int, options ...OptionFunc) ([]
 
 	return b.Bytes(), resp, err
 }
+
+// Explore gets the list of public snippets.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/snippets.html#explore-all-public-snippets
+func (s *SnippetsService) Explore(opt *ListSnippetsOptions, options ...OptionFunc) ([]*Snippet, *Response, error) {
```
svanharmelen: For consistency, please make an `ExploreSnippetsOptions` struct instead. I understand that in this case this would be just a copy of the `ListSnippetsOptions` struct, but that way we keep things consistent...

svanharmelen:
Thanks @tbruyelle! One small remark, looks fine otherwise...

snippets.go
tbruyelle: Done, I've also renamed `Explore` to `ExploreSnippets` for the same reason.

svanharmelen:
Thanks @tbruyelle! LGTM!

PR #305, opened 2018-01-26
Add support for list project forks

fsamin:


projects.go
```diff
@@ -903,3 +903,37 @@ func (s *ProjectsService) UploadFile(pid interface{}, file string, options ...Op
...
+type ListProjectForksOptions struct {
+	ListProjectsOptions
+	WithIssuesEnabled        bool `url:"with_issues_enabled" json:"with_issues_enabled"`
+	WithMergeRequestsEnabled bool `url:"with_merge_requests_enabled" json:"with_merge_requests_enabled"`
+}
+
+// ListProjectForks gets a list of project forks.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/projects.html#list-project-hooks
```
svanharmelen: The doc URL should be `https://docs.gitlab.com/ce/api/projects.html#list-forks-of-a-project`

projects.go
```diff
@@ -903,3 +903,37 @@ func (s *ProjectsService) UploadFile(pid interface{}, file string, options ...Op
 
 	return uf, resp, nil
 }
+
+// ListProjectForksOptions represents the available ListProjectForks() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/projects.html#list-project-hooks
```
svanharmelen: The doc URL should be `https://docs.gitlab.com/ce/api/projects.html#list-forks-of-a-project`

projects.go
```diff
@@ -903,3 +903,37 @@ func (s *ProjectsService) UploadFile(pid interface{}, file string, options ...Op
...
+type ListProjectForksOptions struct {
+	ListProjectsOptions
+	WithIssuesEnabled        bool `url:"with_issues_enabled" json:"with_issues_enabled"`
+	WithMergeRequestsEnabled bool `url:"with_merge_requests_enabled" json:"with_merge_requests_enabled"`
+}
```
svanharmelen: I think the fields `WithIssuesEnabled` and `WithMergeRequestsEnabled` should be of type `*bool` and that they should be added to the `ListProjectsOptions` struct instead (as they are valid options for that call as well).

And then we can just use `ListProjectsOptions` in the `ListProjectForks ` call.

fsamin:
Thanks for your suggestions, my branch is updated

svanharmelen:
Thanks @fsamin! LGTM!

PR #306, opened 2018-01-28
Project: add ci_config_path property

stapelberg:


projects.go
```diff
@@ -83,8 +83,9 @@ type Project struct {
 		GroupName        string `json:"group_name"`
 		GroupAccessLevel int    `json:"group_access_level"`
 	} `json:"shared_with_groups"`
-	Statistics *ProjectStatistics `json:"statistics"`
-	Links      *Links             `json:"_links,omitempty"`
+	Statistics   *ProjectStatistics `json:"statistics"`
+	Links        *Links             `json:"_links,omitempty"`
+	CiConfigPath *string            `json:"ci_config_path"`
```
zaquestion: Consider capitalizing the `I` as in `CIConfigPath`. Go likes capitalized acronyms 
Ref: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

svanharmelen:
Thanks @stapelberg! If you could just update the field name like @zaquestion suggested, I will merge the PR right after that 👍

projects.go
stapelberg: I copied this from the already-existing https://github.com/xanzy/go-gitlab/blob/062bdff086ddaf6a8de7335bcd57bf7d468c4967/projects.go#L353 — shouldn’t we keep the two consistent?

svanharmelen: Ah... Guess that one slipped through 😞  But I rather fix the other one in that case...

stapelberg:
Done.

svanharmelen:
Thanks again @stapelberg! LGTM!

PR #307, opened 2018-01-28
Add support for Runners API

stapelberg:
Not all methods are implemented, only the ones which I needed and verified.

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -222,6 +222,7 @@ type Client struct {
 	ProtectedBranches    *ProtectedBranchesService
 	Repositories         *RepositoriesService
 	RepositoryFiles      *RepositoryFilesService
+	Runners              *RunnersService
 	Services             *ServicesService
 	Session              *SessionService
 	Settings             *SettingsService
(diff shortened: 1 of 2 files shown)


projects.go
```diff
@@ -930,3 +930,59 @@ func (s *ProjectsService) ListProjectForks(pid interface{}, opt *ListProjectsOpt
...
+	}
+
+	var p *Runner
+	resp, err := s.client.Do(req, &p)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return p, resp, err
+}
```
svanharmelen: I fully understand that you put those methods here, but for I would like to follow the docs as much as possible. This is done in other situations like this (were some methods could be in multiple places) as well, so that it remains very clear for people where to find which methods (by just finding them in the docs to see in which API they are defined).

This means that I think these calls should be called `ListProjectRunner` and `EnableProjectRunner` and be defined on the `runner` service instead...

stapelberg:
Done

svanharmelen:
Thanks @stapelberg! LGTM!

PR #309, opened 2018-01-29
Subgroups

cuotos:


subgroups.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	var sg []*Subgroup
+	resp, err := s.client.Do(req, &sg)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return sg, resp, err
+}
```
svanharmelen: If you add `ListSubgroupsOptions` and `ListSubgroups` to the `groups.go` file, I think we are good to go.

The `Subgroup` type is not needed I think. Isn't that the same as the `Group` type (field wise)? And we don't want/need an `SubgroupsService` as these calls should be methods of the `GroupsService`.

Thanks!

cuotos:
A Subgroup has different fields to Group. If SubGroup is only a subset of fields to Group, would you rather I just remove the Subgroup struct and use a Group (albeit with default fields)?

cuotos:
I now see that Project is used in places that only a subset of fields are actually required, so have done the same with Groups / Subgroups

svanharmelen:
I now miss the `ListSubgroupsOptions` struct, but I'll take care of that one... So thanks for the PR @cuotos!

PR #310, opened 2018-01-29
changed NotificationLevel to pointers to support null values

prune998:
Gitlab 10.4.1 sometimes return `null` values in NotificationLevel fields. changing it to a pointer allow to support it.

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -138,13 +138,13 @@ type Permissions struct {
 // ProjectAccess represents project access.
 type ProjectAccess struct {
 	AccessLevel       AccessLevelValue       `json:"access_level"`
-	NotificationLevel NotificationLevelValue `json:"notification_level"`
+	NotificationLevel *NotificationLevelValue `json:"notification_level"`
 }
 
 // GroupAccess represents group access.
 type GroupAccess struct {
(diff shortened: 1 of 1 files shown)


svanharmelen:
Thanks @prune998! But does that mean somethings breaks when this field is not a pointer? Or will it just receive the zero value of the type?

prune998:
If it's not a pointer you get an error when `decoding` the json (trying to fit a `null` in a `int`...

svanharmelen:
Did you actually check that, or do you only assume that? As normally that is not the case: https://play.golang.org/p/L6AOu6oxC_S

prune998:
I think the behaviour may be different between "unmarshal" and newDecoder().Decode() with an io.Reader, as in the code here. 

I'll take a look tomorrow.

svanharmelen:
I just look a quick look and found the root cause... We have a custom `UnmarshalJSON` on the `NotificationLevelValue` type which causes this 😏 

I'm closing this PR and will make a quick PR to fix your issue in the proper place. Still thanks for the report! As it is a bug that needs to be fixed... So thanks @prune998.

PR #311, opened 2018-01-29
Fix the custom `UnmarshalJSON` function

svanharmelen:
Without this PR the `NotificationLevelValue` cannot properly unmarshal `null`.


svanharmelen:
@prune998 this fixes the marshalling issue you found... Thanks again for pointing it out 👍

prune998:
Working great. I confirm it's OK.
Thanks for your help !

PR #312, opened 2018-01-31
Added function to get all users from a specific project

RoValentim:
I needed a command that's not in the project:

/projects/<ID>/users

So I created in same format as the others in projects.go.

I saw that some minor changes in others structures was sent too, but this ones are changes in space only, sorry for that, but I saw it only after send it to my fork.

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -88,8 +88,8 @@ type Project struct {
 	CIConfigPath *string            `json:"ci_config_path"`
 }
 
-// Project users represents a GitLab project.
-type ProjectUsers struct {
+// ProjectUser represents a GitLab project.
+type ProjectUser struct {
 	ID        int    `json:"id"`
 	Name      string `json:"name"`
 	Username  string `json:"username"`
@@ -250,8 +250,8 @@ func (s *ProjectsService) ListUserProjects(uid interface{}, opt *ListProjectsOpt
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/projects.html#get-project-users
-func (s *ProjectsService) ListProjectsUsers(uid interface{}, opt *ListProjectsOptions, options ...OptionFunc) ([]*ProjectUsers, *Response, error) {
-	project, err := parseID(uid)
+func (s *ProjectsService) ListProjectsUsers(pid interface{}, opt *ListProjectsOptions, options ...OptionFunc) ([]*ProjectUser, *Response, error) {
+	project, err := parseID(pid)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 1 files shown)


projects.go
```diff
@@ -88,6 +88,16 @@ type Project struct {
 	CIConfigPath *string            `json:"ci_config_path"`
 }
 
+// Project users represents a GitLab project.
```
svanharmelen: The comment should start with the same name as the type, so `ProjectUsers` in this case

projects.go
```diff
@@ -88,6 +88,16 @@ type Project struct {
...
+type ProjectUsers struct {
```
svanharmelen: This should be `ProjectUser` (singular) as this describes one user

projects.go
```diff
@@ -236,6 +246,31 @@ func (s *ProjectsService) ListUserProjects(uid interface{}, opt *ListProjectsOpt
 	return p, resp, err
 }
 
+// ListProjectsUsers gets a list of users for the given project.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/projects.html#get-project-users
+func (s *ProjectsService) ListProjectsUsers(uid interface{}, opt *ListProjectsOptions, options ...OptionFunc) ([]*ProjectUsers, *Response, error) {
```
svanharmelen: This function should not take `opt *ListProjectsOptions`. So this line should just use a `nil`, instead of `opt`:

```go
req, err := s.client.NewRequest("GET", u, nil, options)
```

svanharmelen: `uid` should be `pid` (it's the project ID, not the user ID)

RoValentim: With ```opt *ListProjectsOptions``` we have more control for what we may need, if a project have 200 users, we can use pagination, sorting, ordering and so on in an easy way, but at least for what I'm using, ```nil``` aren't a problem, so any one of then are good for me.

projects.go
```diff
@@ -236,6 +246,31 @@ func (s *ProjectsService) ListUserProjects(uid interface{}, opt *ListProjectsOpt
 	return p, resp, err
 }
 
+// ListProjectsUsers gets a list of users for the given project.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/projects.html#get-project-users
+func (s *ProjectsService) ListProjectsUsers(pid interface{}, opt *ListProjectsOptions, options ...OptionFunc) ([]*ProjectUser, *Response, error) {
```
svanharmelen: Almost there... Just this needs to go:`opt *ListProjectsOptions`

projects.go
svanharmelen: Ah, check... But in that case you should make a option struct dedicated for this call. That would be fine too.

RoValentim: Done, thanks for the feedback!

svanharmelen:
LGTM! 🎉  Thanks @RoValentim!

Issue #240, opened 2017-11-07
Emoji Award API


Lorac:
Is this missing in the current state of the project? It looks like it.

https://docs.gitlab.com/ce/api/award_emoji.html

svanharmelen:
@Lorac yes, it seems this API is currently not covered...

gjnoonan:
I'm happy to work on this 👍

svanharmelen:
@gjnoonan feel free to do so... PR's are much appreciated! :wink:

gjnoonan:
@svanharmelen Whilst writing the code for this, I pondered two different ways. One, follows the same pattern as the rest of the library, where AwardEmoji has it's own service

```go
type AwardEmojiService struct {}
type AwardEmoji struct {}

func (as *AwardEmojiService) ListIssueAwardEmoji(pid interface{}, issue int, options ...OptionFunc) ([]*AwardEmoji,*Response, error) {
```
and it's usage ..
```go
b, _, err := git.AwardEmoji.ListIssueAwardEmoji(37, 21)
for _, ae := range b {
	fmt.Printf("%d\t%d\t%s\t%s\n", ae.ID, ae.AwardableId, ae.Name, ae.User.Username)
}
```

Although this provides consistency with the rest of the library, because of the slight difference in API for Award Emoji, I wondered if it was more natural to call the Method on the Issue/MergeRequest/Snippet interfaces ?

```go
func (i Issue) ListAwardEmoji() []*AwardEmoji {}
```
which would be used as such:
```go
issue, _ , _ :=git.Issues.GetIssue(37, 45)
ae, err := issue.ListAwardEmoji()
```

Happy to discuss and take direction, to which you would prefer to merge.

svanharmelen:
Yeah, this is a really good question... In the end my goal is to make this package follow (as much as possible) how the docs are organized, as I believe that will help people to easily find the things they are looking for. So in this case I think we should go for the `AwardEmoijService`, even through the other option would definately be a good solution as well...

gjnoonan:
Thanks for getting back to me @svanharmelen I will put up a PR shortly

svanharmelen:
Thanks 👍

zaquestion:
@gjnoonan are you still working on this? If not would you mind pushing up what you've got? I'm thinking about finishing it up to use in [lab](https://github.com/zaquestion/lab)

Lorac:
[...]

PR #313, opened 2018-01-31
Add Emoji Award API

Lorac:
I know there is no test for now, I'll add them as soon as possible. You can review the code whenever. I'll fix as soon as I can.

Fixes: #240

### DIFF
--- a/award_emojis.go
+++ b/award_emojis.go
@@ -1,0 +1,428 @@
+//
+// Copyright 2017, Arkbriar
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package gitlab
+
+import (
+	"fmt"
+	"net/url"
+	"time"
+)
+
+// AwardEmojiService handles communication with the emoji awards related methods
+// of the GitLab API.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/award_emoji.html
+type AwardEmojiService struct {
+	client *Client
+}
+
+// AwardEmoji represents a GitLab Award Emoji.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/award_emoji.html
+type AwardEmoji struct {
+	ID   int    `json:"id"`
+	Name string `json:"name"`
+	User struct {
+		Name      string `json:"name"`
+		Username  string `json:"username"`
+		ID        int    `json:"id"`
+		State     string `json:"state"`
+		AvatarURL string `json:"avatar_url"`
+		WebURL    string `json:"web_url"`
+	} `json:"user"`
+	CreatedAt     *time.Time `json:"created_at"`
+	UpdatedAt     *time.Time `json:"updated_at"`
+	AwardableID   int        `json:"awardable_id"`
+	AwardableType string     `json:"awardable_type"`
+}
+
+const (
+	awardMergeRequest = "merge_requests"
+	awardIssue        = "issues"
+	awardSnippets     = "snippets"
+)
+
+// ListMergeRequestAwardEmoji gets a list of all award emoji on the merge request
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/award_emoji.html#list-an-awardable-39-s-award-emoji
+func (s *AwardEmojiService) ListMergeRequestAwardEmoji(pid interface{}, mergeRequestIID int, options ...OptionFunc) ([]*AwardEmoji, *Response, error) {
+	return s.listAwardEmoji(pid, awardMergeRequest, mergeRequestIID, options...)
+}
+
+// ListIssueAwardEmoji gets a list of all award emoji on the issue
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/award_emoji.html#list-an-awardable-39-s-award-emoji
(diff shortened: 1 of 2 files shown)


award_emojis.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	return awardEmoji, resp, err
+}
+
+// CreateAwardEmojiOnNote award emoji on a note
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/award_emoji.html#award-a-new-emoji-on-a-note
+func (a *AwardEmojiService) CreateAwardEmojiOnNote(pid interface{}, issueIID, noteID int, options ...OptionFunc) (*AwardEmoji, *Response, error) {
```
zaquestion: Probably want to add a `DeleteAwardEmojiOnNote` to complete the set.

award_emojis.go
```diff
@@ -0,0 +1,219 @@
...
+	UpdatedAt     time.Time `json:"updated_at"`
+	AwardableID   int       `json:"awardable_id"`
+	AwardableType string    `json:"awardable_type"`
+}
+
+// ListAwardEmoji gets a list of all award emoji on the specified resource
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/award_emoji.html#list-an-awardable-39-s-award-emoji
+func (a *AwardEmojiService) ListAwardEmoji(pid interface{}, resource string, resourceID int, options ...OptionFunc) ([]*AwardEmoji, *Response, error) {
```
zaquestion: It seems like we should find a way to handle the resource type without requiring the caller to pass it as a string.

A couple options:
```
const (
    AwardMergeRequest = "merge_request"
    // The rest
)
```

Could unexport these functions and use them as a factory
```
func (a *AwardEmojiService) ListMergeRequestAwardEmoji(...) ... {
   return s.listAwardEmoji(pid, "merge_request", rid)
}
```

award_emojis.go
Lorac: Added

Lorac:
I pushed the new changes.

For the testing what kind of tests are you expecting?
Like the ones in `events_test.go` or like `pipelines_test.go` ?

award_emojis.go
```diff
@@ -0,0 +1,328 @@
...
+		return nil, err
+	}
+	return a.client.Do(req, nil)
+}
+
+// ListAwardEmojiOnNote list single note's award emoji
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/award_emoji.html#list-a-note-39-s-award-emoji
+func (a *AwardEmojiService) ListAwardEmojiOnNote(pid interface{}, issueIID, noteID int, options ...OptionFunc) ([]*AwardEmoji, *Response, error) {
```
zaquestion: Can you add these `OnNote` funcs to work like the others (factory func passing in constants for the resource). Its not super obvious in the docs, but it appear the notes portion of the API works the same

Ref: https://docs.gitlab.com/ce/api/award_emoji.html#award-emoji-on-notes

zaquestion:
@Lorac This is already looking great! Ultimately @svanharmelen has all the authority here. Regarding tests, I would advise towards the `pipelines_test.go` approach, `events_test.go` largely seems to be testing marshaling the types, in the past we've been able to update structs rather quickly.

award_emojis.go
Lorac: Ohh right, I see it. I'm gonna try them to be sure it's a valid endpoint for the other resources.

Lorac: I pushed the new code to support all type of resources

Lorac: I think it's probably the same for `Get` right?

zaquestion: Yeah they talk about it being supported on all resources before describing the API calls.

Lorac: @zaquestion There you go all the "factory" methods are implemented I think...

award_emojis.go
```diff
@@ -0,0 +1,412 @@
...
+	awardMergeRequest = "merge_requests"
+	awardIssue        = "issues"
+	awardSnippets     = "snippets"
+)
+
+// ListMergeRequestAwardEmoji gets a list of all award emoji on the merge request
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/award_emoji.html#list-an-awardable-39-s-award-emoji
+func (a *AwardEmojiService) ListMergeRequestAwardEmoji(pid interface{}, mergeRequestIID int, options ...OptionFunc) ([]*AwardEmoji, *Response, error) {
```
svanharmelen: The convention in this package is to always use `s` (from `service`) as the method receiver name

award_emojis.go
```diff
@@ -0,0 +1,412 @@
...
+func (a *AwardEmojiService) ListSnippetAwardEmoji(pid interface{}, snippetID int, options ...OptionFunc) ([]*AwardEmoji, *Response, error) {
+	return a.listAwardEmoji(pid, awardSnippets, snippetID, options...)
+}
+
+func (a *AwardEmojiService) listAwardEmoji(pid interface{}, resource string, resourceID int, options ...OptionFunc) ([]*AwardEmoji, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/%s/%d/award_emoji", url.QueryEscape(project), resource, resourceID)
```
svanharmelen: Please make sure lines of code ar not longer then 100 chars. This excludes method declarations, as I don't think they should be on multiple lines. But normal lines of code should not be longer then 100 chars...

FYI, for comments I try to keep it to 80 chars max but sometimes the URL's are just a bit longer which is fine...

award_emojis.go
```diff
@@ -0,0 +1,412 @@
...
+
+	req, err := a.client.NewRequest("GET", u, nil, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var awardEmoji []*AwardEmoji
```
svanharmelen: If the method receiver is updated to be `s`, then this could be `as` (or `a` when singular) .This is a convention used in many other locations in this package

Lorac:
@svanharmelen Style fixes are in. You can check them out.

svanharmelen:
LGTM! Thanks again @Lorac!

Issue #315, opened 2018-02-07
Support for Issue links API


JohnnyQQQQ:
Currently the Issue links API is not supported.

https://docs.gitlab.com/ee/api/issue_links.html

Lorac:
@JohnnyQQQQ I'll put a PR asap

DanMayor:
Is there documentation that shows how to use this package?  Such as for ListIssues, what are opt and options supposed to be?

svanharmelen:
@DanMayor the answer is in both the code of this package and the GitLab docs. There are also a few [examples](https://github.com/xanzy/go-gitlab/tree/master/examples) that show the basic way to use this package.

Specifically for the [ListIssues](https://github.com/xanzy/go-gitlab/blob/master/issues.go#L124-L141) call, you can check the related [options](https://github.com/xanzy/go-gitlab/blob/master/issues.go#L106-L122) struct that shows all fields you can set (that's the `opt` part).

The valid `options` you can pass in are described [here](https://github.com/xanzy/go-gitlab/blob/master/gitlab.go#L678-L702). But they are not mandatory and can also be omitted.

Hope this helps...

PR #317, opened 2018-02-08
Add Issue Links API

Lorac:
I'm not too sure about the `IssueLinksResult` struct name, if you have better I would gladly change it.

When you look at the documentation, it's outdated... When I actually do the calls on my side there are a lot more properties inside the response.

That's why `IssueLinks` has more fields than the documentation.

Fixes : #315

### DIFF
--- a/issue_links.go
+++ b/issue_links.go
@@ -1,0 +1,122 @@
+//
+// Copyright 2017, Arkbriar
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package gitlab
+
+import (
+	"fmt"
+	"net/url"
+)
+
+// IssueLinksService handles communication with the issue relations related methods
+// of the GitLab API.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/issue_links.html
+type IssueLinksService struct {
+	client *Client
+}
+
+type IssueLink struct {
+	SourceIssue *Issue `json:"source_issue"`
+	TargetIssue *Issue `json:"target_issue"`
+}
+
+// ListIssueRelations gets a list of related issues of a given issue,
+// sorted by the relationship creation datetime (ascending).
+//
+// Issues will be filtered according to the user authorizations.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/issue_links.html#list-issue-relations
+func (s *IssueLinksService) ListIssueRelations(pid interface{}, issueIID int, options ...OptionFunc) ([]*Issue, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/issues/%d/links", url.QueryEscape(project), issueIID)
+
+	req, err := s.client.NewRequest("GET", u, nil, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var is []*Issue
+	resp, err := s.client.Do(req, &is)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return is, resp, err
+}
+
+type CreateIssueLinkOptions struct {
+	TargetProjectID *string `json:"target_project_id"`
+	TargetIssueIID  *string `json:"target_issue_iid"`
+}
+
+// CreateIssueLink creates a two-way relation between two issues.
+// User must be allowed to update both issues in order to succeed.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/issue_links.html#create-an-issue-link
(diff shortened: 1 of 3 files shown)


issue_links.go
```diff
@@ -0,0 +1,181 @@
...
+	TimeStats        *TimeStats `json:"time_stats,omitempty"`
+	Links            struct {
+		Self       string `json:"self,omitempty"`
+		Notes      string `json:"notes,omitempty"`
+		AwardEmoji string `json:"award_emoji,omitempty"`
+		Project    string `json:"project,omitempty"`
+	} `json:"_links,omitempty"`
+	Subscribed  bool `json:"subscribed,omitempty"`
+	IssueLinkID int  `json:"issue_link_id,omitempty"`
+}
```
svanharmelen: All the `omitempty`'s in these tags should be removed. They make no sense for structs that are used for responses, only struct for requests would need them (so the options structs).

issue_links.go
```diff
@@ -0,0 +1,181 @@
...
+
+type IssueLinksResult struct {
+	SourceIssue *IssueLinks `json:"source_issue"`
+	TargetIssue *IssueLinks `json:"target_issue"`
+}
+
+type IssueLinksOptions struct {
+	TargetProjectID *string `json:"target_project_id"`
+	TargetIssueIID  *string `json:"target_issue_iid"`
```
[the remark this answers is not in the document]
svanharmelen: This should be called `CreateIssueLinkOptions` and be placed above the `CreateIssueLink` method. Additionally you forgot to add the `omitempty` part in the `json` tags.

issue_links.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// ListIssueRelations Get a list of related issues of a given issue,
```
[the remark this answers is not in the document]
svanharmelen: Without the capital :wink: so `gets a list of...` That goes for all comments by the way.

issue_links.go
```diff
@@ -0,0 +1,181 @@
...
+		Project    string `json:"project,omitempty"`
+	} `json:"_links,omitempty"`
+	Subscribed  bool `json:"subscribed,omitempty"`
+	IssueLinkID int  `json:"issue_link_id,omitempty"`
+}
+
+type IssueLinksResult struct {
+	SourceIssue *IssueLinks `json:"source_issue"`
+	TargetIssue *IssueLinks `json:"target_issue"`
+}
```
svanharmelen: Is this really how the output is given? As the docs seem to indicate that this call would return a list of issue links related to the given issue.

issue_links.go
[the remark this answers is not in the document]
Lorac: These options are mandatory for the call to work if they are not passed it's an error message you'll get. That's why I haven't put `omitempty`
`

issue_links.go
Lorac: Yes, I tested it on my end it's only used for the `POST` and `DELETE` endpoints. The `GET` endpoints returns a list of `IssueLink`

issue_links.go
Lorac: I have been coding Go as a fun thing to do recently and just realized that I didn't understand how that `omitempty` parameter was really working. Thanks for the feedback!

svanharmelen: No problem! And hope you enjoy Go 👍

issue_links.go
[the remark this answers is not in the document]
svanharmelen: 👍

issue_links.go
```diff
@@ -0,0 +1,181 @@
...
+		Self       string `json:"self"`
+		Notes      string `json:"notes"`
+		AwardEmoji string `json:"award_emoji"`
+		Project    string `json:"project"`
+	} `json:"_links"`
+	Subscribed  bool `json:"subscribed"`
+	IssueLinkID int  `json:"issue_link_id"`
+}
+
+type IssueLinksResult struct {
```
svanharmelen: Since we now called the other struct `IssueLink`, maybe call this one just `IssueLinks`?

issue_links.go
svanharmelen: Ok... I see it as well now when looking at the docs of the `POST` and `DELETE` calls again.

issue_links.go
svanharmelen: Hmm... Looking at it again. Is what we call `IssueLink` here, not just an `Issue` (which already exists but is missing some fields that we should add)? So I think we should delete `IssueLink` and use the `Issue` instead. And then `IssueLinksResult` can be just `IssueLink`. Right?

svanharmelen: Looks like only these fields need to be added to the `Issue` struct:

```
Assignee
Upvotes
DownVotes
Weight
Links
IssueLinkID
```

And all those fields are actually returned when retrieving an issue, so this seems to be the correct solution.

Lorac:
I pushed the changes to add the new fields in `Issue`, I haven't tested it yet.

What about the mandatory options for `CreateIssueLinkOptions` they are options, but mandatory... Is there any example of that anywhere else in the code?

issue_links.go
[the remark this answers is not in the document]
Lorac: Is there a new to warn the user in another way. Could we force to pass the parameters?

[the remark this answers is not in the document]
Lorac: and then make a map of these parameters, and pass the map as options? Is this viable?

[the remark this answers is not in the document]
svanharmelen: The only solution I can think of, is to make the fields part of the function signature and then use an anonymous struct in the function body.

But I wouldn't go that route. There are more API calls that have required options. I think that is fine and that the developers using this package should have some insights in what they are doing.

svanharmelen:
Thanks again @Lorac!

PR #322, opened 2018-02-10
Add go 1.9.x to travis builds

Lorac:
I'm using 1.9.x at home, maybe you would like to support that version too?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,3 +3,4 @@ language: go
 go:
   - 1.7.x
   - 1.8.x
+  - 1.9.x


zaquestion:
Depending on what @svanharmelen thinks, may be a good move to also test against `tip`/`1.x`. Not sure how far back we support either.

svanharmelen:
I'm also using 1.9.x (don't we all :wink:) Looks good for now, but once 1.10 comes out later this month we should add that one and also use tip (and then maybe drop 1.7). But for now let's start with this 👍

PR #325, opened 2018-02-12
Fix Namespace struct

tmaczukin:
Currently execution of `Client.Namespaces.GetNamespace()` is failing with:

```
json: cannot unmarshal number into Go struct field Namespace.parent_id of type string
```

It's that because `parent_id` is sent as JSON integer, not string. This PR updates `Namespace` struct to reflect this.

### DIFF
--- a/namespaces.go
+++ b/namespaces.go
@@ -37,7 +37,7 @@ type Namespace struct {
 	Path                        string `json:"path"`
 	Kind                        string `json:"kind"`
 	FullPath                    string `json:"full_path"`
-	ParentID                    string `json:"parent_id"`
+	ParentID                    int    `json:"parent_id"`
 	MembersCountWithDescendants int    `json:"members_count_with_descendants"`
 }
 


svanharmelen:
Thanks @tmaczukin! I willing to merge this one, but it seems the docs are not up to date as they clearly show a `string` in the example output: https://docs.gitlab.com/ce/api/namespaces.html#list-namespaces

image

So it would be nice to double check/verify this one with Gitlab first, to make sure we are updating the struct correctly now. Is that something you could do? A way to do this could also be looking in the Gitlab source code to see that is in there (what is actually returned in there)?

tmaczukin:
I't definitely a typo in documentation, since the `parent_id` value is stored in database as integer. So it's sent through API either as integer (so `"parent_id": 9`) or `null` if empty (so `"parent_id": null`). Both examples are wrong in the documentation.

I've prepared a fix for the documentation: https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/17085.

svanharmelen:
Cool... Then let's get this one in 👍  Thanks!

PR #331, opened 2018-02-15
Add Sidekiq API

Lorac:
The `QueueMetrics` struct seems to contain some number of fields. In my instance there were a hundred different ones, that's why I chose to use a map.

### DIFF
--- a/sidekiq_metrics.go
+++ b/sidekiq_metrics.go
@@ -1,0 +1,154 @@
+//
+// Copyright 2018, Sander van Harmelen
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package gitlab
+
+import "time"
+
+// SidekiqService handles communication with the sidekiq service
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/sidekiq_metrics.html
+type SidekiqService struct {
+	client *Client
+}
+
+// QueueMetrics represents the GitLab sidekiq queue metrics.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/sidekiq_metrics.html#get-the-current-queue-metrics
+type QueueMetrics struct {
+	Queues map[string]struct {
+		Backlog int `json:"backlog"`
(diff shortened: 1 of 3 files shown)


sidekiq_metrics.go
```diff
@@ -0,0 +1,149 @@
...
+//
+
+package gitlab
+
+import "time"
+
+// SidekiqService handles communication with the application SettingsService
+// related methods of the GitLab API.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/settings.html
```
svanharmelen: This should be `https://docs.gitlab.com/ce/api/sidekiq_metrics.html`

sidekiq_metrics.go
```diff
@@ -0,0 +1,149 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package gitlab
+
+import "time"
+
+// SidekiqService handles communication with the application SettingsService
```
svanharmelen: This should be: `SidekiqService handles communication with the sidekiq service`

sidekiq_metrics.go
```diff
@@ -0,0 +1,149 @@
...
+
+// QueueMetricsStat represents the QueueMetrics embedded struct.
+type QueueMetricsStat struct {
+	Backlog int `json:"backlog"`
+	Latency int `json:"latency"`
+}
+
+// QueueMetrics represents the GitLab sidekiq queue metrics.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/sidekiq_metrics.html#get-the-current-queue-metrics
```
svanharmelen: Please try to keep comment lines blow 80 chars. So in this case this should be:

```
// GitLab API docs:
// https://docs.gitlab.com/ce/api/sidekiq_metrics.html#get-the-current-queue-metrics
```

sidekiq_metrics.go
```diff
@@ -0,0 +1,149 @@
...
+type QueueMetrics struct {
+	Queues map[string]QueueMetricsStat `json:"queues"`
+}
```
svanharmelen: Using a map is fine here, but I would suggest to just define the `QueueMetricsStat` fields inline here...

```
Queues map[string]struct {
  Backlog int `json:"backlog"`
  Latency int `json:"latency"`
} `json:"queues"`
```

sidekiq_metrics.go
```diff
@@ -0,0 +1,149 @@
...
+
+// GetQueueMetrics lists information about all the registered queues, their backlog and their latency.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/sidekiq_metrics.html#get-the-current-queue-metrics
```
svanharmelen: Think about the length... Goes for all comments of course.

Lorac:
Pushed fixes

svanharmelen:
Nice! Thx!

PR #335, opened 2018-02-17
Add search API

Lorac:
This API will be introduced in GitLab 10.5, as far as I know, the version of Gitlab we are using at work is 10.4.3, so it is not released yet.

But the calls work for the live endpoint (https://gitlab.com/api/v4/search)

### DIFF
--- a/search.go
+++ b/search.go
@@ -24,66 +24,77 @@ import (
 // SearchService handles communication with the search related methods of the
 // GitLab API.
 //
-// GitLab API docs: https://docs.gitlab.com/ee/api/search.html
+// GitLab API docs: https://docs.gitlab.com/ce/api/search.html
 type SearchService struct {
 	client *Client
 }
 
+// SearchOptions represents the available options for all search methods.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/search.html
+type SearchOptions ListOptions
+
+type searchOptions struct {
+	SearchOptions
+	Scope  string `url:"scope" json:"scope"`
+	Search string `url:"search" json:"search"`
+}
+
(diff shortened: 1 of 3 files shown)


search.go
```diff
@@ -0,0 +1,324 @@
...
+
+func (s *SearchService) searchByProject(pid interface{}, scope, query string, opt *ListOptions, result interface{}, options ...OptionFunc) (*Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, err
+	}
+
+	params := make(url.Values)
+	params.Set("scope", scope)
+	params.Set("search", query)
```
zaquestion: Believe we should be creating an options struct for these `params.Set` (here and in the other funcs) and embedding `ListOptions` if appropriate. As I understand the options are encoded into the query when calling `NewRequest`
Ref: https://github.com/xanzy/go-gitlab/blob/master/gitlab.go#L428

Lorac: I didn't want the user to create a struct when you only want to search for a string. I wanted to make it easy for the user. I looked up go-github, they do it for the query 

https://github.com/google/go-github/blob/master/github/search.go#L186

I thought it was a pretty good idea... but I wish I could get rid of the `Set` for the `scope` parameter.

Lorac: I could hardcode the `scope` parameter in the url, but then I would've to do the params thingy for the query as github does it. I think would be not so bad.

Lorac: @svanharmelen any idea?

svanharmelen: Let me get back on this early next week! Sorry for the delay with this one...

svanharmelen:
Thanks @Lorac!

Issue #336, opened 2018-02-19
ListLabels in labels.go does not accept ListOptions, cannot deal with pagination


cdaigle:
We work with a large number of labels and need to be able to pass in pagination options to ListLabels.   From what I can see the ListLabels function calls s.client.NewRequest which accepts a ListOptions but it is hardcoded to nil.  Can we update ListLabels to take in a listOptions and pass it through?

cdaigle:
That was fast, thanks for the quick turnaround!

PR #337, opened 2018-02-19
Add missing ListOptions to ListLabels

Lorac:
Fixes : #336

### DIFF
--- a/labels.go
+++ b/labels.go
@@ -48,17 +48,22 @@ func (l Label) String() string {
 	return Stringify(l)
 }
 
+// ListLabelsOptions represents the available ListLabels() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/labels.html#list-labels
+type ListLabelsOptions ListOptions
+
 // ListLabels gets all labels for given project.
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/labels.html#list-labels
-func (s *LabelsService) ListLabels(pid interface{}, options ...OptionFunc) ([]*Label, *Response, error) {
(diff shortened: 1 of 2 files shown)


labels.go
```diff
@@ -51,14 +51,14 @@ func (l Label) String() string {
 // ListLabels gets all labels for given project.
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/labels.html#list-labels
-func (s *LabelsService) ListLabels(pid interface{}, options ...OptionFunc) ([]*Label, *Response, error) {
+func (s *LabelsService) ListLabels(pid interface{}, opt *ListOptions, options ...OptionFunc) ([]*Label, *Response, error) {
```
svanharmelen: I try to prevent creating options structs that are randomly applicable to different functions. So where options are applicable, I always make sure there is a dedicated options struct. If the struct is the same it can just be a named type referencing the type you want. So in this case it could be written as:

```
// ListLabelOptions represents the available ListLabel() options.
//
// GitLab API docs: https://docs.gitlab.com/ce/api/labels.html#list-labels
type ListLabelOptions ListOptions
``` 

It's one additional line of code, but it makes things really clear IMO...

Lorac: Looks clean to me, only one problem tho. It doesn't seem to be the case everywhere in the library right now. Maybe I could do a PR to create a referenced dedicated type for each ListOption?

Lorac: Haha my comment is a bit late, just saw the PR you did to fix it all 👍

svanharmelen: Indeed... So if you can update this PR as well, then this one can be merged as well 👍

Lorac:
@svanharmelen Pushed :)

svanharmelen:
Thanks! LGTM!

PR #339, opened 2018-02-19
Add missing pagination option to couple of endpoints

Lorac:
I looked the up the documentation and the gitlab source code to confirm that they do support the pagination for those endpoints.

### DIFF
--- a/boards.go
+++ b/boards.go
@@ -60,14 +60,14 @@ func (b BoardList) String() string {
 // ListIssueBoards gets a list of all issue boards in a project.
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/boards.html#project-board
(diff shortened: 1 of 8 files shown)


Lorac:
I dissected the source code of Gitlab to check if the result they were returning was paginated. So I am 100% sure those endpoints are paginated.

Lorac:
There is only one in there I shouldn't have implemented tho... GetCommitDiff. The options are ignored until it's implemented by gitlab.

https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/17203

svanharmelen:
I'll take this one as your doing so much already :wink: So pulling this one in together with some optimizations in PR #341

svanharmelen:
It's in, thanks @Lorac!

PR #343, opened 2018-02-19
Added Confidential Issue Hook

jbguerraz:
Atdds the confidential issue hook field to the project and service structures

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -665,19 +665,20 @@ type ProjectMember struct {
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/projects.html#list-project-hooks
 type ProjectHook struct {
-	ID                    int        `json:"id"`
-	URL                   string     `json:"url"`
-	ProjectID             int        `json:"project_id"`
-	PushEvents            bool       `json:"push_events"`
-	IssuesEvents          bool       `json:"issues_events"`
-	MergeRequestsEvents   bool       `json:"merge_requests_events"`
-	TagPushEvents         bool       `json:"tag_push_events"`
-	NoteEvents            bool       `json:"note_events"`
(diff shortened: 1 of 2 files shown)


svanharmelen:
@jbguerraz I don't see that field in the docs... Can you show me something that confirms this field is used/available? Thx!

Lorac:
Maybe not in the doc, but it's in the source code at least

https://gitlab.com/gitlab-org/gitlab-ce/blob/master/spec/factories/project_hooks.rb

svanharmelen:
Check! Thanks for the confirmation @Lorac!

jbguerraz:
My bad! this doesn't works :(

Gitlab is managing the confidential issues hooks in a _special_ way; at least those hooks are not exposed to the API but only available using the UI (so even a hook GET will not show the `confidential_issues_events` field).

Opened an issue : https://gitlab.com/gitlab-org/gitlab-ce/issues/43425

Meanwhile, that can be reverted.

svanharmelen:
Reverted for now 👍 (#347)

jbguerraz:
FYI: Some progress on https://gitlab.com/gitlab-org/gitlab-ce/issues/43425

jbguerraz:
@svanharmelen https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/17370 it's merged! it's gonna be part of the next release, maybe we could apply that patch as anyway it won't hurts with current gitlab state  (it's just silently ignored now). What do you think ? :)

svanharmelen:
Thanks for following up @jbguerraz! If you can recreate the PR, I will merge it again :wink:

jbguerraz:
@svanharmelen done! see https://github.com/xanzy/go-gitlab/pull/352

Issue #117, opened 2017-01-23
Pagination


lyda:
This might be something people more used to REST wrappers know, but for me the [Response](https://godoc.org/github.com/xanzy/go-gitlab#Response) docs are incomplete. What is the behaviour of `NextPage`? Should I assume there's more data if `NextPage` is not zero?

Also the GitLab [pagination](http://docs.gitlab.com/ce/api/README.html#pagination) docs indicate there's also `X-Page` and `X-Total-Pages` headers sent back. Would it be helpful to include them?

I note that the GitHub API does not have such fields so perhaps you're trying to stay similar to theirs?

If not, might it be an idea to make iterators that hide the pagination?

jangrewe:
Yes, i just got bitten by this, too. How can this be handled?

jakubknejzlik:
Assuming from this source code:
https://github.com/xanzy/go-gitlab/blob/master/gitlab.go#L428

The XPage is populated from `Link` header which is in format:
```
<https://git.example.com/api/v4/users?active=false&blocked=false&external=false&page=1&per_page=20>; rel="prev",
...
```
So if theres no such link with `rel="next"` the NextPage *should* be 0.

Lorac:
@svanharmelen I guess we could close this one?

svanharmelen:
I agree... I'll close this one for now. Of course it can be reopened or a new issue can be opened if any additional questions pop up. Thx!

PR #345, opened 2018-02-20
Add how many results and pages there are for a paginated api

Lorac:
I'm not sure about the 4 other headers

x-next-page → 2
x-page →1
x-per-page →1
x-prev-page → (empty)

They are provided by headers in the response of the API, maybe instead of parsing the header `link` just use the provided headers by the API?

Relates to : #117

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -38,6 +38,13 @@ const (
 	libraryVersion = "0.2.0"
 	defaultBaseURL = "https://gitlab.com/api/v4/"
 	userAgent      = "go-gitlab/" + libraryVersion
+
+	headerNextPage     = "X-Next-Page"
+	headerTotalPages   = "X-Total-Pages"
+	headertotalResults = "X-Total"
+	headerCurrentPage  = "X-Page"
+	headerPerPage      = "X-Per-Page"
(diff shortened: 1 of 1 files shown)


gitlab.go
```diff
@@ -508,6 +510,14 @@ func newResponse(r *http.Response) *Response {
 // populatePageValues parses the HTTP Link response headers and populates the
 // various pagination link values in the Response.
 func (r *Response) populatePageValues() {
+	if totalPages, ok := r.Response.Header["X-Total-Pages"]; ok && len(totalPages) > 0 {
+		r.TotalPages, _ = strconv.Atoi(totalPages[0])
```
svanharmelen: I would suggest to use `r.Response.Header.Get("X-Total-Pages")` instead... So something like this:

```
if totalPages := r.Response.Header.Get("X-Total-Pages"); totalPages != "" {
  r.TotalPages, _ = strconv.Atoi(totalPages[0])
}
```

Lorac:
I would forget about the link part, what do you think?

svanharmelen:
@Lorac yes, that was what I meant... Rewrite the `link` part to instead just use the individual headers. So LGTM now! Thx...

PR #350, opened 2018-02-23
pages domain verification

Jonarod:
Adding new fields to pages domain struct after new 2-step domain ownership verification.

### DIFF
--- a/pages_domains.go
+++ b/pages_domains.go
@@ -18,10 +18,13 @@ type PagesDomainsService struct {
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/pages_domains.html
 type PagesDomain struct {
-	Domain      string `json:"domain"`
-	URL         string `json:"url"`
-	ProjectID   int    `json:"project_id"`
-	Certificate struct {
+	Domain           string     `json:"domain"`
+	URL              string     `json:"url"`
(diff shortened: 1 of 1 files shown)


Lorac:
@svanharmelen Is there a "commit message" rules if so, I would suggest to change it to something more appropriate. 
```
Add missing pages domain verification fields.

Some description why they were missing...
```

svanharmelen:
No, there are no commit message rules. And personally I don’t think we need one... 

But maybe we could add a few suggestions to a [PR template](https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/),  think that should be more than sufficiant.

zaquestion:
fwiw, I'm personally quite fond of 
```
[Issue-key] (context) what / topic statement

Why in body
```
Yeah though don't think it needs to be a rule and issue/PR templates seem like a great idea lot more  action in the repo lately in particular.

Jonarod:
@svanharmelen check this [solved issue](https://gitlab.com/gitlab-com/support-forum/issues/3002#note_60334972). If you test the "create pages domain" API endpoint, you'll see these new fields, hence the new struct.

svanharmelen:
Check... Thanks again @Jonarod! LGTM!

PR #353, opened 2018-03-06
add ListOptions to GetAllImpersonationTokensOptions

45cali:
I have a use case where I need to use pagination


svanharmelen:
Thanks @45cali! LGTM!

PR #354, opened 2018-03-17
Make ListTagsOptions work correctly with ListOptions

Bouwdie:
The ListTagsOptions was not defined correctly with an invalid documentation reference. Fixed this and added a test suite for tags (like the product test)

### DIFF
--- a/tags.go
+++ b/tags.go
@@ -49,7 +49,7 @@ func (t Tag) String() string {
 // ListTagsOptions represents the available ListTags() options.
 //
 // GitLab API docs:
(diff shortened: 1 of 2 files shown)


svanharmelen:
@Bouwdie what makes you think `ListTagOptions` was defined incorrectly? Seem perfectly valid and correct to me...

Bouwdie:
@svanharmelen You are absolutely right! i reverted that change and updated the test. Maybe the test is still of some use?

svanharmelen:
Sure thing! Tests are always welcome 👍🏻 Thanks!

PR #355, opened 2018-03-17
Add missing last_activity_on value for a user

nogs:
The user object is missing the "last_activity_on" field that is available for admins and when requesting the CurrentUser().

The documentation, https://gitlab.com/help/api/users.md#for-normal-users-1, shows that it has the format "YYYY-MM-DD" which time.Time doesn't appear to handle by default so I had to add a new type and marshal functions for it.   I had a quick fix to get something working that just used a string, but I think it is probably better handled this way.

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -117,6 +117,11 @@ func (t *ISOTime) UnmarshalJSON(data []byte) error {
 	return err
 }
 
+// String implements the Stringer interface
+func (t ISOTime) String() string {
+	return time.Time(t).Format(iso8601)
+}
+
 // NotificationLevelValue represents a notification level.
 type NotificationLevelValue int
 
--- a/users.go
+++ b/users.go
@@ -50,6 +50,7 @@ type User struct {
 	ExternUID        string          `json:"extern_uid"`
 	Provider         string          `json:"provider"`
 	ThemeID          int             `json:"theme_id"`
+	LastActivityOn   *ISOTime        `json:"last_activity_on"`
 	ColorSchemeID    int             `json:"color_scheme_id"`
 	IsAdmin          bool            `json:"is_admin"`
 	AvatarURL        string          `json:"avatar_url"`


svanharmelen:
@nogs there already is a type for this called [`ISOTime`](https://github.com/xanzy/go-gitlab/blob/faf7c98e28a87350f4259dc34af2e4ad5993559b/gitlab.go#L87). So if you could use that type instead, that would be great... Thanks!

nogs:
OK, let me have a look and update.

nogs:
Any objections to adding a format / string method to that class to make it easier to print?

nogs:
Annoyingly I recall seeing that type in the week now and completely forgot about it...

svanharmelen:
It's not a class, it's a type (Go doesn't have classes :wink:). I don't see added value for the `Format` method, but maybe the `String` method could be handy...

nogs:
Jumping between languages too much...

String on it's own should be fine.

gitlab.go
```diff
@@ -117,6 +117,12 @@ func (t *ISOTime) UnmarshalJSON(data []byte) error {
 	return err
 }
 
+// implement String method for ISOTime
```
svanharmelen: To satisfy `golint` (and other Go linters) the comment should start with the name of the method. So in this case `String implements the Stringer interface` seems like a good comment...

gitlab.go
```diff
@@ -117,6 +117,12 @@ func (t *ISOTime) UnmarshalJSON(data []byte) error {
...
+func (j ISOTime) String() string {
```
svanharmelen: Again to satisfy linters, the receiver name should be consistent with other declarations. So in this case it should be `(t ISOTime)`.

gitlab.go
```diff
@@ -117,6 +117,12 @@ func (t *ISOTime) UnmarshalJSON(data []byte) error {
...
+	t := time.Time(j)
+	return t.Format(iso8601)
```
svanharmelen: This could be written as just `return time.Time(t).Format(iso8601)`

nogs:
Changes made.

svanharmelen:
LGTM! Thanks @nogs!

PR #359, opened 2018-03-20
add tests for validate

pityonline:
* valid case
* invalid case

Posting empty content was marked as **TODO** because [GitLab API](https://docs.gitlab.com/ce/api/lint.html) returns deferent type.

### DIFF
--- a/validate_test.go
+++ b/validate_test.go
@@ -7,58 +7,53 @@ import (
 	"testing"
 )
 
-// TODO: better test posting empty string
 func TestValidate(t *testing.T) {
-	validContent := `
-	build1:
-		stage: build
-		script:
-			- echo "Do your build here"`
-
-	invalidContent := `
-	build1:
-		- echo "Do your build here"`
-
-	validRes := `{
-			"status": "valid",
-			"errors": []
-		}`
-
-	invalidRes := `{
-			"status": "invalid",
-			"errors": [
-				"error message when content is invalid"
-			]
-		}`
-
-	wantValid := &LintResult{
-		Status: "valid",
-		Errors: []string{},
-	}
-
-	wantInvalid := &LintResult{
-		Status: "invalid",
-		Errors: []string{"error message when content is invalid"},
-	}
-
 	testCases := []struct {
-		desc    string
-		content string
-		res     string
-		want    *LintResult
+		description string
+		content     string
+		response    string
+		want        *LintResult
 	}{
-		{"valid", validContent, validRes, wantValid},
-		{"invalid", invalidContent, invalidRes, wantInvalid},
+		{
+			description: "valid",
+			content: `
+				build1:
+					stage: build
+					script:
+						- echo "Do your build here"`,
+			response: `{
+				"status": "valid",
+				"errors": []
+			}`,
+			want: &LintResult{
+				Status: "valid",
+				Errors: []string{},
+			},
+		},
+		{
+			description: "invalid",
+			content: `
+				build1:
+					- echo "Do your build here"`,
+			response: `{
+				"status": "invalid",
+				"errors": ["error message when content is invalid"]
+			}`,
+			want: &LintResult{
+				Status: "invalid",
+				Errors: []string{"error message when content is invalid"},
+			},
+		},
 	}
 
 	for _, tc := range testCases {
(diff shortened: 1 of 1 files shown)


validate_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"testing"
+)
+
+// XXX: move this func out if necessary
+// content read a file, return its content as string
+func content(f string) string {
+	b, err := ioutil.ReadFile(f)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
```
svanharmelen: `log.Fatal(err)` does the same as `fmt.Println(err)` and `os.Exit(1)` do together. But maybe just return an error here as well and use `t.Fatal(err)` in the calling function instead?

validate_test.go
```diff
@@ -0,0 +1,97 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"reflect"
+	"testing"
+)
+
+// XXX: move this func out if necessary
```
svanharmelen: This line should be removed...

validate_test.go
```diff
@@ -0,0 +1,97 @@
...
+			]
+		}`
+
+	mux.HandleFunc("/ci/lint", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "POST")
+		fmt.Fprintf(w, res)
+	})
+
+	testFunc := func(t *testing.T, got, want *LintResult) {
+		t.Helper()
```
svanharmelen: `t.Helper()` is only available since Go 1.9, not sure if we should already use it here...

validate_test.go
```diff
@@ -0,0 +1,97 @@
...
+		e := make([]string, 1)
+		e[0] = "variables config should be a hash of key value pairs"
+
+		want := &LintResult{
+			Status: "invalid",
+			Errors: e,
+		}
+
+		testFunc(t, got, want)
+	})
```
svanharmelen: I guess a better approach would be to use table driven tests to prevent this code duplication.

validate_test.go
```diff
@@ -0,0 +1,97 @@
...
+				"variables config should be a hash of key value pairs"
+			]
+		}`
+
+	mux.HandleFunc("/ci/lint", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "POST")
+		fmt.Fprintf(w, res)
+	})
+
+	testFunc := func(t *testing.T, got, want *LintResult) {
```
zaquestion: @svanharmelen Not sure how you feel about bringing in a test dependency. Would suggest replacing `testFunc` with a call to testify/assert
```
assert.Equal(t, expected, actual)
```
Ref: https://github.com/stretchr/testify

validate_test.go
zaquestion: Suggest trying to avoid `os.Exit` (and `log.Fatal`) in tests. The situation might be better in go1.10, but generally speaking this breaks generating coverage when testing in these cases. If its purely in a test context though as @svanharmelen mentions `t.Fatal` should be safe. having this return an error seems solid

zaquestion: Might even want to just inline this function in the test after changing the tests to be table driven

validate_test.go
zaquestion: Theres a section on table driven tests with subtests here: https://blog.golang.org/subtests

zaquestion: Possibly better, here are the table driven tests that `lab` is using to test the lint/validate API. https://github.com/zaquestion/lab/blob/master/internal/gitlab/gitlab_test.go#L45

pityonline:
@svanharmelen @zaquestion Thanks for the review. I lenarned a lot, and will refactor it in a day.

validate_test.go
svanharmelen: Not sure. Not really a fan of using external packages for writing tests (especially for client packages like this).

About the `testFunc` specifically, I think it should be factored out once using table driven tests.

svanharmelen:
@pityonline please see https://github.com/pityonline/go-gitlab/pull/1

svanharmelen:
Thanks @pityonline! LGTM!

PR #361, opened 2018-03-22
Fix type for commit.duration in BuildEvent

nyxi:
Somewhere between v10.5.2 and v10.5.6 of Gitlab CE `commit.duration` in BuildEvent changed from an empty string to an integer.


svanharmelen:
Thanks @nyxi! LGTM!

PR #373, opened 2018-04-06
add support for creating releases in tags api

starblind:


svanharmelen:
LGTM! Only thing I wonder about, is if we should also add the update release call in this PR to make it complete? What do you think?

svanharmelen:
Thanks @starblind! LGTM!

Issue #369, opened 2018-04-03
Delete Project Milestone method missing


johanbrandhorst:
The documentation lists the ability to delete a project milestone: https://docs.gitlab.com/ce/api/milestones.html#delete-project-milestone. This does not appear to be implemented currently. This issue should track the implementation of this method.

Lorac:
@svanharmelen I started implementing it, but I saw that `group_milestones` and my `project_milestones` would pretty much duplicate code. The only difference is the URL We could unify the code under `milestones.go`, so we can do something like : 

```
// ListProjectMilestones gets a list of a project's milestones.
//
// GitLab API docs:
// https://docs.gitlab.com/ce/api/members.html#list-all-members-of-a-group-or-project
func (s *MilestoneService) ListProjectMilestones(pid interface{}, opt *ListMilestonesOptions, options ...OptionFunc) ([]*Milestone, *Response, error) {
	project, err := parseID(pid)
	if err != nil {
		return nil, nil, err
	}
	u := fmt.Sprintf("projects/%s/milestones", url.QueryEscape(project))
	return listMilestones(u, opt, options)
}

// ListGroupMilestones gets a list of a groups's milestones.
//
// GitLab API docs:
// https://docs.gitlab.com/ee/api/group_milestones.html#list-group-milestones
func (s *MilestoneService) ListGroupMilestones(gid interface{}, opt *ListMilestonesOptions, options ...OptionFunc) ([]*Milestone, *Response, error) {
	group, err := parseID(gid)
	if err != nil {
		return nil, nil, err
	}
	u := fmt.Sprintf("groups/%s/milestones", url.QueryEscape(project))
	return listMilestones(u, opt, options)
}

func (s *MilestoneService) listMilestones(url string, opt *ListMilestonesOptions, options ...OptionFunc) ([]*Milestone, *Response, error) {
	req, err := s.client.NewRequest("GET", url, opt, options)
	if err != nil {
		return nil, nil, err
	}

	var pm []*Milestones
	resp, err := s.client.Do(req, &pm)
	if err != nil {
		return nil, resp, err
	}

	return pm, resp, err
}
```

What do you think?

johanbrandhorst:
I actually think we intentionally duplicate a lot of similar code in order to make changes down the line easier. I could be wrong though.

svanharmelen:
[...]

PR #374, opened 2018-04-10
Merge Project/Group to MilestoneService and add missing milestone endpoint

Lorac:
I merge both Project and Group milestone under the same service `MilestoneService`, thus causing some breaking changes to the current API

Fixes: #369

### DIFF
--- a/milestones.go
+++ b/milestones.go
@@ -247,3 +247,21 @@ func (s *MilestonesService) GetMilestoneMergeRequests(pid interface{}, milestone
 
 	return mr, resp, err
 }
+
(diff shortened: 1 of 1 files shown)


svanharmelen:
Hi @Lorac! Thanks for this PR, but I would really like to keep the `GroupMilestoneService`. I try to follow the entries in the docs as much as possible, and in there project and group milestones are two different entries:

image

So that means I think we should have these two services as well...

svanharmelen:
Thanks @Lorac! I'm going to rename the method to make it inline with the rest, but I'll merge this as is now 👍  Thx!

PR #376, opened 2018-04-17
Add ListOptions to ListTreeOptions

nogs:
I think ```ListOptions``` is missing from ```ListTreeOptions```.  

The API does seem to page the response when you have a lot of files so probably worth adding the option to the struct.


svanharmelen:
Thanks @nogs! LGTM!

Issue #370, opened 2018-04-04
Add support for new project import/export API


brendano86:
GitLab recently released an API to import/export projects.

Proposal would be to add support to `go-gitlab` for this: https://docs.gitlab.com/ee/api/project_import_export.html

svanharmelen:
Not sure if/when I have time for this myself, but seems to me we should indeed add support for this API. Of course a PR adding this is very welcome and appreciated!

RoValentim:
I need it too @brendano86, did you start to make this methods?

If yes, how can I help you?

Snehal1112:
Guys anyone working on this enhancement

svanharmelen:
Nope... At least now as far as I know.

Snehal1112:
@svanharmelen I would like to work on this enhancement.

svanharmelen:
Thanks @Shehal1112! Feel free to pick it up and create a PR for it 👍🏻

Have a look at the rest of the existing code an please try to keep things consistent.

Thanks!

matejvelikonja:
@svanharmelen let us know when you start working on it. We would really need this feature.

matejvelikonja:
sorry @svanharmelen, I tagged the wrong person :) I wanted to ask @Snehal1112

Snehal1112:
@matejvelikonja @svanharmelen I have already started work on in I just stuck with import project stuff. 
I will try to commit the change ASAP.

matejvelikonja:
@Snehal1112 could you at least make a PR for the export stuff? I would really need this functionality :)

Snehal1112:
@matejvelikonja sorry for the late replay. last week is very hectic for me. 
I have done with the following things 
- Export status
- Export download
- Import status

Remaining thing 
- Import a file 
    - this feature was some problem and because of that i am able to reproduce this following issue 
        https://gitlab.com/gitlab-org/gitlab-ce/issues/57579
        https://gitlab.com/gitlab-org/gitlab-ce/issues/57585
    which is I think fixed now by gitlab 
- Schedule an export

PR #379, opened 2018-04-19
Add new / missing API methods

brendano86:
Working on #370, but first adding all new methods.


svanharmelen:
Thanks @brendano86! LGTM!

PR #380, opened 2018-04-20
Add closed_by and closed_at to merge_request struct

shanedasilva:


svanharmelen:
Thanks for the PR @shanedasilva! LGTM!

PR #382, opened 2018-04-26
Add unshare project from group

jorcau:
Add possibility to stop sharing a project with a group as described in [Gitlab API Doc](https://docs.gitlab.com/ee/api/projects.html#delete-a-shared-project-link-within-a-group).

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -648,7 +648,7 @@ func (s *ProjectsService) ShareProjectWithGroup(pid interface{}, opt *ShareWithG
 
 // UnshareProjectFromGroup allows to unshare a project from a group.
 //
-// GitLab API docs: https://docs.gitlab.com/ee/api/projects.html#delete-a-shared-project-link-within-a-group
+// GitLab API docs: https://docs.gitlab.com/ce/api/projects.html#delete-a-shared-project-link-within-a-group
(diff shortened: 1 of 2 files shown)


projects.go
```diff
@@ -646,6 +646,24 @@ func (s *ProjectsService) ShareProjectWithGroup(pid interface{}, opt *ShareWithG
 	return s.client.Do(req, nil)
 }
 
+// UnshareProjectFromGroup allows to unshare a project from a group.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/projects.html#delete-a-shared-project-link-within-a-group
```
svanharmelen: Please refer to the `ce` docs whenever possible

projects.go
```diff
@@ -646,6 +646,24 @@ func (s *ProjectsService) ShareProjectWithGroup(pid interface{}, opt *ShareWithG
...
+func (s *ProjectsService) UnshareProjectFromGroup(pid interface{}, groupID int, options ...OptionFunc) (*Response, error) {
```
svanharmelen: I think a name like `DeleteSharedProjectFromGroup` is a much closer match to description in the API docs. And to make methods easy searchable, I prefer to keep as close to the docs as possible.

jorcau:
Thanks for the quick review @svanharmelen!

PR #384, opened 2018-04-27
Fix block/unblock

joerocklin:
While doing some testing with error handling coming back from the block/unblock actions we discovered some issues with how errors were being handled in the gitlab library. The provided tests do not pass with the code in master and do with the changes,

### DIFF
--- a/users.go
+++ b/users.go
@@ -22,6 +22,12 @@ import (
 	"time"
 )
 
+var (
+	UserBlockPreventedError   = fmt.Errorf("Cannot block a user that is already blocked by LDAP synchronization")
+	UserNotFoundError         = fmt.Errorf("User does not exist")
+	UserUnblockPreventedError = fmt.Errorf("Cannot unblock a user that is blocked by LDAP synchronization")
+)
+
 // UsersService handles communication with the user related methods of
 // the GitLab API.
 //
--- a/users.go
+++ b/users.go
@@ -22,9 +22,9 @@ import (
 )
 
 var (
-	UserBlockPreventedError   = fmt.Errorf("Cannot block a user that is already blocked by LDAP synchronization")
-	UserNotFoundError         = fmt.Errorf("User does not exist")
-	UserUnblockPreventedError = fmt.Errorf("Cannot unblock a user that is blocked by LDAP synchronization")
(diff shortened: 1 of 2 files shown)


users.go
```diff
@@ -441,17 +446,17 @@ func (s *UsersService) UnblockUser(user int, options ...OptionFunc) error {
 	}
 
 	resp, err := s.client.Do(req, nil)
-	if err != nil {
+	if err != nil && resp == nil {
```
svanharmelen: If a method returns an error, you should not try to use any of the other returned values. As the error indicates things didn't go as planned, you cannot trust what the other values contain.

So please do not use `resp` if the same call returning `resp` also returned an error. This is common and best practice as far as I know...

svanharmelen:
Thanks @joerocklin! I have one small comment, the rest looks great!

users.go
zaquestion: Haven't tested this out myself, but I'm assuming that the error from the GitLab API is less than useful (as it is the case on some endpoints). Without asserting the resp is nil here I'm not sure what a cleaner way to leverage the `switch` below for informative errors.

Of course if the gilab api is returning useful error messages and they are already getting propagated through the client these switch cases can probably be deleted.

svanharmelen: Looking at the rest of the package, I would almost say it's much more inline to just remove the whole switch statement here. We don't do that in any other location and the error returned is already parsed and put into a custom helpful format in the `Do` method.

We do try to return the received response so users can choose to use/see the raw response (mainly for debugging and investigating in case of issues). So using the `resp` it in this case (even when you got an error) doesn't seem to be a bad choice. But still, do we really want this? Do we want to sugarcoat all the errors we return with custom messages for each individual call there is?

I think we shouldn't do that... What we now return is something like `GET https://gitlab.com/api/v3/some/path: 403 error message from API` (https://github.com/xanzy/go-gitlab/blob/master/gitlab.go#L608-L612) I think that is just fine right? If people using this package want to sugarcoat that a bit more because the want it to be returned to the user or a UI, it's easy to do that in their own package right?

svanharmelen:
It took a while, but I thought it was time to just pull this in :wink: Thanks for the PR @joerocklin!

PR #387, opened 2018-05-04
Auto update libraryVersion when creating releases/tags

zaquestion:
If you'd like, this is is travis job that will auto update the `libraryVersion`
variable when you publish tags/releases on github.

You'll need to setup a secure environment variable with travis:
https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml

You can see a sample run of this job here: https://travis-ci.org/zaquestion/go-gitlab/jobs/374748652
Note: the push fails because env vars are actually tied to the repo (I learned)
and I tried to copy the one I use for lab

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: go
 
+# use containers which run faster and have cache
+sudo: false
+
 go:
   - 1.8.x
   - 1.9.x
@@ -9,10 +12,13 @@ go:
 stages:
   - lint
   - test
+  - name: deploy
+    if: tag IS present
 
 jobs:
   include:
     - stage: lint
+      go: 1.10.x
       script:
         - go get github.com/golang/lint/golint
         - golint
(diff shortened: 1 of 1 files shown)


pityonline:
Maybe it’s better to decouple version printing from gitlab.go first?

zaquestion:
There a downside here that the actual tagged sha doesn't contain the updated library version. I could change this up to be a script that you run that updates the library version first and pushs the ned tag for the release. Let me know how that sounds.

pityonline:
@svanharmelen @zaquestion I tried this last weekend, here's the [demo](https://github.com/pityonline/go-gitlab/commits/master). You can check the [build logs](https://travis-ci.org/pityonline/go-gitlab/builds/375733233), the version is auto updated with the tag.

```
=== RUN   TestVersion
=== RUN   TestVersion/test_when_git_tag_not_exists
=== RUN   TestVersion/test_when_git_tag_exists
--- PASS: TestVersion (0.02s)
    --- PASS: TestVersion/test_when_git_tag_not_exists (0.00s)
    --- PASS: TestVersion/test_when_git_tag_exists (0.00s)
=== RUN   TestValidate
=== RUN   TestValidate/valid
r = go-gitlab/v0.9.0-test-3
=== RUN   TestValidate/invalid
r = go-gitlab/v0.9.0-test-3
--- PASS: TestValidate (0.00s)
    --- PASS: TestValidate/valid (0.00s)
    --- PASS: TestValidate/invalid (0.00s)
PASS
ok  	github.com/pityonline/go-gitlab	0.065s
?   	github.com/pityonline/go-gitlab/examples	[no test files]
```

But I think maybe it's a very common demand for Go project, maybe separate it out as an independent pkg?

pityonline:
@zaquestion You are right. `git` maybe not for every Go user, it's just a `git` way to auto-update the library version with git tags. I don't know if users care about the version, I didn't notice the version string before this PR. 😆

@svanharmelen What do you think?

svanharmelen:
I think the alternative is indeed better... Let's just remove the `libraryVersion`. Agree?

PR #388, opened 2018-05-07
Update ListUsersOptions: add missing parameters

jorcau:
Hi there!

I'd like to add 4 options for listing users ([docs](https://docs.gitlab.com/ce/api/users.html#for-admins)) which seem to be missing:
- `provider`: which makes `extern_uid` unusable since they need to be used together
- `created_after`
- `order_by`
- `sort`

Thanks!


svanharmelen:
Thanks @jorcau! LGMT!

PR #393, opened 2018-05-10
Handle tokens with no expiration

jayschwa:
My GitLab instance returns tokens with no expiration. Without this check, the package spams GitLab with requests for new tokens.

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -357,6 +357,9 @@ func NewBasicAuthClient(httpClient *http.Client, endpoint, username, password st
 		for {
 			client.token = t.AccessToken
 
+			if t.ExpiresIn == 0 {
+				return // Token does not expire
+			}
 			if t.ExpiresIn > 60 {
 				t.ExpiresIn -= 60
 			}


svanharmelen:
Thanks @jayschwa!

svanharmelen:
@jayschwa @zaquestion could you maybe check the update in [this](https://github.com/xanzy/go-gitlab/tree/f-token) branch to see if this also works as expected? Think this should be a little more robust.

jayschwa:
I checked out the branch. My main piece of feedback is that I think you should use Go's OAuth2 package rather than rolling your own. #397 is a PR to do that.

I like the idea of getting the token on an as-needed basis rather than running a background loop that can panic. It does look like the implementation could be susceptible to infinite recursion if getting a token is successful, but the endpoint returns 401 for some other reason.

svanharmelen:
Yeah, I made a note about that to myself as well... This still wasn't a good/complete solution. Just saw your PR, looks great! Thanks!

PR #401, opened 2018-05-17
Some small fixes

pityonline:
Sorry, it was my fault not set exit status for golint, now it's fixed, and `go vet` will be verbose now if anyone has interest in the vet checks.


svanharmelen:
Nice @pityonline! Thanks!

PR #403, opened 2018-05-19
Adds NewSelfHostedClient helper function 

binarix:
This helper function is created so that a new token based client can be created with a self hosted base url supplied.

Hoping that it will be used in https://github.com/jenkins-x/jx

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -342,6 +342,31 @@ func NewClient(httpClient *http.Client, token string) *Client {
 	return client
 }
 
+// NewSelfHostedClient returns a new GitLab API client with provided baseURL.
+// If baseURL does not have a trailing slash, one is added automatically.
+// If a nil httpClient is provided, http.DefaultClient will be used.
+// To use API methods which require authentication, provide a valid private
+// or personal token.
+//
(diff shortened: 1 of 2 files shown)


svanharmelen:
Thanks for your PR @binarix, but unfortunately I must admit I don't really like to add this helper function. I am trying to keep this package as close to the options and features you can find in the Gitlab docs as possible, and a custom helper function like this doesn't really fit into that.

Next to that it seems this PR doesn't seem to solve anything that isn't already possible with just 2 lines of code:

```go
client := gitlab.NewClient(nil, "yourtokengoeshere")
client.SetBaseURL("https://gitlab.com/api/v4")
```

So unless you can convince me otherwise, I'm tempted to close this PR without merging it...

binarix:
Using client.SetBaseURL() requires you to define that the api is v4 which I felt was leaking implementation knowledge.

svanharmelen:
@binarix, that sounds like something we can indeed improve 👍

Please see PR #404 which seems to improve/solve this, but then without adding a custom new helper function. Would this work for you?

binarix:
Closing this PR as modifying the SetBaseURL() is a nicer approach than introducing a new helper function.

PR #405, opened 2018-05-21
Makes SetBaseURL more robust by matching on apiVersionPath suffix and…

binarix:
… adds unit test

Can I suggest matching on apiVersionPath instead of / ?

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -36,7 +36,8 @@ import (
 )
 
 const (
-	defaultBaseURL = "https://gitlab.com/api/v4/"
+	defaultBaseURL = "https://gitlab.com/"
+	apiVersionPath = "api/v4/"
 	userAgent      = "go-gitlab"
 )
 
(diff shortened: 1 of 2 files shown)


svanharmelen:
Well... The idea here was to only add the `api/v4/` part if the user didn't supply a specific path already. If the user already supplied something specific we should not try to do magic on top of that. So only when the user didn't supply a path at all, we should add the minimum needed part (IMHO).

Say for example that someone tries to set `https://gitlab.company.com/api/v3`, we should not just add `api/v4/` right? And even when a user tries to set something faulty like `https://gitlab.company.com/something/wrong` I don't think we should then blindly concatenate the `api/v4/` part.

So I like the unit test, but we should not change the matching.

binarix:
I have to admit that I have not actually considered the scenario of allowing someone to overwrite the api version. - I would probably argue that the api version is something that is tightly coupled to the implementation of the functionality supported and should therefore not be allowed to be overwritten. For example, setting api/v5 when it only supports api/v4 under the hood probably will not work out too well.

The robustness I was suggesting was more along the lines of https://mycompany.com/gitlab/ as an example if someone had that rewrite rule setup. - This is YAGNI territory and I would be ok with not implementing it this way if you disagree with my argument above. Just thought it might be a nice way of being able to support additional use case while reinforcing the strictness of the api version to the supported implementation.

There were also some minor refactoring such as having the apiVersionPath defined as a constant so that there just a single place to update that value.

svanharmelen:
Agreed, so LGTM! Thanks @binarix!

Issue #411, opened 2018-05-31
ListProjectMemberOptions and ListGroupMemberOptions should have the same type


bzon:
Problem:

* [__ListProjectMembersOptions__](https://godoc.org/github.com/xanzy/go-gitlab#ListGroupMembersOptions) is its own struct type and [__ListGroupMemberOption__](https://godoc.org/github.com/xanzy/go-gitlab#ListProjectMembersOptions) is a [__ListOption__](https://godoc.org/github.com/xanzy/go-gitlab#ListOptions) type. 
* ListGroupMemberOption is also missing a Query field.

Proposed Solution:

For package consistency, my proposal is to create a __ListMembersOptions__ type where both __ListGroupMemberOption__ and __ListProjectMembersOptions__ can be created. My reference is the [Group and project members API](https://docs.gitlab.com/ce/api/members.html#list-all-members-of-a-group-or-project).

svanharmelen:
If `ListGroupMemberOptions` needs more fields, then I suggest to just create a dedicated struct for it. So something like:

```go
type ListGroupMemberOptions struct {
  ListOptions
  OtherFields   string
}
```

Creating a 3rd struct that then can be used by two others doesn't feel right to me. Even if the structs are actually the same, I prefer to just have 2 separate ones in that case.

bzon:
@svanharmelen sounds fair to me.

PR #412, opened 2018-06-01
Change ListGroupMembersOptions to be its own struct

bzon:
This fixes https://github.com/xanzy/go-gitlab/issues/411


svanharmelen:
Thanks @bzon!

Issue #414, opened 2018-06-04
Groups.TransferGroup 'project' parameter should be an 'interface{}' not 'int'


bzon:
Current code: https://godoc.org/github.com/xanzy/go-gitlab#GroupsService.TransferGroup

```golang
func (s *GroupsService) TransferGroup(gid interface{}, project int, options ...OptionFunc) (*Group, *Response, error)
```

API doc: https://docs.gitlab.com/ce/api/groups.html#transfer-project-to-group

svanharmelen:
Fixed by PR #415

PR #415, opened 2018-06-04
Fix issue 414 for TransferGroup. Add tests for Groups

bzon:
__Work in Progress__ - Add test for all groups.go api.

Fixes issue https://github.com/xanzy/go-gitlab/issues/414.

### DIFF
--- a/groups.go
+++ b/groups.go
@@ -142,12 +142,19 @@ func (s *GroupsService) CreateGroup(opt *CreateGroupOptions, options ...OptionFu
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/groups.html#transfer-project-to-group
-func (s *GroupsService) TransferGroup(gid interface{}, project int, options ...OptionFunc) (*Group, *Response, error) {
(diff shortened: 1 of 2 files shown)


groups.go
```diff
@@ -142,7 +142,7 @@ func (s *GroupsService) CreateGroup(opt *CreateGroupOptions, options ...OptionFu
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/groups.html#transfer-project-to-group
-func (s *GroupsService) TransferGroup(gid interface{}, project int, options ...OptionFunc) (*Group, *Response, error) {
+func (s *GroupsService) TransferGroup(gid interface{}, project interface{}, options ...OptionFunc) (*Group, *Response, error) {
```
svanharmelen: Nice! You should also add something like `project, err := parseID(pid)` to make sure the ID is parsed correctly 👍

svanharmelen: And then maybe rename the argument from `project` to `pid` to be inline with other ode in this package. Thanks!

bzon:
@svanharmelen This is ready for review. I also applied your suggested change. 👍

svanharmelen:
Looks good @bzon! Your on 🔥

Thanks!

PR #416, opened 2018-06-05
Add ResetPassword  field for CreateUserOptions

bzon:
The latest documentation for user creation has the `reset_password` field. 

Reference: https://docs.gitlab.com/ce/api/users.html#user-creation


svanharmelen:
Nice! Thanks @bzon!

Issue #433, opened 2018-07-18
pagination


ecornelisse:
Most endpoints of gitlab use pagination so your `ListProjects` call returns only 20 results. How should i handle pagination with this package?

Lorac:
You can read `TotalPages` from the `Response struct` [here](https://github.com/xanzy/go-gitlab/blob/fb3ece09ef3e80cfa5c1e52dc75fcfb56c12e026/gitlab.go#L558)

From there you can loop for with the `ListProjectsOptions` by passing the `Page` number to the `ListProjects` call.

There is the `PerPage` attribute that you can modify, but I think it's cap to 20 results max.

Lorac:
@svanharmelen Might be a good idea to add an example of pagination in the directory examples. What do you think?

svanharmelen:
Yeah, that would be nice/helpful @Lorac :+1:

ecornelisse:
should it be something like this?:
```golang
        currentPage := 0
	for {
		currentPage++
		projects, response, err := git.Projects.ListProjects(options, func(request *http.Request) error {
			request.URL.Query().Add("page", string(currentPage))
			request.URL.Query().Add("per_page", string(10))
			return nil
		})
            }
```

Lorac:
@ecornelisse I added a pull request, that is my understanding of how pagination is working. I might be wrong tho.

PR #434, opened 2018-07-19
Add pagination example

Lorac:
This is my comprehension of how pagination is working.

Is it the intended way @svanharmelen ?

Fixes: #433

### DIFF
--- a/examples/pagination.go
+++ b/examples/pagination.go
@@ -1,0 +1,28 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/xanzy/go-gitlab"
+)
+
+func pagination() {
+	git := gitlab.NewClient(nil, "yourtokengoeshere")
+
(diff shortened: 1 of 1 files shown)


examples/pagination.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"fmt"
+
+	"github.com/xanzy/go-gitlab"
+)
+
+func pagination() {
+	git := gitlab.NewClient(nil, "yourtokengoeshere")
+	git.SetBaseURL("https://gitlab.com/api/v4")
```
svanharmelen: This is only needed if you want to use a custom URL. So I would not put it here using the default URL 👍

Lorac: Fixed in the new changes.

svanharmelen:
Thanks for this @Lorac!

PR #435, opened 2018-07-24
GetCommitRefs method added

killmeplz:
Implementation of https://docs.gitlab.com/ce/api/commits.html#get-references-a-commit-is-pushed-to

### DIFF
--- a/commits.go
+++ b/commits.go
@@ -50,6 +50,15 @@ type Commit struct {
 	Status         *BuildStateValue `json:"status"`
 }
 
+
+// CommitRefs represents the reference of branches/tags in a commit.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/commits.html#get-references-a-commit-is-pushed-to
+type CommitRef struct {
+	Type	string	`json:"type"`
+	Name	string	`json:"name"`
+}
+
 // CommitStats represents the number of added and deleted files in a commit.
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/commits.html
@@ -73,6 +82,7 @@ type ListCommitsOptions struct {
 	Until   *time.Time `url:"until,omitempty" json:"until,omitempty"`
 	Path    *string    `url:"path,omitempty" json:"path,omitempty"`
 	All     *bool      `url:"all,omitempty" json:"all,omitempty"`
+	WithStats	*bool		 `url:"with_stats,omitempty" json:"with_stats,omitempty"`
 }
 
 // ListCommits gets a list of repository commits in a project.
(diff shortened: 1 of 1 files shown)


commits.go
```diff
@@ -73,6 +82,7 @@ type ListCommitsOptions struct {
 	Until   *time.Time `url:"until,omitempty" json:"until,omitempty"`
 	Path    *string    `url:"path,omitempty" json:"path,omitempty"`
 	All     *bool      `url:"all,omitempty" json:"all,omitempty"`
+	WithStats	*bool		 `url:"with_stats,omitempty" json:"with_stats,omitempty"`
```
svanharmelen: It looks like you didn't run `gofmt` or `goimports` to format your changes...

commits.go
```diff
@@ -121,6 +131,31 @@ type CommitAction struct {
 	Encoding     string     `url:"encoding,omitempty" json:"encoding,omitempty"`
 }
 
+// GetCommitRefs Get all references (from branches or tags) a commit is pushed to
+// _type can be one of "all, branch, tag"
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/commits.html#get-references-a-commit-is-pushed-to
```
svanharmelen: If the URL is too long, then please place it on a second line:

```go
// GitLab API docs:
// https://docs.gitlab.com/ce/api/commits.html#get-references-a-commit-is-pushed-to
```

commits.go
```diff
@@ -121,6 +131,31 @@ type CommitAction struct {
 	Encoding     string     `url:"encoding,omitempty" json:"encoding,omitempty"`
 }
 
+// GetCommitRefs Get all references (from branches or tags) a commit is pushed to
+// _type can be one of "all, branch, tag"
```
svanharmelen: This comment looks a bit weird... Maybe just use the first sentence?

```go
// GetCommitRefs gets all references (from branches or tags) a commit is pushed to.
```

commits.go
```diff
@@ -50,6 +50,15 @@ type Commit struct {
 	Status         *BuildStateValue `json:"status"`
 }
 
+
+// CommitRef represents the reference of branches/tags in a commit.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/commits.html#get-references-a-commit-is-pushed-to
```
svanharmelen: If the URL is too long, then please place it on a second line:

```go
// GitLab API docs
// https://docs.gitlab.com/ce/api/commits.html#get-references-a-commit-is-pushed-to
```

commits.go
```diff
@@ -50,6 +50,15 @@ type Commit struct {
...
+type CommitRef struct {
+	Type	string	`json:"type"`
+	Name	string	`json:"name"`
+}
```
svanharmelen: Maybe declare this type above the method where it's used?

commits.go
```diff
@@ -121,6 +131,31 @@ type CommitAction struct {
 	Encoding     string     `url:"encoding,omitempty" json:"encoding,omitempty"`
 }
 
+// GetCommitRefs Get all references (from branches or tags) a commit is pushed to
+// _type can be one of "all, branch, tag"
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/commits.html#get-references-a-commit-is-pushed-to
+func (s *CommitsService) GetCommitRefs(pid interface{}, sha string, _type string, options ...OptionFunc) ([]CommitRef, *Response, error) {
```
svanharmelen: The approach in this package is to use method parameters for all variables that go into the path (URL) and option struct containing all the parameters that go into body. So in this case you need to make options struct for the `_type` variable.

And since this endpoint also supports pagination, you should then also embed the `ListOptions` struct.

Last but not least, please don't use `_type` as a variable name :wink: And since only a few specific values are allowed, we prefer to make a specific type for that so people cannot make a mistake by supplying a non-valid value.

commits.go
```diff
@@ -121,6 +131,31 @@ type CommitAction struct {
...
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/repository/commits/%s/refs?type=%s", url.QueryEscape(project), sha, _type)
+
+	req, err := s.client.NewRequest("GET", u, nil, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var c []CommitRef
```
svanharmelen: _nitting_ to indicate that its a list, I prefer to use `cs` instead of `c`.

killmeplz:
@svanharmelen got it in b9650b8

svanharmelen:
LGTM! Thanks @killmeplz!

Issue #428, opened 2018-06-29
ListTagsOptions should has Orderby and Sort


ball6847:
I want to ListTags and order them by name but can not find any way to make that happen

From gitlab api referene, list tags also accept `order_by` and `sort` parameter, so I think `ListTagsOptions` should also accept `OrderBy` and `Sort`

```go
type ListTagsOptions struct {
	ListOptions
	OrderBy *string `url:"order_by,omitempty" json:"order_by,omitempty"`
	Sort    *string `url:"sort,omitempty" json:"sort,omitempty"`
}
```

reference: https://docs.gitlab.com/ce/api/tags.html#list-project-repository-tags

svanharmelen:
Seem like the right solution @ball6847! Are you up for making a PR for this?

ball6847:
My only concern is breaking changes, since this can break someone code.

But if you are fine with this, I will make a PR.

svanharmelen:
As far as I can see this is not a breaking change. People that currently use the `ListTagsOptions` struct can keep doing that without any changes when you change it from:

```go
type ListTagsOptions ListOptions
```

to:

```go
type ListTagsOptions struct {
	ListOptions
	OrderBy *string `url:"order_by,omitempty" json:"order_by,omitempty"`
	Sort    *string `url:"sort,omitempty" json:"sort,omitempty"`
}
```

As the `ListOptions` struct is embedded, it shouldn't change the usage.

So by all means, go right ahead!

ball6847:
@svanharmelen I'm quite new to Go lang and never know about embedded struct before, so thank you about it.

I tried what you said. but  cannot make the test pass.

```sh
$ go test
# github.com/ball6847/go-gitlab
./tags_test.go:19:30: unknown field 'Page' in struct literal of type ListTagsOptions
./tags_test.go:19:42: unknown field 'PerPage' in struct literal of type ListTagsOptions
FAIL    github.com/ball6847/go-gitlab [build failed]
```

```sh
$ go version                                            
go version go1.10.2 linux/amd64
```

any suggestion ?

svanharmelen:
Maybe you can open a PR? Then I can see your code and test myself as well...

svanharmelen:
Fixed by PR #436

PR #436, opened 2018-07-25
listtagopts added params

killmeplz:
According to issue #428

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 # Folders
 _obj
 _test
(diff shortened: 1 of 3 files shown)


svanharmelen:
LGTM! Thanks!

killmeplz:
yep, and it seems to be not only tests, there's no backward compability

svanharmelen:
Yeah, I'll merge this one and then tag it with a new version... Thx!

Issue #142, opened 2017-03-21
Support all Services


johanbrandhorst:
Currently only a couple of the following services are supported:

Drone CI
Hipchat
Slack Notifications

Furthermore, only Drone CI has the getter defined, and there's also definitions for a Gitlab CI service which does not seem to exist anymore (using https://docs.gitlab.com/ce/api/services.html as a reference).

svanharmelen:
Yeah, the Gitlab API is being extended and improved continuously so I guess there are a few new services that are not supported in here yet. I a bit constrained on time, but will try to allocate some time in the near future.

Meanwhile of course any PR's adding what you need are much appreciated!

svanharmelen:
I'm going to close this issue as this is a never ending ongoing effort to keep this package in sync with the GitLab API.

PR #438, opened 2018-07-30
Supplement Slack service

hypnoglow:
This PR adds missing fields to fully support the GitLab Slack integration.

*I'm still playing with it for various scenarios and checking if all fields are correct, so this is WIP for now. But I'm glad to get feedback on this.*

Refs:
- https://docs.gitlab.com/ce/api/services.html#slack-notifications
- #142


services.go
```diff
@@ -252,7 +255,79 @@ type SlackService struct {
...
+	TagPushChannel            *string `json:"tag_push_channel,omitempty"`
+	PipelineChannel           *string `json:"pipeline_channel,omitempty"`
+	WikiPageChannel           *string `json:"wiki_page_channel,omitempty"`
+}
+
+// UnmarshalJSON implements json.Unmarshaler.
+func (p *SlackServiceProperties) UnmarshalJSON(b []byte) error {
+	// The problem here is that in some cases gitlab returns "notify_only_broken_pipelines": true,
+	// and in other cases "notify_only_broken_pipelines": "1". We need to handle this.
+	// The same with "notify_only_default_branch" field.
```
svanharmelen: Did you report this to Gitlab? As that clearly seems a bug (at the Gitlab docs clearly state it should be a `bool`).

services.go
```diff
@@ -252,7 +255,79 @@ type SlackService struct {
...
+	MergeRequestChannel       *string `json:"merge_request_channel,omitempty"`
+	NoteChannel               *string `json:"note_channel,omitempty"`
+	ConfidentialNoteChannel   *string `json:"confidential_note_channel,omitempty"`
+	TagPushChannel            *string `json:"tag_push_channel,omitempty"`
+	PipelineChannel           *string `json:"pipeline_channel,omitempty"`
+	WikiPageChannel           *string `json:"wiki_page_channel,omitempty"`
+}
+
+// UnmarshalJSON implements json.Unmarshaler.
+func (p *SlackServiceProperties) UnmarshalJSON(b []byte) error {
```
svanharmelen: I would prefer to make a custom `bool` type that has a custom `UnmarshalJSON` method. That way we can reuse the logic (if needed) elsewhere and we don't need this large tailored method in this file.

services.go
```diff
@@ -286,9 +361,32 @@ func (s *ServicesService) GetSlackService(pid interface{}, options ...OptionFunc
...
+	NoteChannel               *string `json:"note_channel,omitempty"`
+	ConfidentialNoteEvents    *bool   `json:"confidential_note_events,omitempty"`
+	// TODO: Currently, GitLab ignores this option (not implemented yet?), so
+	// there is no way to set it. Uncomment when this is fixed.
+	// See: https://gitlab.com/gitlab-org/gitlab-ce/issues/49730
+	//ConfidentialNoteChannel   *string `json:"confidential_note_channel,omitempty"`
+	PipelineEvents            *bool   `json:"pipeline_events,omitempty"`
+	PipelineChannel           *string `json:"pipeline_channel,omitempty"`
+	WikiPageEvents            *bool   `json:"wiki_page_events,omitempty"`
+	WikiPageChannel           *string `json:"wiki_page_channel,omitempty"`
```
svanharmelen: Please also add the `url` tags...

services.go
hypnoglow: I reported the issue: https://gitlab.com/gitlab-org/gitlab-ce/issues/50122
Also, I've put the link into the code near those problematic fields.

services.go
hypnoglow: Done. Added a new custom type `BoolValue`.

services.go
hypnoglow: I'm just curious why do we need this? As far as I understand, these fields are only encoded into JSON and send in a PUT request, but never in a query params.

svanharmelen: That is correct. So indeed for this specific options struct it would not be needed. But in order to keep things consistent and to prevent having to look at the type of HTTP call (e.g. `GET` or not) for every options struct, I prefer to just always add the `url` tags so however we use the struct, it will always work as expected.

types_test.go
```diff
@@ -0,0 +1,43 @@
...
+		},
+		"should unmarshal \"0\" as false": {
+			data: []byte(`"0"`),
+			expected: false,
+		},
+	}
+
+	for name, testCase := range testCases {
+		t.Run(name, func(t *testing.T) {
+		    var b BoolValue
```
svanharmelen: Please `go fmt` your code...

services.go
hypnoglow: Done

services.go
```diff
@@ -252,7 +254,27 @@ type SlackService struct {
...
+	Username                  *string   `json:"username,omitempty"`
+	PushChannel               *string   `json:"push_channel,omitempty"`
+	IssueChannel              *string   `json:"issue_channel,omitempty"`
+	ConfidentialIssueChannel  *string   `json:"confidential_issue_channel,omitempty"`
+	MergeRequestChannel       *string   `json:"merge_request_channel,omitempty"`
+	NoteChannel               *string   `json:"note_channel,omitempty"`
+	ConfidentialNoteChannel   *string   `json:"confidential_note_channel,omitempty"`
+	TagPushChannel            *string   `json:"tag_push_channel,omitempty"`
+	PipelineChannel           *string   `json:"pipeline_channel,omitempty"`
+	WikiPageChannel           *string   `json:"wiki_page_channel,omitempty"`
```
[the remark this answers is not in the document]
hypnoglow: Done

types_test.go
hypnoglow: I don't see what was wrong here. Anyway, I think it is formatted properly now.

svanharmelen:
Yeah I also noticed that... But it seems this is already used in a simular way by other services in this file. So thought to just keep the names consistent.

hypnoglow:
> Why renaming into WebHook? The API uses webhook so I guess with Go syntax it should become Webhook since it is a single word.

Yeah, I agree with @svanharmelen that there was already `WebHook` field so we have to keep the same name to support backward compatibility.

> Is this still WIP? Or do you think its ready to be merged?

I think it is ready, but I can take some time to double-check if everything works fine, if you wish.

svanharmelen:
Cool... Thanks for your efforts @hypnoglow!

hypnoglow:
I'm glad to help. Also thanks to @jorcau!

PR #440, opened 2018-08-02
labels are labels

doits:
There's already a helper for those Labels, so let's use it.


doits:
oh and I just see: This fixes a typo, too.

svanharmelen:
Thanks @doits! Missed that one 😏

PR #443, opened 2018-08-03
Adding Group Variables support

pmwebster:


group_variables.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	return vs, resp, err
+}
+
+// ShowVariableDetails gets a variable.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/group_level_variables.html#show-variable-details
+func (s *GroupVariablesService) ShowVariableDetails(gid interface{}, key string, options ...OptionFunc) (*GroupVariable, *Response, error) {
```
svanharmelen: I prefer this to be called just `GetVariable`. Did guess I missed that in the project variables PR. Would be good to update that one as well...

group_variables.go
```diff
@@ -0,0 +1,193 @@
...
+
+func (v GroupVariable) String() string {
+	return Stringify(v)
+}
+
+// ListGroupVariables gets a list of all variables in a group.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/group_level_variables.html#list-group-variables
+func (s *GroupVariablesService) ListGroupVariables(gid interface{}, options ...OptionFunc) ([]*GroupVariable, *Response, error) {
```
svanharmelen: I prefer this to be called just `ListVariables`. Did guess I missed that in the project variables PR. Would be good to update that one as well...

pmwebster:
Not sure why that is failing; If you look at the diff the duplicate structs are gone :/

svanharmelen:
LGTM! Thanks ⚡️

PR #451, opened 2018-08-20
added custom attributes

45cali:
I added custom attributes and provided tests as well. Custom attributes add a lot of flexibility for third party tools against the gitlab api. I'm hoping this can get merged quickly.

### DIFF
--- a/custom_attributes.go
+++ b/custom_attributes.go
@@ -22,44 +22,80 @@ type CustomAttribute struct {
 
 // ListCustomUserAttributes lists the custom attributes of the specified user.
 //
-// GitLab API docs:  https://docs.gitlab.com/ce/api/custom_attributes.html#list-custom-attributes
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/custom_attributes.html#list-custom-attributes
 func (s *CustomAttributesService) ListCustomUserAttributes(user int, options ...OptionFunc) ([]*CustomAttribute, *Response, error) {
 	return s.listCustomAttributes("users", user, options...)
 }
 
 // ListCustomGroupAttributes lists the custom attributes of the specified group.
 //
-// GitLab API docs:  https://docs.gitlab.com/ce/api/custom_attributes.html#list-custom-attributes
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/custom_attributes.html#list-custom-attributes
 func (s *CustomAttributesService) ListCustomGroupAttributes(group int, options ...OptionFunc) ([]*CustomAttribute, *Response, error) {
 	return s.listCustomAttributes("groups", group, options...)
 }
 
 // ListCustomProjectAttributes lists the custom attributes of the specified project.
 //
-// GitLab API docs:  https://docs.gitlab.com/ce/api/custom_attributes.html#list-custom-attributes
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/custom_attributes.html#list-custom-attributes
 func (s *CustomAttributesService) ListCustomProjectAttributes(project int, options ...OptionFunc) ([]*CustomAttribute, *Response, error) {
 	return s.listCustomAttributes("projects", project, options...)
 }
 
+func (s *CustomAttributesService) listCustomAttributes(resource string, id int, options ...OptionFunc) ([]*CustomAttribute, *Response, error) {
+	u := fmt.Sprintf("%s/%d/custom_attributes", resource, id)
+	req, err := s.client.NewRequest("GET", u, nil, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var cas []*CustomAttribute
+	resp, err := s.client.Do(req, &cas)
+	if err != nil {
+		return nil, resp, err
+	}
+	return cas, resp, err
+}
+
 // GetCustomUserAttribute returns the user attribute with a speciifc key.
 //
-// GitLab API docs:  https://docs.gitlab.com/ce/api/custom_attributes.html#single-custom-attribute
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/custom_attributes.html#single-custom-attribute
 func (s *CustomAttributesService) GetCustomUserAttribute(user int, key string, options ...OptionFunc) (*CustomAttribute, *Response, error) {
 	return s.getCustomAttribute("users", user, key, options...)
 }
 
 // GetCustomGroupAttribute returns the group attribute with a speciifc key.
 //
-// GitLab API docs:  https://docs.gitlab.com/ce/api/custom_attributes.html#single-custom-attribute
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/custom_attributes.html#single-custom-attribute
 func (s *CustomAttributesService) GetCustomGroupAttribute(group int, key string, options ...OptionFunc) (*CustomAttribute, *Response, error) {
 	return s.getCustomAttribute("groups", group, key, options...)
 }
 
 // GetCustomProjectAttribute returns the project attribute with a speciifc key.
 //
-// GitLab API docs:  https://docs.gitlab.com/ce/api/custom_attributes.html#single-custom-attribute
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/custom_attributes.html#single-custom-attribute
 func (s *CustomAttributesService) GetCustomProjectAttribute(project int, key string, options ...OptionFunc) (*CustomAttribute, *Response, error) {
 	return s.getCustomAttribute("projects", project, key, options...)
+}
+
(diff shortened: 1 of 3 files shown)


custom_attributes.go
```diff
@@ -0,0 +1,125 @@
...
+// ListCustomAttributes lists the custom attributes of the specified resource
+//
+// GitLab API docs:  https://docs.gitlab.com/ee/api/custom_attributes.html#list-custom-attributes
+func (s *CustomeAttributesService) ListCustomAttributes(opt CustomAttributeOptions, options ...OptionFunc) ([]*CustomAttributes, *Response, error) {
+	resourceID, err := parseID(opt.CustomeAttributeResourceID)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	u := fmt.Sprintf("%s/%s/custom_attributes", opt.CustomeAttributeResource, url.QueryEscape(resourceID))
```
svanharmelen: Please see some of the other services for comparison... The default approach is to have the parts that make up the URL, be function parameters. And all additional fields should be in the options structs.

So for this function that should be:

```
func (s *CustomAttributesService) ListCustomAttributes(resource int, options ...OptionFunc) ([]*CustomAttribute, *Response, error) {
```

Additionally you could have a look at the `search` service: https://github.com/xanzy/go-gitlab/blob/master/search.go As you can see it uses fully named functions that then call a private function to execute the actual logic. So in this service it could be something like this:

```
func (s *CustomAttributesService) ListCustomUserAttributes(user int, options ...OptionFunc) ([]*CustomAttribute, *Response, error) {
  return s.listCustomAttributes("users", user, options...)
}
```

And

```
func (s *CustomAttributesService) ListCustomGroupAttributes(group int, options ...OptionFunc) ([]*CustomAttribute, *Response, error) {
  return s.listCustomAttributes("groups", group, options...)
}
```

Which then call a private function like:

```
func (s *CustomAttributesService) listCustomAttributes(resource string, id int, options ...OptionFunc) ([]*CustomAttribute, *Response, error) {
  u := fmt.Sprintf("%s/%d/custom_attributes", resource, id)
  req, err := s.client.NewRequest("GET", u, nil, options)
  if err != nil {
    return nil, nil, err
  }
  
  var ca []*CustomAttribute
  resp, err := s.client.Do(req, &ca)
  if err != nil {
    return nil, resp, err
  }

  return ca, resp, err
}
```

This provides a much better UX and is more consistent with the other services in this package. Of course this comment goes for all the functions of this service...

custom_attributes.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	var ca []*CustomAttributes
+	resp, err := s.client.Do(req, &ca)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return ca, resp, err
+
```
svanharmelen: This blank line should be removed...

custom_attributes.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	var ca *CustomAttributes
+	resp, err := s.client.Do(req, &ca)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return ca, resp, err
+
```
svanharmelen: This blank line should be removed...

custom_attributes.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// GetCustomAttribute returns the attribute with a speciifc key
```
svanharmelen: Typo: `speciifc` -> `specific`

custom_attributes.go
```diff
@@ -0,0 +1,125 @@
...
+//
+// GitLab API docs:  https://docs.gitlab.com/ee/api/custom_attributes.html#single-custom-attribute
```
svanharmelen: Please put the URL on its own line.

custom_attributes.go
```diff
@@ -0,0 +1,125 @@
...
+// CustomeAttributesService handles communication with the
+// group, project and user custom attributes related methods
+// of the GitLab API.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/custom_attributes.html
+type CustomeAttributesService struct {
+	client *Client
+}
+
+// CustomAttributes struct is used to unmarshal response to api calls
```
svanharmelen: Please make sure all comments end with a dot, so it a full sentence.

custom_attributes.go
```diff
@@ -0,0 +1,125 @@
...
+		return nil, resp, err
+	}
+
+	return ca, resp, err
+
+}
+
+// SetCustomAttribute sets the custom attributes of the specified resource
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/custom_attributes.html#set-custom-attribute
```
svanharmelen: Please put the URL on its own line.

custom_attributes.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return ca, resp, err
+}
+
+// DeleteCustomAttribute removes the customattribute fromt the specified resource
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/custom_attributes.html#delete-custom-attribute
```
svanharmelen: Please put the URL on its own line.

custom_attributes.go
```diff
@@ -0,0 +1,125 @@
...
+
+// CustomAttributes struct is used to unmarshal response to api calls
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/custom_attributes.html
+type CustomAttributes struct {
+	Key   string `json:"key"`
+	Value string `json:"value"`
+}
+
+// CustomAttributeOptions struct contains all options required to call the methods below
```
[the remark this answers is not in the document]
svanharmelen: Please make sure all comments end with a dot, so it a full sentence.

custom_attributes.go
svanharmelen: Please make sure all comments end with a dot, so it a full sentence.

custom_attributes.go
```diff
@@ -0,0 +1,125 @@
...
+	resp, err := s.client.Do(req, &ca)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return ca, resp, err
+
+}
+
+// SetCustomAttribute sets the custom attributes of the specified resource
```
svanharmelen: Please make sure all comments end with a dot, so it a full sentence.

custom_attributes.go
```diff
@@ -0,0 +1,125 @@
...
+	ca := new(CustomAttributes)
+	resp, err := s.client.Do(req, ca)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return ca, resp, err
+}
+
+// DeleteCustomAttribute removes the customattribute fromt the specified resource
```
svanharmelen: Please make sure all comments end with a dot, so it a full sentence.

gitlab.go
```diff
@@ -460,7 +462,6 @@ func newClient(httpClient *http.Client) *Client {
 	c.Validate = &ValidateService{client: c}
 	c.Version = &VersionService{client: c}
 	c.Wikis = &WikisService{client: c}
-
```
svanharmelen: I actually prefer to have this blank line here 😏

custom_attributes.go
```diff
@@ -0,0 +1,169 @@
...
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/custom_attributes.html
+type CustomAttribute struct {
+	Key   string `json:"key"`
+	Value string `json:"value"`
+}
+
+// ListCustomUserAttributes lists the custom attributes of the specified user.
+//
+// GitLab API docs:  https://docs.gitlab.com/ce/api/custom_attributes.html#list-custom-attributes
```
svanharmelen: I try to keep comments below 80 chars. When the URL is long and exceeds 80 chars, then just put it on a line of its own like this:

```
// GitLab API docs: 
// https://docs.gitlab.com/ce/api/custom_attributes.html#list-custom-attributes
```

For code I try to keep a limit of 100 chars, except for function signatures which I don't limit at all.

custom_attributes.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func (s *CustomAttributesService) listCustomAttributes(resource string, id int, options ...OptionFunc) ([]*CustomAttribute, *Response, error) {
+	u := fmt.Sprintf("%s/%d/custom_attributes", resource, id)
+	req, err := s.client.NewRequest("GET", u, nil, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var ca []*CustomAttribute
```
svanharmelen: This should be `cas` to indicate it's not a single item.

45cali:
@svanharmelen is their anything else that needs to be done for this PR?

svanharmelen:
Thanks for this one @45cali! Looks really good now 👍

PR #452, opened 2018-08-22
feat: add microsoft teams integration

sascha-andres:
Add ability to configure Microsoft Teams integration

### DIFF
--- a/services.go
+++ b/services.go
@@ -513,3 +513,91 @@ func (s *ServicesService) DeleteJenkinsCIService(pid interface{}, options ...Opt
 
 	return s.client.Do(req, nil)
 }
+
+// MicrosoftTeamsService represents Microsoft Teams service settings.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/services.html#microsoft-teams
+type MicrosoftTeamsService struct {
+	Service
+	Properties *MicrosoftTeamsServiceProperties `json:"properties"`
+}
+
+// MicrosoftTeamsServiceProperties represents Microsoft Teams specific properties.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/services.html#microsoft-teams
+type MicrosoftTeamsServiceProperties struct {
+	WebHook string `json:"webhook"`
+}
+
+// GetMicrosoftTeamsService gets MicrosoftTeams service settings for a project.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/services.html#get-microsoft-teams-service-settings
+func (s *ServicesService) GetMicrosoftTeamsService(pid interface{}, options ...OptionFunc) (*MicrosoftTeamsService, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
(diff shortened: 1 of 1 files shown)


services.go
```diff
@@ -513,3 +513,90 @@ func (s *ServicesService) DeleteJenkinsCIService(pid interface{}, options ...Opt
...
+	Service
+	Properties *MicrosoftTeamsServiceProperties `json:"properties"`
+}
+
+// MicrosoftTeamsServiceProperties represents Microsoft Teams specific properties.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/services.html#microsoft-teams
+type MicrosoftTeamsServiceProperties struct {
+	WebHook                   *string `webhook:"webhook,omitempty" json:"webhook,omitempty"`
```
svanharmelen: It looks like you didn’t `gofmt` your code. Could you please do that to make sure its formatted correctly?

services.go
```diff
@@ -513,3 +513,90 @@ func (s *ServicesService) DeleteJenkinsCIService(pid interface{}, options ...Opt
...
+	}
+	u := fmt.Sprintf("projects/%s/services/microsoft-teams", url.QueryEscape(project))
+
+	req, err := s.client.NewRequest("DELETE", u, nil, options)
+	if err != nil {
+		return nil, err
+	}
+
+	return s.client.Do(req, nil)
+}
```
svanharmelen: Please end the file with a newline... Maybe `gofmt` does this as well, but not sure about that.

services.go
svanharmelen: You should not use a pointer here and the tag is not correct either... Please make sure they are inline with the other examples in this file.

services.go
```diff
@@ -513,3 +513,90 @@ func (s *ServicesService) DeleteJenkinsCIService(pid interface{}, options ...Opt
...
+
+	return svc, resp, err
+}
+
+// SetMicrosoftTeamsServiceOptions represents the available SetMicrosoftTeamsService()
+// options.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/services.html#create-edit-microsoft-teams-service
+type SetMicrosoftTeamsServiceOptions MicrosoftTeamsServiceProperties
```
svanharmelen: This should be a different type with a pointer to the `webhook` value and different tags.

sascha-andres:
@svanharmelen changes done

services.go
```diff
@@ -513,3 +513,91 @@ func (s *ServicesService) DeleteJenkinsCIService(pid interface{}, options ...Opt
...
+	Service
+	Properties *MicrosoftTeamsServiceProperties `json:"properties"`
+}
+
+// MicrosoftTeamsServiceProperties represents Microsoft Teams specific properties.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/services.html#microsoft-teams
+type MicrosoftTeamsServiceProperties struct {
+	WebHook string `url:"webhook,omitempty" json:"webhook,omitempty"`
```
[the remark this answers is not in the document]
sascha-andres: changed

PR #454, opened 2018-08-23
Update runners

jorcau:
**Description**

This PR aims to update runners:
* Add missing methods: register new runner, delete registered runner, verify registered runner
* Fix typo: `Token` -> `token` in json field
* Add missing parameter: `maximum_timeout`

Refs: https://docs.gitlab.com/ce/api/runners.html

### DIFF
--- a/runners.go
+++ b/runners.go
@@ -64,11 +64,12 @@ type RunnerDetails struct {
 		Path              string `json:"path"`
 		PathWithNamespace string `json:"path_with_namespace"`
 	} `json:"projects"`
-	Token       string   `json:"Token"`
-	Revision    string   `json:"revision,omitempty"`
-	TagList     []string `json:"tag_list"`
-	Version     string   `json:"version,omitempty"`
-	AccessLevel string   `json:"access_level"`
+	Token          string   `json:"Token"`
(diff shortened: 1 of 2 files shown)


runners.go
```diff
@@ -324,3 +326,90 @@ func (s *RunnersService) DisableProjectRunner(pid interface{}, rid interface{},
...
 	return s.client.Do(req, nil)
 }
+
+// NewRunner represents the GitLab CI runner registration response.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/runners.html#register-a-new-runner
+type NewRunner struct {
+	ID    string `url:"id" json:"id"`
+	Token string `url:"token" json:"token"`
+}
```
svanharmelen: I think I prefer to reuse the `Runner` struct for this one. We can do that by just adding the `token` field to that struct. Does that make sense?

jorcau: I totally agree. I just got disturbed by the doc which is misleading: it implies the `id` is a string which actually isn't, it's an int. So the Runner struct is appropriate. Thanks for the review!

svanharmelen:
Nice! Thanks @jorcau!

PR #455, opened 2018-08-23
Change from using url.Opaque to url.Path and url.RawPath

### LINKED ISSUE #453: Requests can not go through forward proxy
I am trying to proxy go-gitlab's requests through a forward proxy in a corporate environment. It seems that this is not possible with the current implementation. The problem is that forward proxies expect absolute URLs to be sent in the request, but because go-gitlab sets an Opaque path with no host part, a relative path is always used. If the `Opaque` property is set on the URL, and a proxy is being used, Go's HTTP library expects that the Opaque property will include an absolute URL. The solution here is to change [this line](https://github.com/xanzy/go-gitlab/blob/master/gitlab.go#L504) from `u.Opaque` to `u.Path`. I have no idea what the other implications of this change would be, as I know this field has been the subject of previous issues like #10.

Here's a small example program that shows the issue:

```
package main

import (
	"fmt"
	"net/http"
	"net/url"
)

type stringWriter struct {
	data string
}

func (s *stringWriter) Write(data []byte) (int, error) {
	s.data += string(data)
	return len(data), nil
}

func (s *stringWriter) String() string {
	return s.data
}

func main() {
	u, _ := url.Parse("http://google.com/")
	u.Opaque = "/"

	req := &http.Request{
		Method:     "GET",
		URL:        u,
		Proto:      "HTTP/1.1",
		ProtoMajor: 1,
		ProtoMinor: 1,
		Header:     make(http.Header),
		Host:       u.Host,
	}

	s := &stringWriter{}
	req.WriteProxy(s)
	fmt.Print(s)
	fmt.Print("\n")

	u, _ = url.Parse("http://google.com/")
	u.Path = "/"
	req.URL = u

	s = &stringWriter{}
	req.WriteProxy(s)
	fmt.Print(s)
}
```

This program first prints out the request after setting the `Opaque` property of the url, then the same request with the `Path` set, instead of `Opaque`. This is the output:

```
GET / HTTP/1.1
Host: google.com
User-Agent: Go-http-client/1.1


GET http://google.com/ HTTP/1.1
Host: google.com
User-Agent: Go-http-client/1.1
```

The first request ommits the host and scheme, and is therefore invalid. Per RFC 2068 (definition of HTTP/1.1), section 5.1.2

>    The absoluteURI form is required when the request is being made to a
>   proxy. The proxy is requested to forward the request or service it
>   from a valid cache, and return the response. Note that the proxy MAY
>   forward the request on to another proxy or directly to the server
>   specified by the absoluteURI. In order to avoid request loops, a
>   proxy MUST be able to recognize all of its server names, including
>   any aliases, local variations, and the numeric IP address. An example
>   Request-Line would be:
>
>          GET http://www.w3.org/pub/WWW/TheProject.html HTTP/1.1
>
>   To allow for transition to absoluteURIs in all requests in future
>   versions of HTTP, all HTTP/1.1 servers MUST accept the absoluteURI
>   form in requests, even though HTTP/1.1 clients will only generate
>   them in requests to proxies.

natemara:
This fixes #453. I'm not sure if it will break #10.

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -500,8 +500,14 @@ func (c *Client) SetBaseURL(urlStr string) error {
 // request body.
 func (c *Client) NewRequest(method, path string, opt interface{}, options []OptionFunc) (*http.Request, error) {
 	u := *c.baseURL
-	// Set the encoded opaque data
-	u.Opaque = c.baseURL.Path + path
+	unescaped, err := url.PathUnescape(path)
+	if err != nil {
+		return nil, err
+	}
+
+	// Set the encoded path data
+	u.RawPath = c.baseURL.Path + path
+	u.Path = c.baseURL.Path + unescaped
 
 	if opt != nil {
 		q, err := query.Values(opt)
@@ -681,7 +687,7 @@ type ErrorResponse struct {
 }
 
 func (e *ErrorResponse) Error() string {
-	path, _ := url.QueryUnescape(e.Response.Request.URL.Opaque)
+	path, _ := url.QueryUnescape(e.Response.Request.URL.Path)
 	u := fmt.Sprintf("%s://%s%s", e.Response.Request.URL.Scheme, e.Response.Request.URL.Host, path)
 	return fmt.Sprintf("%s %s: %d %s", e.Response.Request.Method, u, e.Response.StatusCode, e.Message)
 }


svanharmelen:
Hi @natemara!

Thanks for this PR! I needed to look into this for a bit to make sure I understood this correctly. But I see what you did here. Instead of reverting PR #10, you choose to use the "new" `u.RawPath` field (which essentially has the same effect).

I say "new" here, as that field was not available when this package was first created. So back then using the `u.Opaque` field was the only option to support using URIs with an encoded `namespace/project` part in it. But I guess since `u.RawPath` exists, using that is indeed a better approach.

So that part is all good, but the question remains if this breaks the use case discussed in PR #10. Reading RFC 2068 it seems the problem would be related to the Apache and/or the used Apache config instead of the fact that this package was sending absolute URIs.

My only problem is that I don't have a setup to test or verify this... So that makes it a bit difficult to merge this PR with a certain amount of certainty that it will not break stuff for existing users. Any idea how to move forward with this? Are you in the position to test this?

natemara:
I was able to verify that if there is no forward proxy set up, go will not include the host in the GET request when using the `Path` and `RawPath` fields. Here's an example program that shows that:

```
package main

import (
	"fmt"
	"net/http"
	"net/url"
)

type stringWriter struct {
	data string
}

func (s *stringWriter) Write(data []byte) (int, error) {
	s.data += string(data)
	return len(data), nil
}

func (s *stringWriter) String() string {
	return s.data
}

func main() {
	u, _ := url.Parse("http://google.com/")
	u.Path = "/"
	u.RawPath = "/"

	req := &http.Request{
		Method:     "GET",
		URL:        u,
		Proto:      "HTTP/1.1",
		ProtoMajor: 1,
		ProtoMinor: 1,
		Header:     make(http.Header),
		Host:       u.Host,
	}

	s := &stringWriter{}
	req.WriteProxy(s)
	fmt.Print(s)
	fmt.Print("\n")

	s = &stringWriter{}
	req.Write(s)
	fmt.Print(s)
}
```

It outputs:

```
GET http://google.com/ HTTP/1.1
Host: google.com
User-Agent: Go-http-client/1.1


GET / HTTP/1.1
Host: google.com
User-Agent: Go-http-client/1.1
```

`request.Write` is used for writing request data to TCP streams in normal circumstances, and `request.WriteProxy` is used when writing to an HTTP proxy.

svanharmelen:
Oh, very nice @natemara! I didn’t even think about checking that... 😏

But that seems to be all we need to know to be able to safely merge this change.

Thanks again for this one! 👍🏻

natemara:
No problem, thanks for the merge!

MagicSong:
Does this pr will modify c.baseURL? I thing u is c.baseURL，so next i reuse this client，the path will be wrong like /usersusers

svanharmelen:
Hi @MagicSong! Did you actually test this? As I don’t see how `c.baseURL` is used differently then before. Thanks.

MagicSong:
@svanharmelen ,after more tests, I found  it is the bug in my program, not yours. So sorry

svanharmelen:
No problem!

PR #459, opened 2018-08-30
Implement DownloadSingleArtifactsFile

MarkusFreitag:
I've missed the possibility to download a single file from a jobs artifacts archive, as I need it in another tool.

P.S.: I've checked the `DownloadArtifactsFile` while implementing this and I think that all variables should be escaped before building the url. Also `project` is currently escaped using `url.QueryEscape`, but as it is part of the url path `url.PathEscape` should be used.
```
u := fmt.Sprintf(
  "projects/%s/jobs/artifacts/%s/download?job=%s",
  url.PathEscape(project),
  url.PathEscape(refName),
  url.QueryEscape(job),
)
```

### DIFF
--- a/jobs.go
+++ b/jobs.go
@@ -212,7 +212,7 @@ func (s *JobsService) DownloadSingleArtifactsFile(pid interface{}, jobID int, ar
 
 	u := fmt.Sprintf(
 		"projects/%s/jobs/%d/artifacts/%s",
-		url.PathEscape(project),
+		url.QueryEscape(project),
 		jobID
 		url.PathEscape(artifactPath),
 	)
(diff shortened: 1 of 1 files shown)


jobs.go
```diff
@@ -197,6 +197,44 @@ func (s *JobsService) DownloadArtifactsFile(pid interface{}, refName string, job
...
+	job, err := parseID(jid)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	u := fmt.Sprintf(
+		"projects/%s/jobs/%s/artifacts/%s",
+		url.PathEscape(project),
+		url.PathEscape(job),
+		url.PathEscape(artifactPath),
```
svanharmelen: According to this example (copied from the docs) this should not be escaped:

```
curl --header "PRIVATE-TOKEN: 9koXpg98eAheJpvBs5tK" "https://gitlab.example.com/api/v4/projects/1/jobs/5/artifacts/some/release/file.pdf"
```

MarkusFreitag:
Thank you for the quick review, I've done the required changes.

svanharmelen:
Thanks again @MarkusFreitag! Looks great now 👍

PR #461, opened 2018-08-30
projects.go: adding approvals_before_merge to CreateProjectOptions

pmwebster:
I know this isn't in the CE edition, however it is in the EE edition (ref: https://docs.gitlab.com/ee/api/projects.html#create-project)


svanharmelen:
If you update the doc link from `ce` to `ee` above this struct, we are good 👍 Thanks!

pmwebster:
Good point! (added)

svanharmelen:
LGTM! Thanks @pmwebster!

PR #468, opened 2018-09-11
 Add GetProjectLanguages to get languages from a project 

topochan:
Implements the languages project API https://docs.gitlab.com/ce/api/projects.html#languages

Small detail: 
* Defines ProjectLanguages as map of string interfaces because the response is not structured and will have a list of languages with percentage.

* Includes an example

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -282,6 +282,35 @@ func (s *ProjectsService) ListProjectsUsers(pid interface{}, opt *ListProjectUse
 	return p, resp, err
 }
 
+// ProjectLanguages is a map of strings because the response is arbitrary
+// Gitlab API docs with example: https://docs.gitlab.com/ce/api/projects.html#languages
+type ProjectLanguages map[string]interface{}
+
+// GetProjectLanguages gets a list of languages of the project ID
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/projects.html#languages
+func (s *ProjectsService) GetProjectLanguages(pid interface{}, options ...OptionFunc) (*ProjectLanguages, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/languages", url.QueryEscape(project))
+
+	req, err := s.client.NewRequest("GET", u, nil, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	p := new(ProjectLanguages)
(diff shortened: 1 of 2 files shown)


projects.go
```diff
@@ -282,6 +282,35 @@ func (s *ProjectsService) ListProjectsUsers(pid interface{}, opt *ListProjectUse
 	return p, resp, err
 }
 
+// ProjectLanguages is a map of strings because the response is arbitrary
+// Gitlab API docs with example: https://docs.gitlab.com/ce/api/projects.html#languages
```
svanharmelen: Please put a blank comment line in between these lines and call it just `Gitlab API docs:`:

```go
// ProjectLanguages is a map of strings because the response is arbitrary
//
// Gitlab API docs: https://docs.gitlab.com/ce/api/projects.html#languages
```

projects.go
```diff
@@ -282,6 +282,35 @@ func (s *ProjectsService) ListProjectsUsers(pid interface{}, opt *ListProjectUse
...
+type ProjectLanguages map[string]interface{}
```
svanharmelen: Reading the docs it seems to me this should be a `map[string]float`

projects.go
```diff
@@ -282,6 +282,35 @@ func (s *ProjectsService) ListProjectsUsers(pid interface{}, opt *ListProjectUse
...
+
+// GetProjectLanguages gets a list of languages of the project ID
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/projects.html#languages
```
svanharmelen: This comment is shot enough (<80 chars) to put on one line.

projects.go
```diff
@@ -282,6 +282,35 @@ func (s *ProjectsService) ListProjectsUsers(pid interface{}, opt *ListProjectUse
 	return p, resp, err
 }
 
+// ProjectLanguages is a map of strings because the response is arbitrary
+// Gitlab API docs with example: https://docs.gitlab.com/ce/api/projects.html#languages
+type ProjectLanguages map[string]interface{}
+
+// GetProjectLanguages gets a list of languages of the project ID
```
svanharmelen: Maybe describe this as: `// GetProjectLanguages gets a list of languages used by the project`

projects.go
topochan: You are right! thanks!

topochan:
I made the changes that you suggested, thanks for the input, is much better now

svanharmelen:
LGTM! Thanks @topochan!

PR #469, opened 2018-09-11
Adds OrderBy/Sort to ListIssuesNotesOptions

mpearson-mpt:
The OrderBy and Sort API optional params were missing for the ListIssuesNotesOptions struct.


svanharmelen:
LGTM! Thanks @mpearson-mpt!

Issue #477, opened 2018-09-27
Miss the Changes field of MergeEvent struct in event_type.go


SeananXu:
Cloud you please add this field in the MergeEvent struct in event_type.go? I can use this field to know the gitlab changes in mr event.
![image](https://user-images.githubusercontent.com/26948962/46129980-bd1e4680-c26a-11e8-9ac3-ca125c8084be.png)

#478

svanharmelen:
Fixed by PR #479

PR #479, opened 2018-09-27
Add (us)subscribe to/from issues

### LINKED ISSUE #476: Missing "Unsubscribe" from issues
Hello,

I can't find the method to unsubscribe from issues: 

https://docs.gitlab.com/ce/api/issues.html#unsubscribe-from-an-issue

("Subscribe" isn't available as well)

svanharmelen:
And extend the MergeEvent struct.

Fixes #476 and #477

### DIFF
--- a/event_types.go
+++ b/event_types.go
@@ -490,6 +490,24 @@ type MergeEvent struct {
 		Username  string `json:"username"`
 		AvatarURL string `json:"avatar_url"`
 	} `json:"assignee"`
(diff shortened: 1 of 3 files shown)


svanharmelen:
Whoops, added the tag (through it would work just as well without it)! Thanks!

And just for your info, you can also comment on specific lines of the commit right here on Github. That way you don't have to copy/paste images, but can just make inline remarks.

PR #482, opened 2018-10-03
Query encoding should use url.PathEscape

Wolfchin:
In repository_files.go, url encoding is done with url.QueryEscape, but this makes it impossible to get files with a space in their name. Using url.PathEncodes changes the space encoding from "+" to "%20" and fixes the issue.

### DIFF
--- a/repository_files.go
+++ b/repository_files.go
@@ -66,7 +66,7 @@ func (s *RepositoryFilesService) GetFile(pid interface{}, fileName string, opt *
 	if err != nil {
 		return nil, nil, err
 	}
-	u := fmt.Sprintf("projects/%s/repository/files/%s", url.QueryEscape(project), url.QueryEscape(fileName))
+	u := fmt.Sprintf("projects/%s/repository/files/%s", url.PathEscape(project), url.PathEscape(fileName))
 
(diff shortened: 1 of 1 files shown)


svanharmelen:
Even through this is a very correct solution, I would like to limit the changes from `QueryEscape` to `PathEscape`. At least until we decide to update it everywhere and to keep things consistent.

So this fix the actual issue, could you update the PR to only change it when escaping the filenames?

repository_files.go
```diff
@@ -99,7 +99,7 @@ func (s *RepositoryFilesService) GetRawFile(pid interface{}, fileName string, op
 	if err != nil {
 		return nil, nil, err
 	}
-	u := fmt.Sprintf("projects/%s/repository/files/%s/raw", url.QueryEscape(project), url.QueryEscape(fileName))
+	u := fmt.Sprintf("projects/%s/repository/files/%s/raw", url.QuesryEscape(project), url.PathEscape(fileName))
```
svanharmelen: There a typo in here: `QuesryEscape` (twice actually)

Wolfchin: Oh sorry, i've just fixed that.

svanharmelen:
LGTM! Thanks @Wolfchin!

PR #489, opened 2018-10-12
Add support of filename, path and commit id to GetFileOptions

groovili:
This change will give the ability to get repository files by name and commit sha.
Detailed info about this options available at [GitLab API](https://docs.gitlab.com/ee/api/repository_files.html#get-file-from-repository).

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -72,8 +72,8 @@ const (
 	OwnerPermissions      AccessLevelValue = 50
 
 	// These are deprecated and should be removed in a future version
-	MasterPermissions     AccessLevelValue = 40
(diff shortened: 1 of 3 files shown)


svanharmelen:
Thanks for your PR @groovili! Looking at the docs I cannot see that these are valid options for getting as repo file. I can only find/see these:

image

So where did you read/see that the other fields are valid options as well?

svanharmelen:
I was already looking at that page, but it only shows those fields in the response. Not in the requests...

groovili:
Sorry, my bad. Missed that :)

svanharmelen:
No problem 👍🏻

Issue #487, opened 2018-10-11
The Label model in gitlab api is different with the label model in gitlab event


SeananXu:
Label model in gitlab api:
https://docs.gitlab.com/ee/api/labels.html
![image](https://user-images.githubusercontent.com/26948962/46796148-81ad6d00-cd7e-11e8-85e0-20cdee7af115.png)
Label model in gitlab event:
https://docs.gitlab.com/ee/user/project/integrations/webhooks.html#merge-request-events
So the same Label model should not be used in these two places.
![image](https://user-images.githubusercontent.com/26948962/46796434-4a8b8b80-cd7f-11e8-9d4d-83060f78894d.png)
![image](https://user-images.githubusercontent.com/26948962/46796537-87f01900-cd7f-11e8-9e52-9897b2074ce9.png)

svanharmelen:
Guess we either need to add a `Title` field to the `Label` struct or create a specific struct for this.

SeananXu:
@svanharmelen  lgtm, thank you very much

PR #490, opened 2018-10-15
Support the Gitlab inconsistency around label names

svanharmelen:
Adding the custom unmarshal method enables this struct to be used in
both the label endpoints as well as the event types. It seems the Gitlab
API itself is a bit inconsistent here as it mixes the `name` and `title`
fields in the respective types.

### DIFF
--- a/labels.go
+++ b/labels.go
@@ -17,6 +17,7 @@
 package gitlab
 
 import (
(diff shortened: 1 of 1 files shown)


svanharmelen:
Fixed #487

PR #491, opened 2018-10-17
Added WebUrl to Job struct that is provided by the Gitlab API json, b…

nikogura:
…ut not included in the client Job struct.


svanharmelen:
LGTM!

PR #494, opened 2018-10-19
Addling support for license api endpoint

pmwebster:


license.go
```diff
@@ -0,0 +1,97 @@
...
+
+	return l, resp, err
+}
+
+// AddLicense adds a new license.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/license.html#add-a-new-license
+func (s *LicenseService) AddLicense(license string) (*License, *Response, error) {
+	u := fmt.Sprintf("license?license=%s", url.QueryEscape(license))
```
svanharmelen: Since this is a `POST`, I prefer to use an option struct here that we pass to the `NewRequest` method. This way this methods remains consistent with the rest of this package.

license.go
```diff
@@ -0,0 +1,97 @@
...
+func (l License) String() string {
+	return Stringify(l)
+}
+
+// GetLicense retrieves information about the current license.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/license.html#retrieve-information-about-the-current-license
+func (s *LicenseService) GetLicense() (*License, *Response, error) {
+	u := "license"
```
svanharmelen: Since `"license"` is just a single string literal, I suggest we place that directly in the `NewRequest` call.

license.go
```diff
@@ -0,0 +1,97 @@
...
+	client *Client
+}
+
+// License represents a GitLab license.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/license.html
+type License struct {
+	StartsAt  *time.Time `json:"starts_at"`
+	ExpiresAt *time.Time `json:"expires_at"`
```
svanharmelen: This doesn't work with the timestamp formats that the API returns here. This should be `*ISOTime` to make it work properly I think.

pmwebster: fixed

license.go
pmwebster: makes perfect sense! fixed.

license.go
pmwebster: Good call. This has also been fixed!

svanharmelen:
Looks good @pmwebster! Thanks!

PR #497, opened 2018-10-19
runners.go: Adding IPAddress field, updated docs url

pmwebster:


pmwebster:
Sorry, I forgot to rebase :/ Fortunately those are already merged in.

runners.go
```diff
@@ -32,16 +33,17 @@ type RunnersService struct {
 
 // Runner represents a GitLab CI Runner.
 //
-// GitLab API docs: https://docs.gitlab.com/ce/api/runners.html
+// GitLab API docs: https://docs.gitlab.com/ee/api/runners.html
```
svanharmelen: Why are you updating the URL? I prefer to reference the `ce` docs, unless there are fields or methods in there that are only supported by `ee`.

svanharmelen: Think I missed that in the other PR, can you refert that?

pmwebster: sure thing

pmwebster:
The users fields *are* different between CE and EE, given as much, leaving as in. Sorry for the confusion :)

pmwebster:
Would it be easier if I just rebased and make a new PR? I don't know why this is failing.

svanharmelen:
LGTM! Thanks!

PR #498, opened 2018-10-19
Adding test for keys endpoint

pmwebster:
Another simple one, only one method to test.

### DIFF
--- a/keys_test.go
+++ b/keys_test.go
@@ -1,0 +1,86 @@
+package gitlab
+
+import (
+	"fmt"
+	"net/http"
+	"reflect"
+	"testing"
+)
+
+func TestGetKeyWithUser(t *testing.T) {
+	mux, server, client := setup()
+	defer teardown(server)
+
+	mux.HandleFunc("/api/v4/keys/1",
+		func(w http.ResponseWriter, r *http.Request) {
+			testMethod(t, r, "GET")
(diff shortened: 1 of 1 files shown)


keys_test.go
```diff
@@ -0,0 +1,86 @@
...
+			AvatarURL:        "http://www.gravatar.com/avatar/cfa35b8cd2ec278026357769582fa563?s=40\u0026d=identicon",
+			CanCreateGroup:   true,
+			CanCreateProject: true,
+			ProjectsLimit:    10,
+			TwoFactorEnabled: false,
+			Identities:       []*UserIdentity{},
+			External:         false,
+			PublicEmail:      "john@example.com",
+		},
+	}
```
svanharmelen: I'm not sure if tests like this add much value as they don't "touch" a real API. So the struct and the JSON response are written right next to each other and any changes to the API will not be caught by test like this.

pmwebster: That's a good point. I can't think of a better way to test something this simple, so unless you have a suggestion, feel free to close this PR.

svanharmelen: I guess the only tests that really make sense would be to wire the tests up to a container running Gitlab and then make actual requests to that throw away Gitlab instance.

Yet setting that up would require quite a bit more effort... But I'll merge this one as it doesn't hurt to have it in.

PR #499, opened 2018-10-19
Removing V3 Sessions API

stjohnjohnson:
Sessions is a V3-only API and should probably no longer be included in this library.

The last reference I can find is in the V3 docs here: https://gitlab.com/gitlab-org/gitlab-ce/blob/8-16-stable/doc/api/session.md


svanharmelen:
Thanks @stjohnjohnson! I think it's fine to remove the endpoint as there is a stale V3 branch for people who still need to work with the V3 API. So LGTM!

PR #506, opened 2018-10-30
Add MergeBase repo method.

### LINKED ISSUE #505: About New API Merge Base
https://docs.gitlab.com/ee/api/repositories.html#merge-base

I want to use this api. Can you support it?

svanharmelen:
Fixes #505

### DIFF
--- a/repositories.go
+++ b/repositories.go
@@ -230,8 +230,7 @@ func (c Contributor) String() string {
 	return Stringify(c)
 }
 
-// ListContributorsOptions represents the available ListContributorsOptions()
-// options.
+// ListContributorsOptions represents the available ListContributors() options.
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/repositories.html#contributors
 type ListContributorsOptions ListOptions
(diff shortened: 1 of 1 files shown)


yodark1995:
Have you tested it?

This Api should have more args?

I tried put two branch name in `[]Ref`,then `400 {message: Could not find refs: master, test}`

svanharmelen:
@yodark1995 no I didn't test it, that's why I asked you if this would work for you. Mind you that I don't need this method myself, but thought to help you out by adding a quick implementation.

Looking at the error message, that seems to come from the Gitlab API. So what happens when you try this: 

```
curl --header "PRIVATE-TOKEN: <YOUR_TOKEN>" "https://<YOUR_GITLAB_INSTANCE>/api/v4/projects/<YOUR_PROJECT_ID>/repository/merge_base?refs[]=master&refs[]=test"
```

Does that give the expected result, or do you get the same error?

yodark1995:
@svanharmelen sorry.It was a problem with my environment just now.
It's good for me. I retested it.
Thank you very much.

PR #511, opened 2018-11-09
add new endpoint HEAD get file metadata

bakhdaulet:


repository_files.go
```diff
@@ -48,6 +49,9 @@ func (r File) String() string {
 	return Stringify(r)
 }
 
+
+
+
```
svanharmelen: Can you remove these 3 blank lines?

repository_files.go
```diff
@@ -82,6 +86,54 @@ func (s *RepositoryFilesService) GetFile(pid interface{}, fileName string, opt *
 	return f, resp, err
 }
 
+
+
+
+
```
svanharmelen: Can you remove these 4 blank lines?

repository_files.go
```diff
@@ -82,6 +86,54 @@ func (s *RepositoryFilesService) GetFile(pid interface{}, fileName string, opt *
...
+// GetFileOptions represents the available GetFile() options.
```
svanharmelen: This should be `GetFileMetaOptions`

repository_files.go
```diff
@@ -82,6 +86,54 @@ func (s *RepositoryFilesService) GetFile(pid interface{}, fileName string, opt *
...
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/repository_files.html#get-file-from-repository
+type GetFileMetaOptions struct {
+	Ref *string `url:"ref,omitempty" json:"ref,omitempty"`
+}
+
+// GetFile allows you to receive information about a file in repository like
```
svanharmelen: This should be `GetFileMeta`, but maybe we should call them (this one and the options struct `GetFileMetaData`? Sounds better to me...

repository_files.go
```diff
@@ -82,6 +86,54 @@ func (s *RepositoryFilesService) GetFile(pid interface{}, fileName string, opt *
...
+	}
+
+	f.BlobID = resp.Header.Get("X-Gitlab-Blob-Id")
+	f.CommitID = resp.Header.Get("X-Gitlab-Last-Commit-Id")
+	f.Encoding = resp.Header.Get("X-Gitlab-Encoding")
+	f.FileName = resp.Header.Get("X-Gitlab-File-Name")
+	f.FilePath = resp.Header.Get("X-Gitlab-File-Path")
+	f.Ref = resp.Header.Get("X-Gitlab-Ref")
+	if sizeString := resp.Header.Get("X-Gitlab-Size"); sizeString != "" {
+		f.Size,_ = strconv.Atoi(sizeString)
```
svanharmelen: Could you use `gofmt` to format your code? And are you sure we can safely ignore the error here?

repository_files.go
bakhdaulet: removed

repository_files.go
bakhdaulet: removed

repository_files.go
bakhdaulet: changed

repository_files.go
bakhdaulet: changed

repository_files.go
bakhdaulet: added check error

svanharmelen:
Thanks @bakhdaulet! LGTM!

PR #513, opened 2018-11-10
Add ParseWebhook function to automatically parse webhook requests

adamrothman:
Google's [go-github library](https://github.com/google/go-github) has this really nice functionality that does all the heavy lifting of parsing webhook requests into the corresponding struct type: https://github.com/google/go-github/blob/99760a16213d6fdde13f4e477438f876b6c9c6eb/github/messages.go#L237

I'm working on an integration for both GitHub and GitLab, and ended up having to write all this logic myself for the GitLab side of things. IMO it's super useful, so I thought I'd upstream it!

Here's how you use it:

```go
func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
    payload, err := ioutil.ReadAll(r.Body)
    if err != nil { ... }
    event, err := gitlab.ParseWebhook(gitlab.WebhookType(r), payload)
    if err != nil { ... }
    switch event := event.(type) {
        case *gitlab.PushEvent:
            processPushEvent(event)
        case *gitlab.MergeEvent:
            processMergeEvent(event)
        ...
    }
}
```

GitHub's webhooks are... kind of terrible. So there's a bit of gymnastics involved, e.g. in figuring out exactly which kind of comment we received. But it works, and I wrote tests.

### DIFF
--- a/event_parsing.go
+++ b/event_parsing.go
@@ -1,0 +1,114 @@
+package gitlab
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+const eventTypeHeader = "X-Gitlab-Event"
+
+// WebhookType returns the event type for the given request.
+func WebhookType(r *http.Request) string {
+	return r.Header.Get(eventTypeHeader)
+}
+
+// ParseWebhook parses the event payload. For recognized event types, a
+// value of the corresponding struct type will be returned. An error will
+// be returned for unrecognized event types.
+//
+// Example usage:
+//
+//     func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+//       payload, err := ioutil.ReadAll(r.Body)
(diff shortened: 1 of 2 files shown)


adamrothman:
Sure! I like that change. Update coming shortly.

event_parsing.go
```diff
@@ -0,0 +1,135 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+const eventTypeHeader = "X-Gitlab-Event"
+
+// webhookEventType returns the event type for the given request.
+func webhookEventType(r *http.Request) string {
```
svanharmelen: I think this should be public?

adamrothman: Oops! Overly aggressive symbol renaming.

svanharmelen:
Thanks @adamrothman! LGTM!

PR #514, opened 2018-11-12
Tweak the webhook parsing logic a bit

svanharmelen:


event_parsing.go
```diff
@@ -6,11 +6,47 @@ import (
...
+	EventTypeNote         EventType = "Note Hook"
+	EventTypePipeline     EventType = "Pipeline Hook"
+	EventTypePush         EventType = "Push Hook"
+	EventTypeTagPush      EventType = "Tag Push Hook"
+	EventTypeWikiPage     EventType = "Wiki Page Hook"
+)
+
+// RawWebhookEvent represents a raw event containing
+// both the event type and raw payload.
+type RawWebhookEvent struct {
```
adamrothman: I don't know if this needs to be exported – I can't think of a situation in which an end user would want to use this struct directly.

svanharmelen: Maybe you're right... But looking at it again, if this is a private type that is only used for this one single thing I don't think we need the type at all. It doesn't give any additional functionality or options, so I'll just remove it.

PR #515, opened 2018-11-14
Fix adding or updating discussion notes

sirlatrom:
There were a number of mistakes in my previous PR #481 leading to expecting wrong return types. This PR changes the expected return type from Discussion to Note in all Add or Update endpoints for discussion notes.

Signed-off-by: Sune Keller <absukl@almbrand.dk>


svanharmelen:
Thanks the fixes @sirlatrom! 👍

Issue #516, opened 2018-11-15
The properties in the GroupMember struct are not in-sync with the API response


praneetloke:
Looks like GitLab has updated the API response. Source: https://docs.gitlab.com/ce/api/members.html. The `email` and `created_at` fields are no longer available and instead `avatar_url` and `web_url` are now returned in the response to listing a group's members.

svanharmelen:
There's not a real schedule, so mainly when people ask or when I think enough has been changed/added.

So since you asked... v0.11.7 is just released :wink:

PR #517, opened 2018-11-15
Fix for issue #516.

praneetloke:
Update the `GroupMember` struct with the most recent properties from https://docs.gitlab.com/ce/api/members.html.

### DIFF
--- a/group_members.go
+++ b/group_members.go
@@ -19,7 +19,6 @@ package gitlab
 import (
 	"fmt"
 	"net/url"
-	"time"
 )
 
 // GroupMembersService handles communication with the group members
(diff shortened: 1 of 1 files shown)


group_members.go
```diff
@@ -36,12 +35,12 @@ type GroupMembersService struct {
...
 	ID          int              `json:"id"`
 	Username    string           `json:"username"`
-	Email       string           `json:"email"`
 	Name        string           `json:"name"`
 	State       string           `json:"state"`
-	CreatedAt   *time.Time       `json:"created_at"`
 	AccessLevel AccessLevelValue `json:"access_level"`
 	ExpiresAt   *ISOTime         `json:"expires_at"`
+	AvatarURL   string           `json:"avatar_url"`
+	WebURL      string           `json:"web_url"`
```
svanharmelen: One small question... Could you reorder the last four lines so they match the order in the docs? I'm trying to follow the docs ordering a bit so it's easier to spot missing or new entries over time. So that would be:

```
	AvatarURL   string           `json:"avatar_url"`
	WebURL      string           `json:"web_url"`
	ExpiresAt   *ISOTime         `json:"expires_at"`
	AccessLevel AccessLevelValue `json:"access_level"`
```

praneetloke: Yep. Absolutely. I will do that.

praneetloke:
@svanharmelen done.

svanharmelen:
Thanks for the additional update @praneetloke! LGTM!

Issue #519, opened 2018-11-19
Duplicated variables in ProjectVariablesService.UpdateVariable


killmeplz:
```
func (s *ProjectVariablesService) UpdateVariable(pid interface{}, key string, opt *UpdateVariableOptions, options ...OptionFunc) (*ProjectVariable, *Response, error) {
	project, err := parseID(pid)
	if err != nil {
		return nil, nil, err
	}
	u := fmt.Sprintf("projects/%s/variables/%s",
		url.QueryEscape(project),
		url.QueryEscape(key),
	)
```
You are accepting 2 "key" variables. In function and as a parameter of UpdateVariableOptions.
Can we use just one instead of two? 
Does it make sense or am I missing something?

svanharmelen:
I think you are right @killmeplz, I think this is a bug/error in the code and that the `key` field from the options struct should be removed.

@pmwebster as you added this code, do you agree? I don't have time to test or validate this myself, but it looks like we should remove the `key` field.

pmwebster:
Absolutely! 

Thanks for catching it @killmeplz. I can get a PR together in a little bit, I'm just now waking up for the day.

svanharmelen:
Ah, time zones 😏 But thanks and no rush @pmwebster!

pmwebster:
So after waking up a little bit and re-reading, this is actually correct according to the gitlab docs: https://docs.gitlab.com/ee/api/project_level_variables.html#update-variable

The key of the variable can change, so you have to post the old one as well potentially the new one.

I did realize I missed a comment when I was playing with this originally though, pr has been submitted.

svanharmelen:
I noticed that as well, but in the docs the URL that is used to get that response is this:

```
curl --request PUT --header "PRIVATE-TOKEN: 9koXpg98eAheJpvBs5tK" "https://gitlab.example.com/api/v4/projects/1/variables/NEW_VARIABLE" --form "value=updated value"
```

So there the `NEW_VARIABLE` value did not change (and that same name was actually used in other examples as well, like in the create right above the update endpoint).

I'll go ahead and test this against the API in a sec, but I don't think the key is expected (and so used) in the body.

svanharmelen:
Just confirmed, the `Key` field in the option struct is ignored by the API and so should be removed from the options struct:

```go
func main() {
	git := gitlab.NewClient(nil, "my-token")

[...]

PR #521, opened 2018-11-19
project_variables.go: removing Key from UpdateOptions struct

pmwebster:


svanharmelen:
Fixes #519 

Thanks @pmwebster!

PR #522, opened 2018-11-25
Add public_jobs setting for Create/UpdateProject

dvulpe:


projects.go
```diff
@@ -419,6 +419,7 @@ type CreateProjectOptions struct {
 	Visibility                                *VisibilityValue  `url:"visibility,omitempty" json:"visibility,omitempty"`
 	ImportURL                                 *string           `url:"import_url,omitempty" json:"import_url,omitempty"`
 	PublicJobs                                *bool             `url:"public_jobs,omitempty" json:"public_jobs,omitempty"`
+	PublicBuilds                              *bool             `url:"public_builds,omitempty" json:"public_builds,omitempty"`
```
svanharmelen: Looking at https://gitlab.com/gitlab-org/gitlab-ce/commit/0ab4918cc7cf1c6fe6e1af5efa7285267129cb78 I think this should replace the `PublicJobs` fields instead of adding it as an additional field. Right?

PR #526, opened 2018-12-02
Added unit test for issues and settings

Snehal1112:


issues.go
```diff
@@ -33,62 +33,67 @@ type IssuesService struct {
...
+// IssueAssignees represents collections of assignees of the issue.
+type IssueAssignees []IssueAssignee
+
+// IssueLinks represents links of the issue.
+type IssueLinks struct {
+	Self       string `json:"self"`
+	Notes      string `json:"notes"`
+	AwardEmoji string `json:"award_emoji"`
+	Project    string `json:"project"`
+}
```
svanharmelen: I'm not sure if there is much value in specifying these embedded struct as specific types, but I guess it doesn't hurt either.

The only one that I would leave out is `IssueAssignees []IssueAssignee`. That one really doesn't add any value as `[]IssueAssignee` already is a specific type.

issues.go
```diff
@@ -33,62 +33,67 @@ type IssuesService struct {
...
-	} `json:"_links"`
-	IssueLinkID int `json:"issue_link_id"`
+	ID               int             `json:"id"`
+	IID              int             `json:"iid"`
+	ProjectID        int             `json:"project_id"`
+	Milestone        *Milestone      `json:"milestone"`
+	Author           *IssueAuthor    `json:"author"`
+	Description      string          `json:"description"`
+	State            string          `json:"state"`
+	Assignees        *IssueAssignees `json:"assignees"`
```
svanharmelen: This should be `[]*IssueAssignee`

issues.go
Snehal1112: I observer that it helps a lot in writing a unit test so just change struct to specific struct type but ya i can replace ```IssueAssignees []IssueAssignee``` to ```[]*IssueAssignee```

issues.go
Snehal1112: fixed

svanharmelen:
Thanks @Snehal1112! LGMT!

PR #527, opened 2018-12-13
fix godoc for RepositoriesService.RawFileContent

0xmohit:
As an aside, would it be more clear if `RawFileContent` was named `Blob` instead?


svanharmelen:
Thanks @0xmohit! And yes, I think we can change it to `GetBlob`. Do you want to add that to this PR?

0xmohit:
Done.

svanharmelen:
LGTM! Thanks @0xmohit!

PR #528, opened 2018-12-14
Add project level MergeRequest API

bilby91:


bilby91:
Anything I can do to speed up the review of this one @svanharmelen :) ?

Looking forward to merge this MR in order to add support for this configurations on https://github.com/terraform-providers/terraform-provider-gitlab

merge_request_approvals.go
```diff
@@ -80,6 +108,58 @@ type ApproveMergeRequestOptions struct {
...
+	}
+	u := fmt.Sprintf("projects/%s/approvals", url.QueryEscape(project))
+
+	req, err := s.client.NewRequest("POST", u, opt, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	mrp := new(MergeRequestProjectApprovals)
+
```
svanharmelen: Renaming this to just `m` and removing the blank line would make the code more inline with the rest of the package.

merge_request_approvals.go
```diff
@@ -80,6 +108,58 @@ type ApproveMergeRequestOptions struct {
...
+	}
+	u := fmt.Sprintf("projects/%s/approvers", url.QueryEscape(project))
+
+	req, err := s.client.NewRequest("POST", u, opt, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	mrp := new(MergeRequestProjectApprovals)
+
```
svanharmelen: Renaming this to just `m` and removing the blank line would make the code more inline with the rest of the package.

merge_request_approvals.go
```diff
@@ -80,6 +108,58 @@ type ApproveMergeRequestOptions struct {
 	SHA *string `url:"sha,omitempty" json:"sha,omitempty"`
 }
 
+// UpdateMergeRequestProjectApprovals updates the approval configurations for a given project on GitLab.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/merge_request_approvals.html#change-configuration
+func (s *MergeRequestApprovalsService) UpdateMergeRequestProjectApprovals(pid interface{}, opt *UpdateMergeRequestProjectApprovalsOptions, options ...OptionFunc) (*MergeRequestProjectApprovals, *Response, error) {
```
svanharmelen: I would suggest to call this `ConfigureProjectMRApprovals`

merge_request_approvals.go
```diff
@@ -80,6 +108,58 @@ type ApproveMergeRequestOptions struct {
...
+	}
+
+	return mrp, resp, err
+}
+
+// UpdateMergeRequestProjectApprovers updates the list of approvers for a given project on GitLab.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/merge_request_approvals.html#change-allowed-approvers
+func (s *MergeRequestApprovalsService) UpdateMergeRequestProjectApprovers(pid interface{}, opt *UpdateMergeRequestProjectApproversOptions, options ...OptionFunc) (*MergeRequestProjectApprovals, *Response, error) {
```
svanharmelen: I would suggest to call this `ChangeProjectMRApprovers`

merge_request_approvals_test.go
```diff
@@ -0,0 +1,153 @@
+package gitlab
+
+import (
+	"fmt"
+	"net/http"
+	"reflect"
+	"testing"
+)
+
+const exampleResponse = `{
```
svanharmelen: This needs a more specific name as this is to generic and would cause name collisions when used by all tests.

merge_request_approvals_test.go
```diff
@@ -0,0 +1,153 @@
...
+				"full_path": "group1"
+			}
+		}
+	],
+	"approvals_before_merge": 2,
+	"reset_approvals_on_push": true,
+	"disable_overriding_approvers_per_merge_request": false	
+}`
+
+func exampleParsedResponse() *MergeRequestProjectApprovals {
```
svanharmelen: Same for this name.

merge_request_approvals_test.go
```diff
@@ -0,0 +1,153 @@
...
+		},
+	}
+}
+func TestUpdateMergeRequestProjectApprovals(t *testing.T) {
+	mux, server, client := setup()
+	defer teardown(server)
+
+	mux.HandleFunc("/api/v4/projects/1/approvals", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "POST")
+		testBody(t, r, `{"approvals_before_merge":3}`)
```
svanharmelen: I think this is the only line of the test that actually test something. The rest of this function only tests what is defined in this same file. So I wonder if we should add these tests?

svanharmelen: Yeah I get that, but both the raw JSON and the type are defined in this package and so if the API changes, our test will still pass but actual calls would fail.

I would leave out at least the `exampleParsedResponse` and add the raw JSON inside the method body of the test functions like done in other tests in this package. If you then specify just a handful of fields and check those fields in the returned response, it would be a good indication that the basic setup works.

bilby91:
@svanharmelen Made the requested changes! 

Ended up moving the calls to the `projects.go` file since it made sense. I kept some of the structures under `merge_request_approvals.go` since the other api was also reusing some of the structs. 

Let me know if it looks good now or you want any further change.

svanharmelen:
Thanks @bilby91! I'm going to make one or two final small tweaks before cutting a new release, but this looks very nice now!

PR #532, opened 2018-12-21
handle paginated results that are only one page

mikecook:


svanharmelen:
Good catch @mikecook! Thanks!

PR #493, opened 2018-10-19
Added missing SHA option in ListProjectPipelines API method.

tgregory:
This option allows specific commit pipeline selection.

PR #535, opened 2018-12-22
Fix inconsistently cased pipeline sha

jamesramsay:
Added in https://github.com/xanzy/go-gitlab/pull/493 the capitalization of `SHA` is inconsistent with the rest of the Pipelines methods.

There is also inconsistency in other files `Sha` appears to be used more frequently. I can open a more comprehensive merge request to update this elsewhere if desired. Both this and a sweeping change would be a breaking change.


svanharmelen:
Hi @jamesramsay, according to the Go review guidelines it should be `SHA`. So if we want to make this consistent (which I like!), we should update any occurences of `Sha` to `SHA`.

PR #543, opened 2019-01-07
Add support for protected tags

ailsachiu:
# Objective
This PR adds initial support for the [protected tags API](https://docs.gitlab.com/ee/api/protected_tags.html) with basic unit tests.

# Context
I'd like to use this library for handling protected tags for a repo. Most of the implementation/styling comes from https://github.com/xanzy/go-gitlab/blob/master/protected_branches.go.

### DIFF
--- a/protected_tags.go
+++ b/protected_tags.go
@@ -1,0 +1,141 @@
+package gitlab
+
+import (
+	"fmt"
+	"net/url"
+)
+
+// ProtectedTagsService handles communication with the protected tag methods of the GitLab API.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/protected_tags.html
+type ProtectedTagsService struct {
+	client *Client
+}
+
+// TagAccessDescription reperesents the access decription for a protected tag.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/protected_tags.html#list-protected-tags
+type TagAccessDescription struct {
+	AccessLevel            AccessLevelValue `json:"access_level"`
+	AccessLevelDescription string           `json:"access_level_description"`
+}
+
+// ProtectedTag represents a protected tag.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/protected_tags.html#list-protected-tags
+type ProtectedTag struct {
+	Name               string                  `json:"name"`
+	CreateAccessLevels []*TagAccessDescription `json:"create_access_levels"`
+}
+
+// ListProtectedTagsOptions represents available ListProtectedTags() options.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/protected_tags.html#list-protected-tags
+type ListProtectedTagsOptions ListOptions
+
+// ListProtectedTags returns a list of protected tags from a project.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/protected_tags.html#list-protected-tags
+func (s *ProtectedTagsService) ListProtectedTags(pid interface{}, opt *ListProtectedTagsOptions, options ...OptionFunc) ([]*ProtectedTag, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/protected_tags", url.QueryEscape(project))
+
+	req, err := s.client.NewRequest("GET", u, opt, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var p []*ProtectedTag
+	resp, err := s.client.Do(req, &p)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return p, resp, err
+}
+
+// GetProtectedTag returns a single protected tag or wildcard protected tag.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/protected_tags.html#get-a-single-protected-tag-or-wildcard-protected-tag
+func (s *ProtectedTagsService) GetProtectedTag(pid interface{}, tag string, options ...OptionFunc) (*ProtectedTag, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


protected_tags.go
```diff
@@ -0,0 +1,141 @@
+package gitlab
+
+import (
+	"fmt"
+	"net/url"
+)
+
+// ProtectedTagsService handles communication with the protected tag methods of the GitLab API.
```
svanharmelen: Please try to keep comments below 80 chars. Some doc URLs are longer and that is fine, but normal comments should not exceed 80 chars.

protected_tags.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// TagAccessDescription reperesents the access decription for a protected tag.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/protected_tags.html#list-protected-tags
+type TagAccessDescription struct {
+	AccessLevel            AccessLevelValue `json:"access_level"`
+	AccessLevelDescription string           `json:"access_level_description"`
+}
```
svanharmelen: I'm not sure if there is much value in defining this as a specific type. So I prefer to add these 2 fields directly to `ProtectedTag` using an inline struct.

protected_tags.go
```diff
@@ -0,0 +1,141 @@
...
+
+// ProtectedTag represents a protected tag.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/protected_tags.html#list-protected-tags
```
svanharmelen: The URL should probably omit `#list-protected-tags` as this is the generic type that is used by all endpoints.

protected_tags.go
```diff
@@ -0,0 +1,141 @@
...
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/protected_tags", url.QueryEscape(project))
+
+	req, err := s.client.NewRequest("GET", u, opt, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var p []*ProtectedTag
```
svanharmelen: Maybe `pt` is a better resemblance for `ProtectedTag`? And then this should become `pts` to indicate is are multiple tags.

protected_tags.go
```diff
@@ -0,0 +1,141 @@
...
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/protected_tags/%s", url.QueryEscape(project), tag)
+
+	req, err := s.client.NewRequest("GET", u, nil, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	p := new(ProtectedTag)
```
svanharmelen: Maybe `pt` is a better resemblance for `ProtectedTag`?

protected_tags.go
```diff
@@ -0,0 +1,141 @@
...
+
+	return p, resp, err
+}
+
+// ProtectRepositoryTagsOptions represents available ProtectRepositoryTags() options.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/protected_tags.html#protect-repository-tags
+type ProtectRepositoryTagsOptions struct {
+	Name              string            `url:"name,omitempty" json:"name,omitempty"`
```
svanharmelen: All the fields in any option struct should be pointers (unless the zero value of the type is `nil`). So this should be `*string`. And since this is a required field I think we can omit the `omitempty`?

protected_tags.go
```diff
@@ -0,0 +1,141 @@
...
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/protected_tags", url.QueryEscape(project))
+
+	req, err := s.client.NewRequest("POST", u, opt, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	p := new(ProtectedTag)
```
svanharmelen: Maybe `pt` is a better resemblance for `ProtectedTag`?

protected_tags.go
ailsachiu: Thanks for the feedback! Moving `TagAccessDescription` into an anonymous struct inside `ProtectedTag` requires adding some duplicate definition in the unit tests due to how you define anonymous structs:

```
type ProtectedTag struct {
	Name               string `json:"name"`
	CreateAccessLevels []struct {
		AccessLevel            AccessLevelValue `json:"access_level"`
		AccessLevelDescription string           `json:"access_level_description"`
	} `json:"create_access_levels"`
}
```

Example updated unit test:
```
func TestListProtectedTags(t *testing.T) {
	mux, server, client := setup()
	defer teardown(server)

	mux.HandleFunc("/api/v4/projects/1/protected_tags", func(w http.ResponseWriter, r *http.Request) {
		testMethod(t, r, "GET")
		fmt.Fprint(w, `[{"name":"1.0.0", "create_access_levels": [{"access_level": 40, "access_level_description": "Maintainers"}]},{"name":"*-release", "create_access_levels": [{"access_level": 30, "access_level_description": "Developers + Maintainers"}]}]`)
	})

	expected := []*ProtectedTag{
		{
			Name: "1.0.0",
			CreateAccessLevels: []struct {
				AccessLevel            AccessLevelValue `json:"access_level"`
				AccessLevelDescription string           `json:"access_level_description"`
			}{
				{
					AccessLevel:            40,
					AccessLevelDescription: "Maintainers",
				},
			},
		},
		{
			Name: "*-release",
			CreateAccessLevels: []struct {
				AccessLevel            AccessLevelValue `json:"access_level"`
				AccessLevelDescription string           `json:"access_level_description"`
			}{
				{
					AccessLevel:            30,
					AccessLevelDescription: "Developers + Maintainers",
				},
			},
		},
	}

	opt := &ListProtectedTagsOptions{}
	tags, _, err := client.ProtectedTags.ListProtectedTags(1, opt)
	assert.NoError(t, err, "failed to get response")
	assert.Equal(t, expected, tags)
}
``` 

Is this what you mean by an inline struct? If so, and if this duplication in the tests is fine, I don't mind making this change.

PR #545, opened 2019-01-11
Update service jira: fix JiraIssueTransitionID type from string to int

jorcau:
Hi there,

This is a tiny fix for the JIRA service.

JiraIssueTransitionID parameter is currently a string when it should be an int, as mentioned in the [doc](https://docs.gitlab.com/ce/api/services.html#jira).

This currently creates error when unmarshalling the Json into the Go struct:
`json: cannot unmarshal number into Go struct field JiraServiceProperties.jira_issue_transition_id of type string`

Thanks,


svanharmelen:
LGTM!

PR #547, opened 2019-01-20
Implement UnmarshalJSON for Jira Service Properties

fourplusone:
The Gitlab API returns under some circumstances JiraIssueTransitionID as a string and sometimes as an integer (although not documented). This PR makes sure, that it's always converted to int.

### DIFF
--- a/services.go
+++ b/services.go
@@ -17,8 +17,11 @@
 package gitlab
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/url"
+	"reflect"
+	"strconv"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


svanharmelen:
@fourplusone could you maybe point me to some Gitlab docs that describe this behavior? As it very much feels like a bug in there API. So curious to know the details before merging a fix like this.

fourplusone:
Hi @svanharmelen 
 
The current docs do not specify any explicit return type, although one might guess that they are the same as the PUT Parameters.  https://docs.gitlab.com/ee/api/services.html#get-jira-service-settings In practice, the API returns strings right now, at least at the public gitlab instance.
 
It seems like it was changed some time ago: https://gitlab.com/gitlab-org/gitlab-ce/commit/593c912151eaef865d31fc2a8307ef6d337c2349#bb2ba7ca0e10bd01609ab50236882ea82a183e60_484_517

svanharmelen:
I think we should wait with trying to solve it in here, until it is solved in the Gitlab API. And then follow the direction they used to fix it (so probably either force it to be a string or just update the field to be of type `*string`).

jorcau:
Hi there, since 11.9 release, **JiraIssueTransitionID** will always be a string cf. (https://gitlab.com/gitlab-org/gitlab-ce/blob/master/CHANGELOG.md#L443).

Should we simply change this parameter from int to string or implement UnmarshalJSON to make the parameter compatible with previous Gitlab releases?

svanharmelen:
I think adding `UnmarshalJSON` would be the best way forward here.

svanharmelen:
Ow, I noticed this PR is actually an earlier version of #623 so I guess I should have merged this one instead. Sorry about that @fourplusone!

But the good new is that the functionality is now merged, so you should no longer have this problem. Again sorry for not merging this one instead!

Issue #546, opened 2019-01-13
Support for the upcoming releases and release links apis


juranki:
Looks like Gitlab 11.7 will promote [Releases](https://docs.gitlab.com/ee/user/project/releases/index.html) from a tag extension to a top level feature. 

Two new apis are added for managing releases:
- [Releases](https://docs.gitlab.com/ee/api/releases/index.html), and
- [Release links](https://docs.gitlab.com/ee/api/releases/links.html)

It would be great if go-gitlab would cover the new API's

juranki:
I started working [towards a PR](https://github.com/xanzy/go-gitlab/compare/master...juranki:releases) and immediately bumped into a naming issue.

Tags api has also some support releases and defines a [Release](https://github.com/xanzy/go-gitlab/blob/master/tags.go#L45) type.

Releases api adds a lot more content to the same concept. I named it [ProjectRelease](https://github.com/juranki/go-gitlab/blob/releases/releases.go#L20), for now. Should I continue with this, or do you have better ideas?

svanharmelen:
Hi @juranki, I wonder if we could use the same struct in both places? Guess that depends at what fields the API is able to return when requesting a tag, so maybe we should check the Gitlab source code to see if that would work.

If not, I think I feel more for renaming the existing `Release` struct to `TagRelease`. Or maybe even better, just make it an inline struct if it only contains the `TagName` and `Description` fields. I very much doubt if anyone is using those struct directly, so that should be a very low impacting breaking change.

But let's first check is they maybe share the same fields and that one struct can be used in both places.

svanharmelen:
Thanks for the PR @juranki!

PR #551, opened 2019-01-26
Add support for Releases and Release links

juranki:
Add support for the new [Releases](https://docs.gitlab.com/ce/api/releases/index.html) and [Release links](https://docs.gitlab.com/ee/api/releases/links.html) apis (#546).

Naming conflicts with overlapping functionality in TagsService were resolved by prefixing release related types in TagsService with `Tag`.

[Tags api documentation](https://docs.gitlab.com/ee/api/tags.html) or [11.7 release note](https://about.gitlab.com/2019/01/22/gitlab-11-7-released/#release-deprecations) don't mention deprecations in tags api, but based on *tags.rb* (eg. lines [48](https://gitlab.com/gitlab-org/gitlab-ce/blob/c2b3f64cb44114e69ddeeb453412ad4dd762baf9/lib/api/tags.rb#L48), [108](https://gitlab.com/gitlab-org/gitlab-ce/blob/c2b3f64cb44114e69ddeeb453412ad4dd762baf9/lib/api/tags.rb#L108), [140](https://gitlab.com/gitlab-org/gitlab-ce/blob/c2b3f64cb44114e69ddeeb453412ad4dd762baf9/lib/api/tags.rb#L140)...) deprecation comments were added to Release related parts of TagsService.

### DIFF
--- a/releaselinks.go
+++ b/releaselinks.go
@@ -1,0 +1,177 @@
+package gitlab
+
+import (
+	"fmt"
+	"net/url"
+)
+
+// ReleaseLinksService handles communication with the release link methods
+// of the GitLab API.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/releases/links.html
+type ReleaseLinksService struct {
+	client *Client
+}
+
+// ReleaseLink represents a release link.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/releases/links.html
+type ReleaseLink struct {
+	ID       int    `json:"id"`
+	Name     string `json:"name"`
+	URL      string `json:"url"`
+	External bool   `json:"external"`
+}
+
+// ListReleaseLinksOptions represents ListReleaseLinks() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/releases/links.html#get-links
+type ListReleaseLinksOptions ListOptions
+
+// ListReleaseLinks gets assets as links from a Release.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/releases/links.html#get-links
+func (s *ReleaseLinksService) ListReleaseLinks(pid interface{}, tagName string, opt *ListReleaseLinksOptions, options ...OptionFunc) ([]*ReleaseLink, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/releases/%s/assets/links", url.QueryEscape(project), tagName)
+
+	req, err := s.client.NewRequest("GET", u, opt, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	rls := make([]*ReleaseLink, 0)
+	resp, err := s.client.Do(req, &rls)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return rls, resp, err
+}
+
+// GetReleaseLink returns a link from release assets.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/releases/links.html#get-a-link
+func (s *ReleaseLinksService) GetReleaseLink(pid interface{}, tagName string, link int, options ...OptionFunc) (*ReleaseLink, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/releases/%s/assets/links/%d",
+		url.QueryEscape(project),
+		tagName,
+		link)
+
+	req, err := s.client.NewRequest("GET", u, nil, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	rl := new(ReleaseLink)
+	resp, err := s.client.Do(req, rl)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return rl, resp, err
+}
+
+// CreateReleaseLinkOptions represents CreateReleaseLink() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/releases/links.html#create-a-link
+type CreateReleaseLinkOptions struct {
+	Name *string `url:"name" json:"name"`
+	URL  *string `url:"url" json:"url"`
+}
+
+// CreateReleaseLink creates a link.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/releases/links.html#create-a-link
+func (s *ReleaseLinksService) CreateReleaseLink(pid interface{}, tagName string, opt *CreateReleaseLinkOptions, options ...OptionFunc) (*ReleaseLink, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/releases/%s/assets/links", url.QueryEscape(project), tagName)
+
+	req, err := s.client.NewRequest("POST", u, opt, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	rl := new(ReleaseLink)
+	resp, err := s.client.Do(req, rl)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return rl, resp, err
+}
+
+// UpdateReleaseLinkOptions represents UpdateReleaseLink() options.
+//
+// You have to specify at least one of Name of URL.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/releases/links.html#update-a-link
+type UpdateReleaseLinkOptions struct {
+	Name *string `url:"name,omitempty" json:"name,omitempty"`
+	URL  *string `url:"url,omitempty" json:"url,omitempty"`
+}
+
+// UpdateReleaseLink updates an asset link.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/releases/links.html#update-a-link
+func (s *ReleaseLinksService) UpdateReleaseLink(pid interface{}, tagName string, link int, opt *UpdateReleaseLinkOptions, options ...OptionFunc) (*ReleaseLink, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/releases/%s/assets/links/%d",
+		url.QueryEscape(project),
+		tagName,
+		link)
+
+	req, err := s.client.NewRequest("PUT", u, opt, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	rl := new(ReleaseLink)
+	resp, err := s.client.Do(req, rl)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return rl, resp, err
+}
+
+// DeleteReleaseLink deletes a link from release.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/releases/links.html#delete-a-link
+func (s *ReleaseLinksService) DeleteReleaseLink(pid interface{}, tagName string, link int, options ...OptionFunc) (*ReleaseLink, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/releases/%s/assets/links/%d",
+		url.QueryEscape(project),
+		tagName,
+		link,
+	)
+
+	req, err := s.client.NewRequest("DELETE", u, nil, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	rl := new(ReleaseLink)
+	resp, err := s.client.Do(req, rl)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return rl, resp, err
+}
--- a/releases.go
+++ b/releases.go
@@ -1,0 +1,210 @@
+package gitlab
+
+import (
+	"fmt"
+	"net/url"
+	"time"
+)
+
+// ReleasesService handles communication with the releases methods
+// of the GitLab API.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/releases/index.html
+type ReleasesService struct {
+	client *Client
+}
+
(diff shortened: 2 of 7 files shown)


berniedurfee-ge:
Awesome to see this PR! Did you test against the gitlab.com API? I'm getting a `400 {error: assets[links][0][url] is missing}` error when I include links in the `Assets` field of `CreateReleaseOptions` when I call `CreateRelease`.

Here's the client that's getting the error. Am I misusing this feature or is there an issue?

https://gitlab.com/electric-it/cago/blob/master/magebuild.go#L218

juranki:
Hi @berniedurfee-ge,

I've tested against gitlab.com, but it looks like I've only tested the links using the release links endpoints.

I'll try to reproduce the problem, and find out what's wrong.

juranki:
@berniedurfee-ge 
I had put wrong tag on ReleaseAssetLink.URL. Added a fix to the pull request.

releaselinks.go
```diff
@@ -0,0 +1,167 @@
...
+		return nil, resp, err
+	}
+
+	return rls, resp, err
+}
+
+// GetReleaseLink returns a link from release assets.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/releases/links.html#get-a-link
+func (s *ReleaseLinksService) GetReleaseLink(pid interface{}, tag string, linkID int, options ...OptionFunc) (*ReleaseLink, *Response, error) {
```
svanharmelen: Following the conventions used in this package `tag` should be `tagName` and `linkID`, should just be `link`. This goes for all occurrences of course.

releaselinks.go
```diff
@@ -0,0 +1,167 @@
...
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/releases/%s/assets/links/%d", url.QueryEscape(project), tag, linkID)
+
+	req, err := s.client.NewRequest("GET", u, nil, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var rl ReleaseLink
```
svanharmelen: In other services we declare this as `rl := new(ReleaseLink)`. So good to do that here as well to keep things consistent.

releaselinks.go
```diff
@@ -0,0 +1,167 @@
...
+
+// DeleteReleaseLink deletes a link from release.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/releases/links.html#delete-a-link
+func (s *ReleaseLinksService) DeleteReleaseLink(pid interface{}, tag string, linkID int, options ...OptionFunc) (*ReleaseLink, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/releases/%s/assets/links/%d", url.QueryEscape(project), tag, linkID)
```
svanharmelen: For lines of code that are longer than 100 chars, I prefer to use multiple lines:

```suggestion
	u := fmt.Sprintf("projects/%s/releases/%s/assets/links/%d",
		url.QueryEscape(project),
		tag,
		linkID,
	)
```

releases.go
```diff
@@ -0,0 +1,202 @@
...
+// Release represents a project release.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/releases/index.html#list-releases
+type Release struct {
+	TagName         string     `json:"tag_name"`
+	Name            string     `json:"name"`
+	Description     string     `json:"description,omitempty"`
+	DescriptionHTML string     `json:"description_html,omitempty"`
+	CreatedAt       *time.Time `json:"created_at,omitempty"`
+	Author          *struct {
```
svanharmelen: Is there a reason why you made this a pointer?

releases.go
```diff
@@ -0,0 +1,202 @@
...
+		ID        int    `json:"id"`
+		Name      string `json:"name"`
+		Username  string `json:"username"`
+		State     string `json:"state"`
+		AvatarURL string `json:"avatar_url"`
+		WebURL    string `json:"web_url"`
+	} `json:"author,omitempty"`
+	Commit *Commit `json:"commit,omitempty"`
+	Assets *struct {
```
svanharmelen: Is there a reason why you made this a pointer?

releases.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return r, resp, err
+}
+
+// GetRelease returns a single release, identified by a tag name.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/releases/index.html#get-a-release-by-a-tag-name
```
svanharmelen:
```suggestion
// GitLab API docs:
// https://docs.gitlab.com/ce/api/releases/index.html#get-a-release-by-a-tag-name
```

releases.go
```diff
@@ -0,0 +1,202 @@
...
+func (s *ReleasesService) GetRelease(pid interface{}, tag string, options ...OptionFunc) (*Release, *Response, error) {
```
svanharmelen: Same in this file of course, `tag` should be `tagName` in order to remain consistent.

releases.go
```diff
@@ -0,0 +1,202 @@
...
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/releases/%s", url.QueryEscape(project), tag)
+
+	req, err := s.client.NewRequest("GET", u, nil, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var r *Release
```
svanharmelen: And for this as well, I prefer to use `r := new(Release)` for consistency.

releases.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return r, resp, err
+}
+
+// ReleaseAssets represents release assets in CreateRelease() options
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/releases/index.html#create-a-release
```
svanharmelen:
```suggestion
// GitLab API docs:
// https://docs.gitlab.com/ce/api/releases/index.html#create-a-release
```

releases.go
```diff
@@ -0,0 +1,202 @@
...
+type ReleaseAssets struct {
+	Links []ReleaseAssetLink `url:"links" json:"links"`
+}
+
+// ReleaseAssetLink represents release asset link in CreateRelease() options
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/releases/index.html#create-a-release
```
svanharmelen:
```suggestion
// GitLab API docs:
// https://docs.gitlab.com/ce/api/releases/index.html#create-a-release
```

releases.go
```diff
@@ -0,0 +1,202 @@
...
+
+// CreateReleaseOptions represents CreateRelease() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/releases/index.html#create-a-release
+type CreateReleaseOptions struct {
+	Name        string         `url:"name" json:"name"`
+	TagName     string         `url:"tag_name" json:"tag_name"`
+	Description string         `url:"description" json:"description"`
+	Ref         string         `url:"ref,omitempty" json:"ref,omitempty"`
+	Assets      *ReleaseAssets `url:"assets,omitempty" json:"assets,omitempty"`
```
svanharmelen: All fields in option structs need to be pointers.

releaselinks.go
```diff
@@ -0,0 +1,167 @@
...
+	return &rl, resp, err
+}
+
+// CreateReleaseLinkOptions represents CreateReleaseLink() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/releases/links.html#create-a-link
+type CreateReleaseLinkOptions struct {
+	Name string `url:"name" json:"name"`
+	URL  string `url:"url" json:"url"`
+}
```
svanharmelen: All fields in option structs need to be pointers.

releaselinks.go
```diff
@@ -0,0 +1,167 @@
...
+
+// UpdateReleaseLinkOptions represents UpdateReleaseLink() options.
+//
+// You have to specify at least one of Name of URL.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/releases/links.html#update-a-link
+type UpdateReleaseLinkOptions struct {
+	Name string `url:"name,omitempty" json:"name,omitempty"`
+	URL  string `url:"url,omitempty" json:"url,omitempty"`
+}
```
svanharmelen: All fields in option structs need to be pointers.

releases.go
```diff
@@ -0,0 +1,202 @@
...
+	return r, resp, err
+}
+
+// UpdateReleaseOptions represents UpdateRelease() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/releases/index.html#update-a-release
+type UpdateReleaseOptions struct {
+	Name        string `url:"name" json:"name"`
+	Description string `url:"description" json:"description"`
+}
```
svanharmelen: All fields in option structs need to be pointers.

releases_test.go
```diff
@@ -0,0 +1,343 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+	"testing"
+)
+
+const exampleReleaseListRsp = `[
```
svanharmelen: I think it would improve readability if we simplify these examples do they are more concise. For example the descriptions can be just one or two words instead of an actual full description.

tags.go
```diff
@@ -33,16 +33,16 @@ type TagsService struct {
...
+	Name    string      `json:"name"`
+	Message string      `json:"message"`
 }
 
-// Release represents a GitLab version release.
+// TagRelease represents a GitLab version release.
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/tags.html
-type Release struct {
+type TagRelease struct {
```
svanharmelen: I think I like `ReleaseTag` better. To me is sounds and reads a bit nicer. What do you think, would it fit this use case or is `TagRelease` better?

tags.go
```diff
@@ -168,20 +169,24 @@ func (s *TagsService) DeleteTag(pid interface{}, tag string, options ...OptionFu
...
 
 // CreateRelease Add release notes to the existing git tag.
 // If there already exists a release for the given tag, status code 409 is returned.
 //
+// Deprecated: This feature was deprecated in GitLab 11.7.
+//
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/tags.html#create-a-new-release
-func (s *TagsService) CreateRelease(pid interface{}, tag string, opt *CreateReleaseOptions, options ...OptionFunc) (*Release, *Response, error) {
+func (s *TagsService) CreateRelease(pid interface{}, tag string, opt *CreateTagReleaseOptions, options ...OptionFunc) (*TagRelease, *Response, error) {
```
svanharmelen: Option struct always use the same name as the functions. So in this case the function should be renamed to `CreateTagRelease` (or `CreateReleaseTag` if we like that better).

tags.go
```diff
@@ -202,19 +207,21 @@ func (s *TagsService) CreateRelease(pid interface{}, tag string, opt *CreateRele
...
 }
 
 // UpdateRelease Updates the release notes of a given release.
 //
+// Deprecated: This feature was deprecated in GitLab 11.7.
+//
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/tags.html#update-a-release
-func (s *TagsService) UpdateRelease(pid interface{}, tag string, opt *UpdateReleaseOptions, options ...OptionFunc) (*Release, *Response, error) {
+func (s *TagsService) UpdateRelease(pid interface{}, tag string, opt *UpdateTagReleaseOptions, options ...OptionFunc) (*TagRelease, *Response, error) {
```
svanharmelen: Same as above for this function name.

svanharmelen:
@juranki I noticed that you already addressed a few comments, so just give me a ping when your ready for another review 👍

juranki:
@svanharmelen 
sure, I'll let you know when I've worked through the comments (or have questions)

berniedurfee-ge:
BTW, this is working great now! Thanks for this PR, really useful. :)

https://gitlab.com/electric-it/cago/blob/master/magebuild.go#L187

https://gitlab.com/electric-it/cago/releases

releases.go
juranki: I can't remember the exact thought process for this, probably I was just unsure if Assets will always be populated. But it looks like it will be. I guess pointer can be removed from Commit, too. And of course the author

tags.go
juranki: I don't know. I agree that ReleaseTag sounds better, but maybe TagRelease still describes the purpose more accurately.

For example CreateReleaseTag suggests that it creates a special kind of tag; CreateTagRelease is closer to the truth.

I think TagDescription, ReleaseDescription and ReleaseNote would also describe the purpose quite accurately.

Actually, I give my vote to ReleaseNote, because it's the term that is used in gitlab documentation. https://docs.gitlab.com/ce/api/tags.html#create-a-new-release

svanharmelen: I'm all good with `ReleaseNote` 👍 Go for it 😉

releases.go
juranki: Finally found this again: https://gitlab.com/gitlab-org/gitlab-ce/blob/master/spec/fixtures/api/schemas/release.json

I don't know the codebase of gitlab well enough to determine if this json schema is relevant or not, but in it author and commit are nullable. I can't think of any situation where that would happen, but it tipped me to be liberal with pointers here.

What do you think?

releases.go
svanharmelen: In all other structs we use `struct` instead of `*struct`, even if the field isn't populated. If the field isn't populated the struct would be initialized using it zero values, so it wouldn't break anything. This also prevents the need to check for `nil` pointers.

svanharmelen: Feel free to make a force push, as long as you don't change the code in that push (as I just reviewed and approved the PR 🎉)

Let me know when your done so I can merge the PR. Thanks!

juranki: The rewrite is done.

Double checked that the diff to the original was empty before resetting releases-pr, so the code should be exactly the same.

PR #554, opened 2019-01-29
Fix ISOTime encoding in GET requests

sirlatrom:
Implement the query.Encoder interface on ISOTime, ensuring correct encoding for GET requests, and add `url` struct tags to structs in events.go to ensure correct parameter names.

### DIFF
--- a/events.go
+++ b/events.go
@@ -35,34 +35,34 @@ type EventsService struct {
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/events.html#get-user-contribution-events
 type ContributionEvent struct {
-	Title       string     `json:"title"`
-	ProjectID   int        `json:"project_id"`
-	ActionName  string     `json:"action_name"`
-	TargetID    int        `json:"target_id"`
-	TargetIID   int        `json:"target_iid"`
-	TargetType  string     `json:"target_type"`
-	AuthorID    int        `json:"author_id"`
-	TargetTitle string     `json:"target_title"`
-	CreatedAt   *time.Time `json:"created_at"`
+	Title       string     `url:"title" json:"title"`
+	ProjectID   int        `url:"project_id" json:"project_id"`
+	ActionName  string     `url:"action_name" json:"action_name"`
+	TargetID    int        `url:"target_id" json:"target_id"`
+	TargetIID   int        `url:"target_iid" json:"target_iid"`
+	TargetType  string     `url:"target_type" json:"target_type"`
+	AuthorID    int        `url:"author_id" json:"author_id"`
+	TargetTitle string     `url:"target_title" json:"target_title"`
+	CreatedAt   *time.Time `url:"created_at" json:"created_at"`
 	PushData    struct {
-		CommitCount int    `json:"commit_count"`
-		Action      string `json:"action"`
-		RefType     string `json:"ref_type"`
-		CommitFrom  string `json:"commit_from"`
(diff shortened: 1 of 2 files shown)


events.go
```diff
@@ -35,34 +35,34 @@ type EventsService struct {
...
-	} `json:"author"`
-	AuthorUsername string `json:"author_username"`
+		Name      string `url:"name" json:"name"`
+		Username  string `url:"username" json:"username"`
+		ID        int    `url:"id" json:"id"`
+		State     string `url:"state" json:"state"`
+		AvatarURL string `url:"avatar_url" json:"avatar_url"`
+		WebURL    string `url:"web_url" json:"web_url"`
+	} `url:"author" json:"author"`
+	AuthorUsername string `url:"author_username" json:"author_username"`
```
svanharmelen: This struct is never used in a `GET` call, so why would we want to add the `url` tags here?

events.go
```diff
@@ -71,11 +71,11 @@ type ContributionEvent struct {
...
-	Action     *EventTypeValue       `json:"action,omitempty"`
-	TargetType *EventTargetTypeValue `json:"target_type,omitempty"`
-	Before     *string               `json:"before,omitempty"`
-	After      *string               `json:"after,omitempty"`
-	Sort       *string               `json:"sort,omitempty"`
+	Action     *EventTypeValue       `url:"action,omitempty" json:"action,omitempty"`
+	TargetType *EventTargetTypeValue `url:"target_type,omitempty" json:"target_type,omitempty"`
+	Before     *ISOTime              `url:"before,omitempty" json:"before,omitempty"`
+	After      *ISOTime              `url:"after,omitempty" json:"after,omitempty"`
+	Sort       *string               `url:"sort,omitempty" json:"sort,omitempty"`
```
svanharmelen: Nice catch, totally missed the `url` tags for this one!

gitlab.go
```diff
@@ -123,6 +123,18 @@ func (t *ISOTime) UnmarshalJSON(data []byte) error {
...
+func (t *ISOTime) EncodeValues(key string, v *url.Values) error {
+	if t == nil {
+		return nil
+	}
+	if (time.Time(*t)).IsZero() {
+		return nil
+	}
+	v.Add(key, t.String())
+	return nil
+}
```
svanharmelen: This is indeed the correct solution 👍

gitlab.go
```diff
@@ -123,6 +123,18 @@ func (t *ISOTime) UnmarshalJSON(data []byte) error {
...
 }
 
+// EncodeValues implements the query.Encoder interface
+func (t *ISOTime) EncodeValues(key string, v *url.Values) error {
+	if t == nil {
+		return nil
+	}
+	if (time.Time(*t)).IsZero() {
+		return nil
+	}
```
svanharmelen: I would place these two checks on one line:

```
	if t == nil || (time.Time(*t)).IsZero() {
		return nil
	}
```

events.go
sirlatrom: That's just a result of me being lazy with a regexp replace all in my editor; I'll revert that part in my next push.

gitlab.go
sirlatrom: > I would place these two checks on one line:
> 
> ```
> 	if t == nil || (time.Time(*t)).IsZero() {
> 		return nil
> 	}
> ```

Should be resolved by 02c5e68630d5af6e610c03bded6741997230578e.

PR #556, opened 2019-01-30
Add project clusters API support

matejvelikonja:
Adds support for Project clusters API which is available from GitLab 11.7

https://docs.gitlab.com/ee/api/project_clusters.html

## Usage

Example usage of this endpoints.

```golang
package main

import (
	"fmt"
	"github.com/xanzy/go-gitlab"
	"time"
)

func main() {
	pid := 1234
	git := gitlab.NewClient(nil, "some_token")

	// LIST CLUSTERS

	clts, _, err := git.ProjectCluster.ListClusters(pid)

	if err != nil {
		panic(err)
	}

	fmt.Printf("%v\n", clts)

	// ADD CLUSTER

	name := "New cluster"
	envScope := "*"
	npc := &gitlab.AddClusterOptions{
		Name:             &name,
		EnvironmentScope: &envScope,
		PlatformKubernetes: &gitlab.PlatformKubernetesOptions{
			APIURL:            gitlab.String("https://123.123.123.12/"),
			Token:             gitlab.String(`token`),
			CaCert:            gitlab.String(`-----BEGIN CERTIFICATE-----\r\some certificate\r\n-----END CERTIFICATE-----`),
			Namespace:         gitlab.String("some-test-namespace"),
			AuthorizationType: gitlab.String("rbac"),
		},
	}

	cluster, _, err := git.ProjectCluster.AddCluster(pid, npc)

	if err != nil {
		panic(err)
	}

	// GET CLUSTER

	clt, _, err := git.ProjectCluster.GetCluster(pid, cluster.ID)

	if err != nil {
		panic(err)
	}

	fmt.Printf("%v\n", clt)

	time.Sleep(time.Second * 3)

	// EDIT CLUSTER

	changedName := "Changed K8S"
	optEdit := gitlab.EditClusterOptions{Name: &changedName}

	cltE, _, err := git.ProjectCluster.EditCluster(pid, cluster.ID, &optEdit)

	if err != nil {
		panic(err)
	}

	fmt.Printf("%v\n", cltE)

	time.Sleep(time.Second * 3)

	// DELETE CLUSTER

	_, err = git.ProjectCluster.DeleteCluster(pid, cluster.ID)

	if err != nil {
		panic(err)
	}
}
```

### DIFF
--- a/project_clusters.go
+++ b/project_clusters.go
@@ -53,8 +53,9 @@ func (v ProjectCluster) String() string {
 	return Stringify(v)
 }
 
+// PlatformKubernetes represents a GitLab Project Cluster PlatformKubernetes.
 type PlatformKubernetes struct {
-	ApiUrl            string `json:"api_url,omitempty"`
+	APIURL            string `json:"api_url,omitempty"`
 	Token             string `json:"token,omitempty"`
 	CaCert            string `json:"ca_cert,omitempty"`
 	Namespace         string `json:"namespace,omitempty"`
--- a/project_clusters.go
+++ b/project_clusters.go
@@ -116,10 +116,10 @@ func (s *ProjectClustersService) GetCluster(pid interface{}, cluster int, option
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/project_clusters.html#add-existing-cluster-to-project
 type AddClusterOptions struct {
-	Name               string              `json:"name,omitempty"`
-	Enabled            *bool               `json:"enabled,omitempty"`
-	EnvironmentScope   string              `json:"environment_scope,omitempty"`
-	PlatformKubernetes *PlatformKubernetes `json:"platform_kubernetes_attributes,omitempty"`
+	Name               string              `url:"name,omitempty" json:"name,omitempty"`
+	Enabled            *bool               `url:"enabled,omitempty" json:"enabled,omitempty"`
+	EnvironmentScope   string              `url:"environment_scope,omitempty" json:"environment_scope,omitempty"`
+	PlatformKubernetes *PlatformKubernetes `url:"platform_kubernetes_attributes,omitempty" json:"platform_kubernetes_attributes,omitempty"`
 }
 
 // AddCluster adds an existing cluster to the project.
@@ -152,9 +152,9 @@ func (s *ProjectClustersService) AddCluster(pid interface{}, opt *AddClusterOpti
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/project_clusters.html#edit-project-cluster
 type EditClusterOptions struct {
-	Name               string              `json:"name"`
-	EnvironmentScope   string              `json:"environment_scope,omitempty"`
-	PlatformKubernetes *PlatformKubernetes `json:"platform_kubernetes_attributes,omitempty"`
+	Name               string              `url:"name,omitempty" json:"name,omitempty"`
+	EnvironmentScope   string              `url:"environment_scope,omitempty" json:"environment_scope,omitempty"`
+	PlatformKubernetes *PlatformKubernetes `url:"platform_kubernetes_attributes,omitempty" json:"platform_kubernetes_attributes,omitempty"`
 }
 
 // EditCluster updates an existing project cluster.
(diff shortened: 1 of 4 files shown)


matejvelikonja:
Anything new regarding merging this PR? I would need this to be merged, so I can start working on implementing this into Terraform provider.

svanharmelen:
Thanks for your PR @matejvelikonja! This week is a very busy week for me, so I'll try to make some time to review this PR next week.

project_clusters.go
```diff
@@ -0,0 +1,204 @@
...
+
+// AddClusterOptions represents the available AddCluster() options.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/project_clusters.html#add-existing-cluster-to-project
+type AddClusterOptions struct {
+	Name               string             `json:"name,omitempty"`
+	Enabled            *bool              `json:"enabled,omitempty"`
+	EnvironmentScope   string             `json:"environment_scope,omitempty"`
+	PlatformKubernetes PlatformKubernetes `json:"platform_kubernetes_attributes,omitempty"`
```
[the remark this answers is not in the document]
matejvelikonja: Why would I add `url` tags here, since GitLab accepts this as json and not as query string?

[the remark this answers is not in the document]
svanharmelen: Because some options struct are also used in `GET` calls, in which case these fields have to be added as query params. So to keep things consistent and to prevent the need to check how every individual option struct can be used, we decided to always set these tags for all options structs.

matejvelikonja:
all the findings have been fixed

project_clusters.go
```diff
@@ -0,0 +1,202 @@
...
+// AddClusterOptions represents the available AddCluster() options.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/project_clusters.html#add-existing-cluster-to-project
+type AddClusterOptions struct {
+	Name               string              `url:"name,omitempty" json:"name,omitempty"`
+	Enabled            *bool               `url:"enabled,omitempty" json:"enabled,omitempty"`
+	EnvironmentScope   string              `url:"environment_scope,omitempty" json:"environment_scope,omitempty"`
+	PlatformKubernetes *PlatformKubernetes `url:"platform_kubernetes_attributes,omitempty" json:"platform_kubernetes_attributes,omitempty"`
+}
```
svanharmelen: All the fields in options structs need to be pointers. So `string` should be `*string`.

project_clusters.go
```diff
@@ -0,0 +1,202 @@
...
+
+// EditClusterOptions represents the available EditCluster() options.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/project_clusters.html#edit-project-cluster
+type EditClusterOptions struct {
+	Name               string              `url:"name,omitempty" json:"name,omitempty"`
+	EnvironmentScope   string              `url:"environment_scope,omitempty" json:"environment_scope,omitempty"`
+	PlatformKubernetes *PlatformKubernetes `url:"platform_kubernetes_attributes,omitempty" json:"platform_kubernetes_attributes,omitempty"`
+}
```
svanharmelen: All the fields in options structs need to be pointers.

project_clusters.go
matejvelikonja: fixed

project_clusters.go
matejvelikonja: fixed

matejvelikonja:
In the `PlatformKubernetes` struct, `CaCert`, `Namespace` and `AuthorizationType` are optional.  `Namespace` can be empty string, but `AuthorizationType` is not allowed to be empty string. But it can be `null`. 

So we have 2 options how to handle this. Either set the `omitempty` tag or have pointers to string. Which one would be preferred? If pointers, should all fields be pointers to strings or just the optional ones?

```golang
type PlatformKubernetes struct {
	APIURL            string `json:"api_url"`
	Token             string `json:"token"`
	CaCert            string `json:"ca_cert"`
	Namespace         string `json:"namespace"`
	AuthorizationType string `json:"authorization_type"`
}
```

matejvelikonja:
So I use `*string` in both `Edit` and `Add`?
In `Add` `APIURL` and `Token` are required, but not in `Edit`

svanharmelen:
Yeah, the idea here is that from a users perspective you should not have to check each field to see if it requires a pointer or just a value, so instead we always use a pointer and assume that people using this package check what is required and what is not.

PR #558, opened 2019-02-02
added custom attributes

45cali:
Allows the ability to filter groups, projects and users by custom attributes

### DIFF
--- a/custom_attributes.go
+++ b/custom_attributes.go
@@ -2,6 +2,7 @@ package gitlab
 
 import (
 	"fmt"
+	"strings"
 )
 
 // CustomAttributesService handles communication with the group, project and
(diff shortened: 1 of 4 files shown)


custom_attributes.go
```diff
@@ -169,3 +170,85 @@ func (s *CustomAttributesService) deleteCustomAttribute(resource string, id int,
...
 }
+
+// ListUsersWithCustomAttributes lists users that have the specified custom attributes.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/users.html#for-admins
+func (s *CustomAttributesService) ListUsersWithCustomAttributes(customAttributes []CustomAttribute, options ...OptionFunc) ([]*User, *Response, error) {
+	return s.listUsersWithCustomAttributes(customAttributes, options...)
+}
+
+func (s *CustomAttributesService) listUsersWithCustomAttributes(customAttributes []CustomAttribute, options ...OptionFunc) ([]*User, *Response, error) {
```
svanharmelen: Can you maybe share the idea behind adding both a public and private method for the same action? Especially since they seem to have a 1:1 relation, I wonder if we should take this approach here?

45cali:
Just trying to follow what was previously done. I can pull it out.

svanharmelen:
@45cali I'm a bit busy these days, but just got around to having a closer look at this one.

The difference with the other custom attribute methods, is that they all share the same signature. for for those it makes sense to make it higher order (more or less) methods. But when there is a 1:1 mapping, that doesn't add much value.

That being said, I doubt if this is the right approach for these methods. These methods are actually just variations on the already existing lists (e.g. `ListGroups`, `ListUsers`, `ListProjects`) methods. So I wonder if it wouldn't be better to add a `CustomAttributes` field to their option structs (e.g. `ListGroupsOptions`, `ListUsersOptions`, `ListProjectsOptions`) instead?

45cali:
if i do something like this
```
type ListGroupsOptions struct {
	ListOptions
	AllAvailable         *bool              `url:"all_available,omitempty" json:"all_available,omitempty"`
	MinAccessLevel       *AccessLevelValue  `url:"min_access_level,omitempty" json:"min_access_level,omitempty"`
	OrderBy              *string            `url:"order_by,omitempty" json:"order_by,omitempty"`
	Owned                *bool              `url:"owned,omitempty" json:"owned,omitempty"`
	Search               *string            `url:"search,omitempty" json:"search,omitempty"`
	SkipGroups           []int              `url:"skip_groups,omitempty" json:"skip_groups,omitempty"`
	Sort                 *string            `url:"sort,omitempty" json:"sort,omitempty"`
	Statistics           *bool              `url:"statistics,omitempty" json:"statistics,omitempty"`
	WithCustomAttributes *bool              `url:"with_custom_attributes,omitempty" json:"with_custom_attributes,omitempty"`
	CustomAttrubutes     []CustomAttribute `url:"custom_attributes,omitempty" json:"custom_attributes,omitempty"`
}
```
with 
```
opt := gitlab.ListGroupsOptions{
		CustomAttrubutes: []gitlab.CustomAttribute{
			gitlab.CustomAttribute{
				Key: "key1",
				Value: "val1",
			},
		},
	}
```
its going to come out something like this
`custom_attributes={key1+val1}` not `custom_attributes[key]=value`

45cali:
if i do this instead 
``` 
CustomAttrubutes     gitlab.CustomAttribute `url:"custom_attributes,omitempty" json:"custom_attributes,omitempty"`
``` 

its going to create `custom_attributes[Key]=key1&custom_attributes[Value]=val1` custom_attributes[key1]=value1

Do have any suggestions on a way I can make this work?

45cali:
from gitlabs api docs https://docs.gitlab.com/ce/api/groups.html#list-groups
```
You can filter by custom attributes with:
GET /groups?custom_attributes[key]=value&custom_attributes[other_key]=other_value
```

the value of the key changes so it cant be a key in a struct, but the GET requests in this library use `github.com/google/go-querystring/query` which expect it to be a struct

svanharmelen:
@45cali I think we should make a specific type for `CustomAttributes` and then implement the `query.Encoder` interface on that type (should be placed in `gitlab.go`).

So something like (didn't actually test it) this: https://play.golang.org/p/-YDlqMpQKM-

45cali:
Ok sounds good ill work on that

svanharmelen:
@45cali  please let me know what to do with this PR. Unless any updates in the next week, I'll close the PR. Thanks!

svanharmelen:
Closing this one due to a lack of response...

PR #559, opened 2019-02-07
ADD ReadmeURL to Project struct

mattclegg:
Does anything else need to be updated?


svanharmelen:
LGTM! Thanks @mattclegg!

PR #560, opened 2019-02-08
Split PlatformKubernetesOptions to add and edit struct

matejvelikonja:
Add and edit PlatformKubernetes are different.
It is not possible to edit `authorization_type`.

I have overlooked that in the first PR.

### DIFF
--- a/project_clusters.go
+++ b/project_clusters.go
@@ -115,14 +115,14 @@ func (s *ProjectClustersService) GetCluster(pid interface{}, cluster int, option
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/project_clusters.html#add-existing-cluster-to-project
 type AddClusterOptions struct {
-	Name               *string                    `url:"name,omitempty" json:"name,omitempty"`
-	Enabled            *bool                      `url:"enabled,omitempty" json:"enabled,omitempty"`
-	EnvironmentScope   *string                    `url:"environment_scope,omitempty" json:"environment_scope,omitempty"`
-	PlatformKubernetes *PlatformKubernetesOptions `url:"platform_kubernetes_attributes,omitempty" json:"platform_kubernetes_attributes,omitempty"`
+	Name               *string                       `url:"name,omitempty" json:"name,omitempty"`
+	Enabled            *bool                         `url:"enabled,omitempty" json:"enabled,omitempty"`
+	EnvironmentScope   *string                       `url:"environment_scope,omitempty" json:"environment_scope,omitempty"`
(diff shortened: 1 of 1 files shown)


svanharmelen:
@matejvelikonja wouldn't it be easier to use an inline anonymous struct in this case? 

```
type EditClusterOptions struct {
	Name               *string                    `url:"name,omitempty" json:"name,omitempty"`
	EnvironmentScope   *string                    `url:"environment_scope,omitempty" json:"environment_scope,omitempty"`	
	PlatformKubernetes  struct {
		APIURL    *string `url:"api_url,omitempty" json:"api_url,omitempty"`
		Token     *string `url:"token,omitempty" json:"token,omitempty"`
		CaCert    *string `url:"ca_cert,omitempty" json:"ca_cert,omitempty"`
		Namespace *string `url:"namespace,omitempty" json:"namespace,omitempty"`
	}
}
```

So would something like that work? If it works, I think its a much nicer API to work with.

matejvelikonja:
the initialisation of the `EditClusterOptions` would be really ugly, since you would actually need to copy the whole `PlatformKubernetes` struct.

```golang
e := EditClusterOptions{
		Name:             gitlab.String("some"),
		EnvironmentScope: gitlab.String("some"),
		PlatformKubernetes: struct {
			APIURL    *string `url:"api_url,omitempty" json:"api_url,omitempty"`
			Token     *string `url:"token,omitempty" json:"token,omitempty"`
			CaCert    *string `url:"ca_cert,omitempty" json:"ca_cert,omitempty"`
			Namespace *string `url:"namespace,omitempty" json:"namespace,omitempty"`
		}{
			APIURL:    gitlab.String("some"),
			Token:     gitlab.String("some"),
			CaCert:    gitlab.String("some"),
			Namespace: gitlab.String("some"),
		},
	}
```

And this would not  be future proof, since if you would add a new field to this struct, all the initialisation would needed to be fixed.

svanharmelen:
Of course you could also write that like this:

```go
e := EditClusterOptions{
		Name:             gitlab.String("some"),
		EnvironmentScope: gitlab.String("some"),
}
e.PlatformKubernetes.APIURL =  gitlab.String("some")
e.PlatformKubernetes.Token =  gitlab.String("some")
e.PlatformKubernetes.CaCert =  gitlab.String("some")
e.PlatformKubernetes.Namespace =  gitlab.String("some")
```

Which would prevent writing the out the whole struct and makes it future proof.

svanharmelen:
Still not too happy about how this looks, but I see your point and technically it's a fine solution for the use case... So I'm going to merge this anyway 👍

PR #568, opened 2019-02-20
Add mirror related fields to Project struct

imkh:
The [official documentation](https://docs.gitlab.com/ee/api/projects.html#get-single-project) doesn't seem to be up-to-date so they're not shown in their example but we can do a check ourselves:

```console
$ curl -s -H "Private-Token: <gitlab_token>" "https://gitlab.com/api/v4/projects/gnachman%2Fiterm2" | jq .
{
  ...
  "mirror": true,
  "mirror_user_id": 55954,
  "mirror_trigger_builds": false,
  "only_mirror_protected_branches": null,
  "mirror_overwrites_diverged_branches": null,
  ...
}
```


svanharmelen:
LGTM! Thanks @imkh!

PR #569, opened 2019-02-21
Add GetProjectOptions to support additional attributes

sindrepm:
This PR adds support for specifying additional attributes to the `GetProject` method,  such as to include project statistics in the response.

Docs: https://docs.gitlab.com/ce/api/projects.html#get-single-project

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -320,19 +320,28 @@ func (s *ProjectsService) GetProjectLanguages(pid interface{}, options ...Option
 	return p, resp, err
 }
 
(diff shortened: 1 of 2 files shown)


sindrepm:
> I think you need to update some tests/examples, but looks good otherwise 👍

Tests added/updated. Not sure if an example is necessary just for this small change. Hopefully this can be merged soon.

svanharmelen:
LGTM! Thanks @sindrepm! I'll release a new version as this is a breaking change...

PR #570, opened 2019-02-23
Fix type for ApprovalsBeforeMerge

ZJvandeWeg:
ApprovalsBeforeMerge denotes the number of approvals required before
merging and will never have the type `bool`. This change updates the
type for the field and adds test coverage for the type.


svanharmelen:
Thanks @ZJvandeWeg! LGTM!

Issue #571, opened 2019-02-26
Support for Container Registry API


ivanetchart:
I was taking a look to the code but I think it's not supported yet. 

https://docs.gitlab.com/ee/api/container_registry.html. If you're ok with it, can I submit a PR for this?

Thanks!

svanharmelen:
Thanks for reaching out @ivanetchart! And by all means, PRs are always welcome 👍

ivanetchart:
Cool

Just made it. Looking forward for feedback,

Thanks!

svanharmelen:
Thanks for the PR @ivanetchart!

PR #572, opened 2019-02-27
Add support for Content Registry API

ivanetchart:
Implements #571

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ to add new and/or missing endpoints. Currently the following services are suppor
 - [x] Broadcast Messages
 - [x] Commits
 - [x] Custom Attributes
+- [x] Content Registry
 - [x] Deploy Keys
 - [x] Deployments
 - [x] Environments
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ to add new and/or missing endpoints. Currently the following services are suppor
 - [x] Broadcast Messages
 - [x] Commits
 - [x] Custom Attributes
-- [x] Content Registry
+- [x] Container Registry
 - [x] Deploy Keys
 - [x] Deployments
 - [x] Environments
--- a/gitlab.go
+++ b/gitlab.go
@@ -332,6 +332,7 @@ type Client struct {
 	Projects              *ProjectsService
 	ProtectedBranches     *ProtectedBranchesService
 	ProtectedTags         *ProtectedTagsService
+	Registry              *RegistryService
 	Repositories          *RepositoriesService
 	RepositoryFiles       *RepositoryFilesService
 	Runners               *RunnersService
@@ -474,6 +475,7 @@ func newClient(httpClient *http.Client) *Client {
 	c.Projects = &ProjectsService{client: c}
 	c.ProtectedBranches = &ProtectedBranchesService{client: c}
 	c.ProtectedTags = &ProtectedTagsService{client: c}
+	c.Registry = &RegistryService{client: c}
 	c.Repositories = &RepositoriesService{client: c}
 	c.RepositoryFiles = &RepositoryFilesService{client: c}
 	c.Runners = &RunnersService{client: c}
--- a/gitlab.go
+++ b/gitlab.go
@@ -332,7 +332,7 @@ type Client struct {
 	Projects              *ProjectsService
 	ProtectedBranches     *ProtectedBranchesService
 	ProtectedTags         *ProtectedTagsService
-	Registry              *RegistryService
+	ContainerRegistry     *ContainerRegistryService
 	Repositories          *RepositoriesService
 	RepositoryFiles       *RepositoryFilesService
 	Runners               *RunnersService
@@ -475,7 +475,7 @@ func newClient(httpClient *http.Client) *Client {
 	c.Projects = &ProjectsService{client: c}
 	c.ProtectedBranches = &ProtectedBranchesService{client: c}
 	c.ProtectedTags = &ProtectedTagsService{client: c}
-	c.Registry = &RegistryService{client: c}
+	c.ContainerRegistry = &ContainerRegistryService{client: c}
 	c.Repositories = &RepositoriesService{client: c}
 	c.RepositoryFiles = &RepositoryFilesService{client: c}
 	c.Runners = &RunnersService{client: c}
--- a/registry.go
+++ b/registry.go
@@ -6,11 +6,11 @@ import (
 	"time"
 )
 
-// RegistryService handles communication with the container registry related methods
-// of the GitLab API.
+// ContainerRegistryService handles communication with the container registry
+// related methods of the GitLab API.
 //
 // GitLab API docs: https://docs.gitlab.com/ee/api/container_registry.html
-type RegistryService struct {
+type ContainerRegistryService struct {
 	client *Client
 }
 
@@ -22,7 +22,7 @@ type RegistryRepository struct {
 	Name      string     `json:"name"`
 	Path      string     `json:"path"`
 	Location  string     `json:"location"`
-	CreatedAt *time.Time `json:"created_at,omitempty"`
+	CreatedAt *time.Time `json:"created_at"`
 }
 
 func (s RegistryRepository) String() string {
@@ -33,19 +33,6 @@ func (s RegistryRepository) String() string {
 //
 // GitLab API docs: https://docs.gitlab.com/ee/api/container_registry.html
 type RegistryRepositoryTag struct {
-	Name     string `json:"name"`
-	Path     string `json:"path"`
-	Location string `json:"location"`
-}
-
-func (s RegistryRepositoryTag) String() string {
-	return Stringify(s)
-}
-
-// RegistryRepositoryTagDetail represents a GitLab registry tag detail.
-//
-// GitLab API docs: https://docs.gitlab.com/ee/api/container_registry.html
-type RegistryRepositoryTagDetail struct {
 	Name          string     `json:"name"`
 	Path          string     `json:"path"`
 	Location      string     `json:"location"`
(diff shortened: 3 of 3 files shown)


README.md
```diff
@@ -30,6 +30,7 @@ to add new and/or missing endpoints. Currently the following services are suppor
 - [x] Broadcast Messages
 - [x] Commits
 - [x] Custom Attributes
+- [x] Content Registry
```
svanharmelen: This should be `Container Registry`

gitlab.go
```diff
@@ -332,6 +332,7 @@ type Client struct {
 	Projects              *ProjectsService
 	ProtectedBranches     *ProtectedBranchesService
 	ProtectedTags         *ProtectedTagsService
+	Registry              *RegistryService
```
svanharmelen: This should be `ContainerRegistry`

registry.go
```diff
@@ -0,0 +1,220 @@
...
+	"fmt"
+	"net/url"
+	"time"
+)
+
+// RegistryService handles communication with the container registry related methods
+// of the GitLab API.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/container_registry.html
+type RegistryService struct {
```
svanharmelen: This should be `ContainerRegistryService`

All these name suggestions are just to keep the naming conventions in this package consistent.

registry.go
```diff
@@ -0,0 +1,220 @@
+package gitlab
+
+import (
+	"fmt"
+	"net/url"
+	"time"
+)
+
+// RegistryService handles communication with the container registry related methods
```
svanharmelen: Just as a FYI... I try to keep comments below 80chars and LoC below 100chars. So this comment would then become:

```
// RegistryService handles communication with the container registry related
// methods of the Gitlab API.
```

A valid exception for comments is when a URL is longer the 80chars and is already on its own line and a valid exception for LoC is function signatures, as I always let them be on a single line. Thanks!

registry.go
```diff
@@ -0,0 +1,220 @@
...
+
+// RegistryRepository represents a GitLab content registry repository.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/container_registry.html
+type RegistryRepository struct {
+	ID        int        `json:"id"`
+	Name      string     `json:"name"`
+	Path      string     `json:"path"`
+	Location  string     `json:"location"`
+	CreatedAt *time.Time `json:"created_at,omitempty"`
```
svanharmelen: `omitempty` can be omitted here as this struct is only used to unmarshal the response.

registry.go
```diff
@@ -0,0 +1,220 @@
...
+type RegistryRepositoryTagDetail struct {
+	Name          string     `json:"name"`
+	Path          string     `json:"path"`
+	Location      string     `json:"location"`
+	Revision      string     `json:"revision"`
+	ShortRevision string     `json:"short_revision"`
+	Digest        string     `json:"digest"`
+	CreatedAt     *time.Time `json:"created_at,omitempty"`
+	TotalSize     int        `json:"total_size"`
+}
```
svanharmelen: I wonder if we should use two different structs for `RegistryRepositoryTag` and `RegistryRepositoryTagDetail`? I would suggest to just use on struct instead. What do you think?

registry.go
```diff
@@ -0,0 +1,220 @@
...
+
+// BulkDeleteRegistryRepositoryTagsOptions represents the available BulkDeleteRegistryRepositoryTags() options.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/container_registry.html#delete-repository-tags-in-bulk
+type BulkDeleteRegistryRepositoryTagsOptions struct {
+	NameRegexp *string `url:"name_regex,omitempty" json:"name_regex,omitempty"`
+	KeepN      *int    `url:"keep_n,omitempty" json:"keep_n,omitempty"`
+	OlderThan  *string `url:"older_than,omitempty" json:"older_than,omitempty"`
+}
```
svanharmelen: Please define these option structs just above the first function that requires them.

registry.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	return repos, resp, err
+}
+
+// DeleteRegistryRepository deletes a repository in a registry.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/container_registry.html#delete-registry-repository
+func (s *RegistryService) DeleteRegistryRepository(pid interface{}, repositoryID int, options ...OptionFunc) (*Response, error) {
```
svanharmelen: To keep the naming consistent, `repositoryID` should be `repository`. Goes for all other methods as well of course.

registry.go
```diff
@@ -0,0 +1,220 @@
...
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, err
+	}
+
```
svanharmelen: Please remove this blank line (again for consistency 😉) Goes for other methods as well of course.

registry.go
```diff
@@ -0,0 +1,220 @@
...
+	u := fmt.Sprintf("projects/%s/registry/repositories/%d/tags", url.QueryEscape(project), repositoryID)
+
+	req, err := s.client.NewRequest("GET", u, opt, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var repos []*RegistryRepositoryTag
```
svanharmelen: `repos` should probably be `tags` here?

registry.go
```diff
@@ -0,0 +1,220 @@
...
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/container_registry.html#get-details-of-a-repository-tag
+func (s *RegistryService) GetRegistryRepositoryTagDetail(pid interface{}, repositoryID int, tagName string, options ...OptionFunc) (*RegistryRepositoryTagDetail, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	u := fmt.Sprintf("projects/%s/registry/repositories/%d/tags/%s", url.QueryEscape(project), repositoryID, tagName)
```
svanharmelen:
```suggestion
	u := fmt.Sprintf("projects/%s/registry/repositories/%d/tags/%s",
		url.QueryEscape(project),
		repositoryID,
		tagName,
	)
```

registry.go
```diff
@@ -0,0 +1,220 @@
...
+
+	req, err := s.client.NewRequest("GET", u, nil, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var tagDetail *RegistryRepositoryTagDetail
```
svanharmelen:
```suggestion
	td := new(RegistryRepositoryTagDetail)
```

svanharmelen: Again, just to keep things consistent ^^

registry.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	return s.client.Do(req, nil)
+}
+
+// BulkDeleteRegistryRepositoryTags deletes repository tags in bulk based on given criteria.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/container_registry.html#delete-repository-tags-in-bulk
+func (s *RegistryService) BulkDeleteRegistryRepositoryTags(pid interface{}, repositoryID int, opt *BulkDeleteRegistryRepositoryTagsOptions, options ...OptionFunc) (*Response, error) {
```
svanharmelen: I would rename this method to `DeleteRegistryRepositoryTags` to make it just a bit shorter.

registry.go
```diff
@@ -0,0 +1,220 @@
...
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/container_registry.html#delete-a-repository-tag
+func (s *RegistryService) DeleteRegistryRepositoryTag(pid interface{}, repositoryID int, tagName string, options ...OptionFunc) (*Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, err
+	}
+
+	u := fmt.Sprintf("projects/%s/registry/repositories/%d/tags/%s", url.QueryEscape(project), repositoryID, tagName)
```
svanharmelen:
```suggestion
	u := fmt.Sprintf("projects/%s/registry/repositories/%d/tags/%s",
		url.QueryEscape(project),
		repositoryID,
		tagName,
	)
```

registry.go
```diff
@@ -0,0 +1,220 @@
...
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/container_registry.html#list-repository-tags
+type ListRegistryRepositoryTagsOptions ListOptions
+
+// BulkDeleteRegistryRepositoryTagsOptions represents the available BulkDeleteRegistryRepositoryTags() options.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/container_registry.html#delete-repository-tags-in-bulk
+type BulkDeleteRegistryRepositoryTagsOptions struct {
```
svanharmelen: I would call this `DeleteRegistryRepositoryTagsOptions` to make it just a bit shorter. I understand that the methods names would them be `DeleteRegistryRepositoryTag` vs `DeleteRegistryRepositoryTags` which is very similar, but still clear enough I think.

README.md
ivanetchart: Totally, I don't know why I used `Content` 😕

registry.go
ivanetchart: I think it makes sense. `RegistryRepositoryTag` is a subset of `
RegistryRepositoryTagDetail`. 👍

ivanetchart:
@svanharmelen I think most of the changes should be covered. I need to test a little bit more the bulk delete operation as it's done asynchronously and obviously need a bigger amount of images to make sure it's actually working.

svanharmelen:
LGTM! I'm going to merge this in as is now. If you need to make tweaks to the bulk methods I prefer those in a new smaller PR 😉 Thanks!

PR #573, opened 2019-03-01
added support for group badges API

tcchrist:


group_badges.go
```diff
@@ -0,0 +1,206 @@
...
+	Kind string `json:"kind"`
+}
+
+// GroupBadgesService handles communication with the group badges
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/group_badges.html
+type GroupBadgesService struct {
+	client *Client
+}
```
svanharmelen: We generally place the service type on top of the file as the first declaration.

group_badges.go
```diff
@@ -0,0 +1,206 @@
...
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/group_badges.html
+type GroupBadge struct {
+	ID               int    `json:"id"`
+	LinkURL          string `json:"link_url"`
+	ImageURL         string `json:"image_url"`
+	RenderedLinkURL  string `json:"rendered_link_url"`
+	RenderedImageURL string `json:"rendered_image_url"`
+	// Kind represents a project badge kind. Can be empty, when used PreviewProjectBadge().
+	Kind string `json:"kind"`
```
svanharmelen: I think it would be nice if we create a specific type for this that represents the kind. Something like we did here: https://github.com/xanzy/go-gitlab/blob/master/gitlab.go#L79-L90

And then maybe call it `BadgeKind` or something? That would make this field a little more strongly types. Through I now notice that we never get to set this field, but we only read it from the API. So that maybe makes this a bit less relevant. What do you think?

tcchrist:
> I have two very minor suggestions, but this looks really nice @tcchrist!

thank you, very kind

group_badges.go
tcchrist: agreed. I've moved it to the top.

group_badges.go
tcchrist: Makes sense. I've introduced a type for badge kind.

svanharmelen:
LGTM! Thanks @tcchrist!

Issue #563, opened 2019-02-13
Support for group labels API


odin243:
I was examining the labels code, and it doesn't look like there's any support for the labels API within a group (rather than a project): https://docs.gitlab.com/ce/api/group_labels.html

Unless I missed it somewhere, is this something you'd be open to a PR for?

svanharmelen:
Sure, I'm always open to PRs 😉

kirecek:
Any update here? I can work on this if there is no assignee :)

svanharmelen:
There's no assignee, so by all means 👍

svanharmelen:
This issue is fixed by PR #579

PR #579, opened 2019-03-10
Add support for GroupLabels

kirecek:
Hello,

this PR adds support for group labels. Since the API is basically same as project labels it's more or less copy/paste of existing labels code.

Issue: https://github.com/xanzy/go-gitlab/issues/563

### DIFF
--- a/group_labels.go
+++ b/group_labels.go
@@ -60,7 +60,8 @@ type CreateGroupLabelOptions CreateLabelOptions
 // CreateGroupLabel creates a new label for given group with given name and
 // color.
 //
-// GitLab API docs: https://docs.gitlab.com/ce/api/group_labels.html#create-a-new-group-label
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/group_labels.html#create-a-new-group-label
 func (s *GroupLabelsService) CreateGroupLabel(gid interface{}, opt *CreateGroupLabelOptions, options ...OptionFunc) (*GroupLabel, *Response, error) {
 	group, err := parseID(gid)
 	if err != nil {
--- a/group_labels.go
+++ b/group_labels.go
@@ -109,7 +109,8 @@ func (s *GroupLabelsService) DeleteGroupLabel(gid interface{}, opt *DeleteGroupL
 
 // UpdateGroupLabelOptions represents the available UpdateGroupLabel() options.
 //
-// GitLab API docs: https://docs.gitlab.com/ce/api/group_labels.html#update-a-group-label
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/group_labels.html#update-a-group-label
 type UpdateGroupLabelOptions UpdateLabelOptions
 
 // UpdateGroupLabel updates an existing label with new name or now color. At least
--- a/group_labels.go
+++ b/group_labels.go
@@ -116,7 +116,8 @@ type UpdateGroupLabelOptions UpdateLabelOptions
 // UpdateGroupLabel updates an existing label with new name or now color. At least
 // one parameter is required, to update the label.
 //
-// GitLab API docs: https://docs.gitlab.com/ce/api/labels.html#edit-an-existing-label
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/group_labels.html#update-a-group-label
 func (s *GroupLabelsService) UpdateGroupLabel(gid interface{}, opt *UpdateGroupLabelOptions, options ...OptionFunc) (*GroupLabel, *Response, error) {
 	group, err := parseID(gid)
 	if err != nil {
--- a/group_labels.go
+++ b/group_labels.go
@@ -85,7 +85,8 @@ func (s *GroupLabelsService) CreateGroupLabel(gid interface{}, opt *CreateGroupL
 
 // DeleteGroupLabelOptions represents the available DeleteGroupLabel() options.
 //
-// GitLab API docs: https://docs.gitlab.com/ce/api/group_labels.html#delete-a-group-label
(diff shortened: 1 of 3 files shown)


group_labels.go
```diff
@@ -0,0 +1,191 @@
...
+
+// ListGroupLabelsOptions represents the available ListGroupLabels() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/labels.html#list-labels
+type ListGroupLabelsOptions ListOptions
+
+// ListGroupLabels gets all labels for given group.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/group_labels.html#list-group-labels
+func (s *GroupLabelsService) ListGroupLabels(gid interface{}, opt *ListGroupLabelsOptions, options ...OptionFunc) ([]*GroupLabel, *Response, error) {
```
svanharmelen: As this method is already on the `GroupLabelsService`, it's probably enough to call this `ListLabels`? I guess that goes for all method names, we can probably lose the `Group` part. What do you think?

group_labels.go
```diff
@@ -0,0 +1,191 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return l, resp, err
+}
+
+// CreateGroupLabelOptions represents the available CreateGroupLabel() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/group_labels.html#create-a-new-group-label
```
svanharmelen:
```suggestion
// GitLab API docs:
// https://docs.gitlab.com/ce/api/group_labels.html#create-a-new-group-label
```

group_labels.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+// ListGroupLabelsOptions represents the available ListGroupLabels() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/labels.html#list-labels
+type ListGroupLabelsOptions ListOptions
+
+// ListGroupLabels gets all labels for given group.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/group_labels.html#list-group-labels
```
svanharmelen:
```suggestion
// GitLab API docs:
// https://docs.gitlab.com/ce/api/group_labels.html#list-group-labels
```

group_labels.go
```diff
@@ -0,0 +1,191 @@
...
+
+// CreateGroupLabelOptions represents the available CreateGroupLabel() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/group_labels.html#create-a-new-group-label
+type CreateGroupLabelOptions CreateLabelOptions
+
+// CreateGroupLabel creates a new label for given group with given name and
+// color.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/group_labels.html#create-a-new-group-label
```
svanharmelen:
```suggestion
// GitLab API docs:
// https://docs.gitlab.com/ce/api/group_labels.html#create-a-new-group-label
```

group_labels.go
```diff
@@ -0,0 +1,191 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return l, resp, err
+}
+
+// DeleteGroupLabelOptions represents the available DeleteGroupLabel() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/group_labels.html#delete-a-group-label
```
svanharmelen:
```suggestion
// GitLab API docs:
// https://docs.gitlab.com/ce/api/group_labels.html#delete-a-group-label
```

group_labels.go
```diff
@@ -0,0 +1,191 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return s.client.Do(req, nil)
+}
+
+// UpdateGroupLabelOptions represents the available UpdateGroupLabel() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/group_labels.html#update-a-group-label
```
svanharmelen:
```suggestion
// GitLab API docs:
// https://docs.gitlab.com/ce/api/group_labels.html#update-a-group-label
```

group_labels.go
```diff
@@ -0,0 +1,191 @@
...
+type UpdateGroupLabelOptions UpdateLabelOptions
+
+// UpdateGroupLabel updates an existing label with new name or now color. At least
+// one parameter is required, to update the label.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/labels.html#edit-an-existing-label
```
svanharmelen:
```suggestion
// GitLab API docs:
// https://docs.gitlab.com/ce/api/group_labels.html#update-a-group-label
```

group_labels.go
kirecek: :+1: I agree. I just tried to be consistent with naming style in other `group_*.go` files. On it.

svanharmelen: Oh, I see... Well in that case then I prefer to keep things consistent. So let's not change it then!

kirecek: ok

svanharmelen: I see that you updated all the comments, so then I think we can just merge the PR now right? Anything else you wanted to add/change?

kirecek: I think it's good to go from my POV ... Maybe just one thing to consider is to add `Project` prefix to previous labels code since everything except labels uses prefixes to separate group and project resources.

kirecek: It would be breaking change tho...

svanharmelen: Agree, yet it does currently match how it names and documented in the Gitlab docs, so I think that's fine for now. Thanks again for your work!

PR #580, opened 2019-03-12
Add omitempty tag to approval_before_merge field of CreateProjectOptions

kchugalinskiy:
Without this change in case when there are no any field changes EditProject method fails with internal server error. It is crucial to be fixed to enhance terraform project resource with this argument.


svanharmelen:
Good catch @kchugalinskiy! LGTM!

PR #583, opened 2019-03-18
Add the MergeRequestsCount to the Issue

ZJvandeWeg:
GitLab would expose this information, but this field wasn't parsed in
the struct. A test case was updated to have an execution test of the
parsing.

GitLab docs: https://docs.gitlab.com/ee/api/issues.html#list-project-issues


svanharmelen:
LGTM! Thanks @ZJvandeWeg!

PR #586, opened 2019-03-21
Add tests for group and project access requests

davidkuridza:
This PR adds tests for group and project access requests API.

There's a lot of repeating code here, let me know what would be the best way of simplifying it please. I've looked at the existing tests and there doesn't seem to be a unified style.

### DIFF
--- a/access_requests_test.go
+++ b/access_requests_test.go
@@ -1,0 +1,401 @@
+package gitlab
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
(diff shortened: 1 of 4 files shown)


svanharmelen:
Thanks for your PR @davidkuridza! I know the tests are not consistent as they didn't originally existed when I initially created this package. And as I'm not working with Gitlab anymore during my daytime job, I didn't get around in adding tests for everything or to define a default approach.

So whenever a PR comes in that adds tests I'm supportive of that and (if they look OK) tent to just merge them in without too much feedback. Maybe (when time permits) I should review the test framework and see if I can add tests to a few services which can then be used as reference.

But in all honesty I will not have time for that anytime soon 😏

PR #590, opened 2019-03-22
add text_label & is_project_label to label type

itscaro:
these fields are added in 11.9 which was released today.

### DIFF
--- a/labels.go
+++ b/labels.go
@@ -37,12 +37,14 @@ type Label struct {
 	ID                     int    `json:"id"`
 	Name                   string `json:"name"`
 	Color                  string `json:"color"`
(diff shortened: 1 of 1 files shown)


labels.go
```diff
@@ -37,12 +37,14 @@ type Label struct {
 	ID                     int    `json:"id"`
 	Name                   string `json:"name"`
 	Color                  string `json:"color"`
+	TextColor              string `json:"text_color"` //11.9+
```
svanharmelen: I appreciate the comments, but I prefer we omit them 🙂

itscaro: Done :-)

svanharmelen:
Thanks! LGTM!

PR #592, opened 2019-03-26
add external wiki service

itscaro:


services.go
```diff
@@ -647,3 +647,91 @@ func (s *ServicesService) DeleteMicrosoftTeamsService(pid interface{}, options .
...
+// SetExternalWikiServiceOptions represents the available SetExternalWikiService()
+// options.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/services.html#createedit-external-wiki-service
+type SetExternalWikiServiceOptions struct {
+	ExternalWikiURL *string `url:"external_wiki_url,omitempty" json:"external_wiki_url,omitempty"`
+}
+
+// SetExternalWikiService sets External Wiki service for a project
```
svanharmelen:
```suggestion
// SetExternalWikiService sets External Wiki service for a project.
```

services.go
```diff
@@ -647,3 +647,91 @@ func (s *ServicesService) DeleteMicrosoftTeamsService(pid interface{}, options .
...
+	if err != nil {
+		return nil, err
+	}
+	u := fmt.Sprintf("projects/%s/services/external-wiki", url.QueryEscape(project))
+
+	req, err := s.client.NewRequest("PUT", u, opt, options)
+	if err != nil {
+		return nil, err
+	}
+	return s.client.Do(req, nil)
```
svanharmelen:
```suggestion

	return s.client.Do(req, nil)
```

services.go
```diff
@@ -647,3 +647,91 @@ func (s *ServicesService) DeleteMicrosoftTeamsService(pid interface{}, options .
...
+// https://docs.gitlab.com/ce/api/services.html#createedit-external-wiki-service
+type SetExternalWikiServiceOptions struct {
+	ExternalWikiURL *string `url:"external_wiki_url,omitempty" json:"external_wiki_url,omitempty"`
+}
+
+// SetExternalWikiService sets External Wiki service for a project
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/services.html#createedit-external-wiki-service
+func (s *ServicesService) SetExternalWikiService(pid interface{}, opt *SetExternalWikiServiceOptions, options ...OptionFunc) (*Response, error) {
```
svanharmelen: Just to double check, doesn't the call to set an external wiki also return the newly set external wiki? If so, we should return it here as well.

itscaro: Yes it does, but to be in sync with other services, I propose to leave the return as it is. In another PR we can do the change for all functions in services.go. But this will be a breaking change.

svanharmelen: Fair enough 👍🏻 I like to keep things consistent as well 😉

PR #593, opened 2019-03-26
 according to the documentation, UpdatedByID field in MergeEvent should be a slice of int

itscaro:


event_types.go
```diff
@@ -545,17 +545,14 @@ type MergeEvent struct {
 			Previous []Label `json:"previous"`
 			Current  []Label `json:"current"`
 		} `json:"labels"`
-		UpdatedByID struct {
-			Previous int `json:"previous"`
-			Current  int `json:"current"`
-		} `json:"updated_by_id"`
+		UpdatedByID []*uint `json:"updated_by_id"`
```
svanharmelen: This should not be a pointer and I prefer to use an `int` instead of a `uint`, even through a `uint` is a very valid option here.

itscaro: it should be pointer because, according to the documentation null can be set here :-)
I will change to `*int`

svanharmelen: `null` will be unmarshalled to `0` which is fine as that indicates no ID.

Having these as pointers makes this slice much more cumbersome to use as you would always have to check how many items are in the slice and dereference the value.

svanharmelen: And when using `[]*int` it would unmarshal `null` to `(*int)(nil)` so the slice would still have 2 values, but one would be a nil-pointer which would be pretty nasty to work with.

itscaro: thanks for your explanation 👍

svanharmelen:
LGTM! Thanks @itscaro!

PR #595, opened 2019-03-28
feat(projects): Add mirror values to CreateProjectOptions

ChristianAlexander:


projects.go
```diff
@@ -445,6 +445,8 @@ type CreateProjectOptions struct {
 	PrintingMergeRequestLinkEnabled           *bool             `url:"printing_merge_request_link_enabled,omitempty" json:"printing_merge_request_link_enabled,omitempty"`
 	CIConfigPath                              *string           `url:"ci_config_path,omitempty" json:"ci_config_path,omitempty"`
 	ApprovalsBeforeMerge                      *int              `url:"approvals_before_merge,omitempty" json:"approvals_before_merge,omitempty"`
+	Mirror                                    *bool             `url:"mirror,omitempty" json:"mirror,omitempty"`
+	MirrorTriggerBuilds                       *bool             `url:"mirror_trigger_builds,omitempty" json:"mirror_trigger_builds,omitempty"`
```
svanharmelen: Maybe we should add `initialize_with_readme` as well to make the list complete again?

ChristianAlexander: Sure.

I also noticed that there is a new `mirror_user_id` parameter which is only present on the project edit route. It might be valuable to split out the `EditProjectOptions` type from the Create type alias that it currently is implemented as.

svanharmelen: If they have diverged so they are not (almost) the same anymore, that is probably a good thing to do 👍

svanharmelen:
LGTM! Thanks @ChristianAlexander!

PR #596, opened 2019-03-28
allow to export/import types in YAML format correctly

itscaro:
add yaml along with json & url annotations

### DIFF
--- a/access_requests.go
+++ b/access_requests.go
@@ -11,13 +11,13 @@ import (
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/access_requests.html
 type AccessRequest struct {
-	ID          int              `json:"id"`
-	Username    string           `json:"username"`
-	Name        string           `json:"name"`
-	State       string           `json:"state"`
-	CreatedAt   *time.Time       `json:"created_at"`
-	RequestedAt *time.Time       `json:"requested_at"`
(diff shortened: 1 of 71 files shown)


svanharmelen:
Hi @itscaro! I'm not sure if I would like to add `yaml` tags. Not because it's `yaml`, but because it isn't needed for or by the API client. So maybe you can share the use case or reason for adding this?

itscaro:
I use yaml to store settings of projects in order to maintain the project's settings via API. I could do it in JSON but YAML allow me to leave comments and YAML or TOML are more widely used to store configurations :-)

If the annotation could be included in the library, I will not need to maintain a copy of types coming from go-gitlab with yaml annotations for my use case.

svanharmelen:
I want to think a little longer about this one, but I lean towards not merging this.

Even through I understand your use case, but if we would merge this PR then we would also need to merge any other PR that adds any (or all) of these well known tags: https://github.com/golang/go/wiki/Well-known-struct-tags

But all those tags are not needed by this package and so feels like we should not add them at all.

svanharmelen:
I'm sorry @itscaro, but I'm not going to merge this one for now. I understand your use case, but don't want to end up adding all possible tags to this package as they aren't needed for the Gitlab API.

Hope you can understand my reasoning! Thanks!

PR #597, opened 2019-03-28
Update ListGroupProjectsOptions to match API

micabot:
Instead of using ListProjectsOptions, ListGroupProjectsOptions now
matches the specific options defined for ListGroupProjects() in the API.


svanharmelen:
LGTM! Thanks @micabot!

PR #598, opened 2019-03-30
Add support for creating issue boards

tanordheim:
Hi!

I noticed support for creating issue boards was lacking in the boards API, so I added that. I don't know if you're OK with adding non-CE functionality to the project but I didn't see any particular notice about it in the README so I'm giving it a shot :)

Thanks for a great project!

### DIFF
--- a/boards.go
+++ b/boards.go
@@ -57,6 +57,90 @@ func (b BoardList) String() string {
 	return Stringify(b)
 }
 
+// CreateIssueBoardOptions represents the available CreateIssueBoard() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/boards.html#create-a-board-starter
+type CreateIssueBoardOptions struct {
+	Name *string `url:"name" json:"name"`
+}
+
+// CreateIssueBoard creates a new issue board.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/boards.html#create-a-board-starter
+func (s *IssueBoardsService) CreateIssueBoard(pid interface{}, opt *CreateIssueBoardOptions, options ...OptionFunc) (*IssueBoard, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


svanharmelen:
Hi @tanordheim! I'm fine with adding these endpoints, but then I think we should also add the update and delete board endpoints to make it possible to manage boards instead of only create them. What do you think?

tanordheim:
That makes sense; I've added this and updated my branch now.

boards.go
```diff
@@ -57,6 +57,90 @@ func (b BoardList) String() string {
...
+}
+
+// UpdateIssueBoardOptions represents the available UpdateIssueBoard() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/boards.html#update-a-board-starter
+type UpdateIssueBoardOptions struct {
+	Name        *string `url:"name" json:"name"`
+	AssigneeID  *int    `url:"assignee_id" json:"assignee_id"`
+	MilestoneID *int    `url:"milestone_id" json:"milestone_id"`
+	Labels      *string `url:"labels" json:"labels"`
```
tanordheim: According to the [API docs](https://docs.gitlab.com/ee/api/boards.html#update-a-board-starter), this endpoint takes in a "comma separated list of label names", so I just made this field a string as well. I couldn't find any other examples of how this is handled previously as I believe most API endpoints that operate on lists of things actually take in an array as well.

Do you want me to keep this as-is or make the model take in an array and transform it internally in the `UpdateIssueBoard` func?

svanharmelen: You can use the `Labels` type for that: https://github.com/xanzy/go-gitlab/blob/eca1beae81589cf0cfc27500316185b52a44eb07/issues.go#L102

tanordheim: Ah, sweet - adjusted the code now. Thanks!

svanharmelen:
LGTM! Thanks @tanordheim!

PR #599, opened 2019-03-30
Update GetCommit entity fields

ZJvandeWeg:
I wanted to leverage this API call, but it was missing data that I'd
like to use. I've added the fields and a test case.

For the test case, I've removed the timestamps as it made the code a bit
ugly. But the rest is fully covered.

### DIFF
--- a/commits.go
+++ b/commits.go
@@ -48,6 +48,8 @@ type Commit struct {
 	ParentIDs      []string         `json:"parent_ids"`
 	Stats          *CommitStats     `json:"stats"`
 	Status         *BuildStateValue `json:"status"`
+	LastPipeline   *PipelineInfo    `json:"last_pipeline"`
+	ProjectID      int              `json:"project_id"`
 }
 
 // CommitStats represents the number of added and deleted files in a commit.
--- a/commits_test.go
+++ b/commits_test.go
@@ -5,7 +5,48 @@ import (
 	"net/http"
 	"reflect"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
 )
+
+func TestGetCommit(t *testing.T) {
+	mux, server, client := setup()
+	defer teardown(server)
+
+	mux.HandleFunc("/api/v4/projects/1/repository/commits/b0b3a907f41409829b307a28b82fdbd552ee5a27", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		mustWriteHTTPResponse(t, w, "testdata/get_commit.json")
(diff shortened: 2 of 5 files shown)


commits_test.go
```diff
@@ -2,11 +2,60 @@ package gitlab
...
 )
 
+func TestGetCommit(t *testing.T) {
+	mux, server, client := setup()
+	defer teardown(server)
+
+	mux.HandleFunc("/api/v4/projects/1/repository/commits/b0b3a907f41409829b307a28b82fdbd552ee5a27", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+
+		f, err := os.Open("testdata/get_commit.json")
```
ZJvandeWeg: I think this cleans up the tests, but this would be the first test case using it. Let me know what you think!

pipelines.go
```diff
@@ -82,6 +82,14 @@ type PipelineList []struct {
 	WebURL string `json:"web_url"`
 }
 
+type SimplePipeline struct {
```
ZJvandeWeg: The simple pipeline could be used in the ListPipeline struct too, as these are the same. That would break the API, and I don't know what the policy is around that pre 1.0 for this project.

commits_test.go
svanharmelen: Maybe we should create a small helper (in `gitlab_test.go`) that takes a file path and returns the content of the file? But I like the approach and all the big JSON strings in the tests don't make it very nice to read 👍

commits.go
```diff
@@ -48,6 +48,8 @@ type Commit struct {
 	ParentIDs      []string         `json:"parent_ids"`
 	Stats          *CommitStats     `json:"stats"`
 	Status         *BuildStateValue `json:"status"`
+	LastPipeline   *SimplePipeline  `json:"last_pipeline,omitempty"`
+	ProjectID      int              `json:"project_id,omitempty"`
```
svanharmelen: These fields do not need the `omitempty` tag.

pipelines.go
svanharmelen: I don't like the name `SimplePipeline` as to me it indicates a type of pipeline. Could we use `PipelineInfo` or `PipelineMeta` of something alike? Updating`ListPipeline` to use this struct makes sense and should be fine.

ZJvandeWeg: Picked PipelineInfo, as I felt it was more descriptive.

svanharmelen:
LGTM! Thanks @ZJvandeWeg!

PR #602, opened 2019-04-06
Fixed the GetIssuesClosedOnMerge path (previously resulted in a 404 error)

beyera:
Quick fix for a bug I encountered:
image

Thanks to this community for the awesome package! I've been having fun with it!


svanharmelen:
Thanks for the PR @beyera! Good catch!

PR #603, opened 2019-04-10
fix-bug/dot-in-groupName

Tonyqu123:


hoozecn:
refer: [python gitlab implementation](https://github.com/python-gitlab/python-gitlab/blob/a3a771310de16be7bba041c962223f7bda9aa4d6/gitlab/utils.py#L50)

svanharmelen:
@Tonyqu123 I just opened PR #613 in order to fix this bug using a different approach. It would be great if you can verify that the PR works correctly and fixes your issue.

PR #606, opened 2019-04-10
Adds ResolveOutdatedDiffDiscussions to Project struct

ThatGerber:


projects.go
```diff
@@ -87,6 +87,7 @@ type Project struct {
 	MirrorTriggerBuilds                       bool              `json:"mirror_trigger_builds"`
 	OnlyMirrorProtectedBranches               bool              `json:"only_mirror_protected_branches"`
 	MirrorOverwritesDivergedBranches          bool              `json:"mirror_overwrites_diverged_branches"`
+	ResolveOutdatedDiffDiscussions            bool              `json:"resolve_outdated_diff_discussions"`
```
svanharmelen: I know this doesn't hold up in a lot of places, but I'm trying to keep the order of fields in our structs inline with how they come back from the API. So if we follow that for this one as well, this should be between `Snippets` and `ContainerRegistryEnabled`.

So if you could move it there that would be great! Thanks!

ThatGerber: Done

PR #607, opened 2019-04-13
fix DeletePipelineSchedule always returning an error due to EOF

thojkooi:
Gitlab API returns 204 on DELETE endpoints. Trying to read out the pipeline schedule from the delete response will always fail. This changes the DeletePipelineSchedule function to behave similar as the DeletePipelineTrigger function.


svanharmelen:
LGTM! Thanks @thojkooi!

PR #609, opened 2019-04-16
Fixed multiple scope jobs selection

biohedge:
At the moment if we provide multiple scopes in listprojectjobs request, the library returns jobs with the last provided state in the scope list.
This is because the scope parameter should have braces in its name as shown in the example: https://docs.gitlab.com/ee/api/jobs.html#list-pipeline-jobs


svanharmelen:
Nice catch @biohedge! Thanks!

PR #614, opened 2019-04-25
Add domain to project cluster

matejvelikonja:
This PR adds a new field to project cluster `domain`.

The feature has been introduced in GitLab 1.10.

### DIFF
--- a/project_clusters.go
+++ b/project_clusters.go
@@ -37,6 +37,7 @@ type ProjectClustersService struct {
 type ProjectCluster struct {
 	ID                 int                 `json:"id"`
 	Name               string              `json:"name"`
+	Domain             string              `json:"domain"`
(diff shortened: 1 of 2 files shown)


svanharmelen:
Thanks @matejvelikonja! LGTM!

matejvelikonja:
@svanharmelen any chance of making also a new release? 🙏

svanharmelen:
I just merged another PR (#613) that I would like to get verified by the community (as in, does everything still work as expected). If you can take the latest code in master for a quick spin to confirm its all good, than I'll cut a release right after your done. Is that an option?

matejvelikonja:
Sure, I will try it out with Terraform provider, since I need to implement domain also there.

svanharmelen:
Cool, let me know if all is good and then I'll add a new tag. Thanks!

PR #615, opened 2019-04-30
updatedById is not an array of int but an object like other fields

manland:
Related [gitlab issue #61187](https://gitlab.com/gitlab-org/gitlab-ce/issues/61187) and [merge request !27897](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/27897) 

Wait and see...


manland:
Merge request on gitlab is merged!

PR #619, opened 2019-05-07
Added field which shows if the branch is default

biohedge:
Added "default" field, which shows if the branch is default in its project


svanharmelen:
LGTM! Thanks @biohedge!

PR #623, opened 2019-05-16
Update Jira service: change JiraTransitionID from int to string

jorcau:
Hi,

This is an update to https://github.com/xanzy/go-gitlab/pull/547.

This makes *JiraTransitionID* a string while still supporting previous Gitlab version (<11.9) when it was an int (cf. https://gitlab.com/gitlab-org/gitlab-ce/blob/master/CHANGELOG.md#L443).

Best,

### DIFF
--- a/services.go
+++ b/services.go
@@ -17,7 +17,11 @@
 package gitlab
 
 import (
+	"encoding/json"
 	"fmt"
+	"net/url"
+	"reflect"
+	"strconv"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


jorcau:
Hi @svanharmelen,

I see you made the change here in https://github.com/xanzy/go-gitlab/commit/73dbd5788854fe0d78ff4b1fbeb41c74f708a5ac.

Should I rebase this PR to keep the UnmarshalJSON or should we drop it?

svanharmelen:
Thanks @jorcau! LGTM!

PR #624, opened 2019-05-17
add "managed" flag for toggling managed clusters during create

bdimcheff:
The gitlab api now has a ["managed"](https://docs.gitlab.com/ee/api/project_clusters.html#add-existing-cluster-to-project) field to disable cluster management features when adding a kubernetes cluster.  Disabling this allows you to specify an unprivileged, namespace-scoped token to gitlab, to avoid granting it cluster admin. This implements that field. Default is true if you don't supply it.

Existing testing doesn't seem to cover stuff at the property-level, so I didn't implement it. Since it defaults to true if the field isn't sent, this _shouldn't_ affect anything else.


svanharmelen:
Thanks @bdimcheff!

PR #625, opened 2019-05-18
Add Fork Project Options

pcarranza:
With the current implementation, it's impossible to set a namespace, a name, or a path.

These attributes are supported by the API, but there's no way to set them, and the implementation of the OptionFun forces everything to be a JSON payload, thus it doesn't work.

With this change, we can set the attributes using a `ForkProjectOptions` symmetrical to the rest of the methods all around.

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -557,18 +557,27 @@ func (s *ProjectsService) EditProject(pid interface{}, opt *EditProjectOptions, 
 	return p, resp, err
 }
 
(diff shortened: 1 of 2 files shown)


pcarranza:
@svanharmelen Apologies, clearly my editor is misconfigured.

I used the opportunity to add a simple test

svanharmelen:
LGTM! Thanks @pcarranza!

PR #627, opened 2019-05-23
Added custom writer to Archive

brandon-welsch:
Added the ability to set a custom Writer in options passed to Archive of RepositoriesService.

Very useful when you do not want to store the contents of the repository in the memory and use the Reader directly.

Storing large repositories in memory is not very convenient.

### DIFF
--- a/repositories.go
+++ b/repositories.go
@@ -19,6 +19,7 @@ package gitlab
 import (
 	"bytes"
 	"fmt"
+	"io"
 )
 
 // RepositoriesService handles communication with the repositories related
@@ -138,6 +139,7 @@ func (s *RepositoriesService) RawBlobContent(pid interface{}, sha string, option
 type ArchiveOptions struct {
 	Format *string `url:"-" json:"-"`
 	SHA    *string `url:"sha,omitempty" json:"sha,omitempty"`
+	Writer *io.Writer `url:"-" json:"-"`
 }
 
 // Archive gets an archive of the repository.
@@ -162,7 +164,13 @@ func (s *RepositoriesService) Archive(pid interface{}, opt *ArchiveOptions, opti
 	}
 
 	var b bytes.Buffer
-	resp, err := s.client.Do(req, &b)
+	var resp *Response
+	if opt != nil && opt.Writer != nil {
+		resp, err = s.client.Do(req, opt.Writer)
+	} else {
+		resp, err = s.client.Do(req, &b)
+	}
+
 	if err != nil {
 		return nil, resp, err
 	}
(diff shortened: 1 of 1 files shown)


repositories.go
```diff
@@ -138,6 +139,7 @@ func (s *RepositoriesService) RawBlobContent(pid interface{}, sha string, option
 type ArchiveOptions struct {
 	Format *string `url:"-" json:"-"`
 	SHA    *string `url:"sha,omitempty" json:"sha,omitempty"`
+	Writer *io.Writer `url:"-" json:"-"`
```
svanharmelen: It looks like your code is not formatted using `gofmt`

repositories.go
```diff
@@ -162,7 +164,13 @@ func (s *RepositoriesService) Archive(pid interface{}, opt *ArchiveOptions, opti
 	}
 
 	var b bytes.Buffer
-	resp, err := s.client.Do(req, &b)
+	var resp *Response
+	if opt != nil && opt.Writer != nil {
+		resp, err = s.client.Do(req, opt.Writer)
+	} else {
+		resp, err = s.client.Do(req, &b)
+	}
```
svanharmelen: I get what you are trying to do, but I don't think this is the best way to implement this. Let me think a little about this one...

repositories.go
brandon-welsch: Sorry, my IDE was not configured correctly with this project, it's fixed

repositories.go
brandon-welsch: I want to stream the content of the repository directly from a micro-service to another micro-service, via HTTP with `io.Copy`.

To do this, I propose this way of doing this, avoiding any breakage of the current method.

Is that clear for you?

svanharmelen: It's perfectly clear to me, but I don't like this particular solution. Let me add an alternative solution for this instead. I'll push it as additional commit to this PR so we can discuss if it will work for you.

svanharmelen:
I prefer this solution over hacking it in the existing method. It keeps the intentions clear and doesn't return an unused slice of bytes.

brandon-welsch:
Good to me, one thing must be added.

Set the metadata of the content file in headers of Writer :
- Content-Disposition
- Content-Length
- Content-Transfer-Encoding
- Content-Type

I suggest replacing `io.Writer` with `http.ResponseWriter`.

The file we get has an incorrect extension, I get `archive.tar` instead of `archive.tar.gz`.

brandon-welsch:
Finally, I just set the following headers :
- Content-Disposition : "attachment; filename="archive.tar.gz"
- Content-Transfer-Encoding : "binary"
- Content-Type : "application/octet-stream"

I left the `Content-Length` header, I was not useful for me.

svanharmelen:
How does this matter for an `io.Writer`? I don't think switching this to a `http.ResponseWriter` makes sense as an `io.Writer` is more flexible in general.

PR #629, opened 2019-05-26
Use pointer to struct types in MergeRequest

doronbehar:
This allows comparing these fields of a `mergeRequest` returned by `GetMergeRequest` against `nil`.

I've encountered the need for these changes after I tried all solutions [here](https://stackoverflow.com/questions/28447297/how-to-check-for-an-empty-struct) and I realized that it would be better to patch the `MergeRequest` type to use pointers :)

I'm a Go beginner so feel free to comment!

### DIFF
--- a/merge_requests.go
+++ b/merge_requests.go
@@ -30,6 +30,24 @@ type MergeRequestsService struct {
 	timeStats *timeStatsService
 }
 
+// MergeRequestAssignee represents the assignee of a merge request.
+type MergeRequestAssignee struct {
+	ID        int        `json:"id"`
+	Username  string     `json:"username"`
+	Name      string     `json:"name"`
+	State     string     `json:"state"`
+	CreatedAt *time.Time `json:"created_at"`
+}
+
+// MergeRequestAuthor represents the author of the merge request.
+type MergeRequestAuthor struct {
+	ID        int        `json:"id"`
+	Username  string     `json:"username"`
+	Name      string     `json:"name"`
+	State     string     `json:"state"`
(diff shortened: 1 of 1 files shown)


svanharmelen:
@doronbehar thanks for the PR and I fully understand what you are trying to achieve, but this change could cause nasty issue for people already using this package.

I'm not afraid to make backwards incompatible changes, but I rather don't make this one. I assume that people currently test if they received a value by doing something like this:

```
if MergeRequest.MergedBy.ID != 0 {
```

When we now make `MergedBy` a pointer and the value would not be in the response, the above test would panic. And since that is not something you can detect during compile time, this could cause nasty production issues for users of this package.

If this was new code, this would absolutely be fine. But adding this now seems like a bad idea.

doronbehar:
Oh right, I'll use that in the meantime. It's rather confusing though that some fields are pointers and some are not.. I need to actually look into the source code to know that.. I hope you'll reconsider merging this although this is backwards incompatible..

BTW, thanks so much for creating this library! It's so well organized and easy to use, wonderful! You have no idea how easy you've made my life. I sure will give you credit when my app will be finished :)

svanharmelen:
Thanks for the nice feedback @doronbehar!

For now I'm going to close this PR. If we ever want to add this kind of change, I suggest we do it throughout the whole package and make a new major release. By doing everything at once, we'll keep the package consistent and hopefully prevent a lot of issues for users of the package.

doronbehar:
Oh right, please let me know if you need someone to do all the boring work, I'd be happy to help.

PR #631, opened 2019-05-30
Enable configuring endpoint for NewClient

doronbehar:
I know this is a breaking change, just like my previous PR :|. I'm
sending this PR anyway in case you may be interested never the less. In
any case, it should be possible only to write a new function that will
enable this functionality.


svanharmelen:
What’s the problem with creating a client and then calling `SetBaseURL` yourself on that client?

doronbehar:
Right, fool me :neutral_face: I haven't noticed `SetBaseURL` is public. Thanks!

PR #633, opened 2019-06-05
Add AvatarURL and WebURL to MergeRequest resource

sue445:
There are already available in v11.11.2
https://docs.gitlab.com/ee/api/merge_requests.html#get-single-mr


svanharmelen:
Thanks @sue445! LGTM!

PR #637, opened 2019-06-14
[Merge requests API] Add missing field: merge_error

mtojek:
Hi,

I've noticed that Merge requests API client doesn't provide "merge error". The field is available once the end client calls MR API with the _ include_rebase_in_progress_ option. The field _rebase_in_progress_ has been already defined correctly.

This PR has been prepared according to the instruction available here:
https://docs.gitlab.com/ee/api/merge_requests.html#rebase-a-merge-request

/cc @svanharmelen 

Thanks,
Marcin

### DIFF
--- a/merge_requests.go
+++ b/merge_requests.go
@@ -71,6 +71,7 @@ type MergeRequest struct {
 	Milestone                 *Milestone `json:"milestone"`
 	MergeWhenPipelineSucceeds bool       `json:"merge_when_pipeline_succeeds"`
 	MergeStatus               string     `json:"merge_status"`
+	MergeError                string     `json:"merge_error"`
 	MergedBy                  struct {
 		ID        int        `json:"id"`
 		Username  string     `json:"username"`


merge_requests.go
```diff
@@ -123,9 +123,10 @@ type MergeRequest struct {
 		HeadSha  string `json:"head_sha"`
 		StartSha string `json:"start_sha"`
 	} `json:"diff_refs"`
-	DivergedCommitsCount int  `json:"diverged_commits_count"`
-	RebaseInProgress     bool `json:"rebase_in_progress"`
-	ApprovalsBeforeMerge int  `json:"approvals_before_merge"`
+	DivergedCommitsCount int     `json:"diverged_commits_count"`
+	RebaseInProgress     bool    `json:"rebase_in_progress"`
+	ApprovalsBeforeMerge int     `json:"approvals_before_merge"`
+	MergeError           *string `json:"merge_error"`
```
svanharmelen: I think this field should be blow `MergeStatus` in order to keep the same ordering as the docs (which I try to use as guideline). And I lso think this shouldn't be a pointer to a string, but just a string...

mtojek: Thanks for the comment. I've adjusted the PR.

svanharmelen:
Thanks @mtojek!

PR #638, opened 2019-06-17
add masked variable support for gitlab projects and groups

Jnoack331:
I am a Go newbie so please double check my code :)

### DIFF
--- a/group_variables.go
+++ b/group_variables.go
@@ -38,6 +38,7 @@ type GroupVariable struct {
 	Key       string `json:"key"`
 	Value     string `json:"value"`
 	Protected bool   `json:"protected"`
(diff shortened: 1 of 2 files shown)


group_variables.go
```diff
@@ -38,6 +38,7 @@ type GroupVariable struct {
 	Key       string `json:"key"`
 	Value     string `json:"value"`
 	Protected bool   `json:"protected"`
+	Masked    bool   `json:"masked"`
```
svanharmelen: I cannot seem to find this field in the docs for group variables. Are you sure its part of the response data?

Jnoack331: your right! it does not look like its available for group vars. I will change the code accordingly.

PR #640, opened 2019-06-21
tell json encoder polling_interval_multiplier is encoded as string type

mLeicknam:
Gitlab return `polling_interval_multiplier` as `string` type since it's a `decimal` type and it's handle in rails as string (due to JSON imprecision).

see example:
https://gist.github.com/ahuret/b3174d18f2d379b6ca9d9a9f7f0eb837

Golang JSON decoder accept as annotation to indicates float64 is encoded as string.


svanharmelen:
Nice catch/fix 👍 Thanks @mLeicknam!

PR #641, opened 2019-06-22
implemneted get environment

ilansh:
I am using this package quite a bit, would be happy to give back some.  
This change was tested with real gitlab api, made sure it didn't break ListEnvironments either.

### DIFF
--- a/environments.go
+++ b/environments.go
@@ -32,10 +32,11 @@ type EnvironmentsService struct {
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/environments.html
 type Environment struct {
-	ID          int    `json:"id"`
-	Name        string `json:"name"`
-	Slug        string `json:"slug"`
-	ExternalURL string `json:"external_url"`
+	ID             int        `json:"id"`
+	Name           string     `json:"name"`
+	Slug           string     `json:"slug"`
+	ExternalURL    string     `json:"external_url"`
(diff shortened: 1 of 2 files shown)


environments.go
```diff
@@ -32,10 +32,11 @@ type EnvironmentsService struct {
...
 type Environment struct {
-	ID          int    `json:"id"`
-	Name        string `json:"name"`
-	Slug        string `json:"slug"`
-	ExternalURL string `json:"external_url"`
+	ID             int        `json:"id"`
+	Name           string     `json:"name"`
+	Slug           string     `json:"slug"`
+	ExternalURL    string     `json:"external_url"`
+	LastDeployment Deployment `json:"last_deployment,omitempty"`
```
svanharmelen: This should be a pointer (`*Deployment`) and the `omitempty` tag is not needed here.

environments.go
```diff
@@ -74,6 +75,31 @@ func (s *EnvironmentsService) ListEnvironments(pid interface{}, opts *ListEnviro
 	return envs, resp, err
 }
 
+// GetEnvironment gets a specific environment from a project
```
svanharmelen:
```suggestion
// GetEnvironment gets a specific environment from a project.
```

ilansh:
Done, should I squash the commits?

svanharmelen:
All good 👍Thanks!

Issue #600, opened 2019-04-01
Add support for resource label events 


ravinwind:
Gitlab EE supports using Resource label events API to keep track about who, when, and which label was added or removed to an issuable.

It's a great API.

Proposal would be to add support to go-gitlab for this: 
https://docs.gitlab.com/ee/api/resource_label_events.html

Guys can check this PR to see some more details.
https://github.com/python-gitlab/python-gitlab/pull/625

mseelye:
We also needed this, so I just temporarily added it to my project with some ugly adhoc code.

QQ: If I were to fork this to help out with this, would this belong in the projects service code, or somewhere else like events? It seems like at least the projects resource label events would be in projects as projects is in the url.

svanharmelen:
In general I'm trying to keep the services defined in the package in line with the documentation: https://docs.gitlab.com/ee/api/README.html

So in this case I think it should probably be best to add another service called `ResourceLabels` or maybe `ResourceLabelEvents`.

mseelye:
@ravinwind @svanharmelen Here is what I have for this. I don't have epics on my account, and we don't use MR labels, so I've only actually tested the project-issue functions.

Please take a peek and comment, if everyone is happy I can put up a MR:

https://github.com/mseelye/go-gitlab/pull/1

mseelye:
I think this is closed. @ravinwind let us know if it works for you too.

svanharmelen:
Yes, this should be fixed by #642 now 👍🏻

ravinwind:
> I think this is closed. @ravinwind let us know if it works for you too.

@mseelye It does work. Thanks for the help. Highly appreciated~

PR #642, opened 2019-06-24
Add resource label events

mseelye:
This is in response to issue #600 

Decided just to create the MR with hopes of getting this in sooner (so I don't have to run on my fork). Note that I've tested the the project-issue functions, however, since I do not have epics and do not use merge labels those have not been thoroughly tested but follow the published API specs. 

I've tried to follow what was done already for some of the other services. For instance the ResourceLabelEvent structure is based on the struct I saw in the Event service.

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -341,6 +341,7 @@ type Client struct {
 	Releases              *ReleasesService
 	Repositories          *RepositoriesService
 	RepositoryFiles       *RepositoryFilesService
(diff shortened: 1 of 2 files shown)


svanharmelen:
I renamed a few methods so they are more inline with the docs and a bit shorter (they were getting pretty long). But looks good otherwise, so thanks for the PR @mseelye!

mseelye:
Awesome thank you!

PR #646, opened 2019-06-27
Add support for getting a commit's GPG signature.

chrsow:
Based on Gitlab's commit reference, https://docs.gitlab.com/ce/api/commits.html#get-gpg-signature-of-a-commit.


commits.go
```diff
@@ -546,3 +547,47 @@ func (s *CommitsService) RevertCommit(pid interface{}, sha string, opt *RevertCo
 
 	return c, resp, err
 }
+
+// GPGSubKey represents a GPG subkey ID.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/commits.html#get-gpg-signature-of-a-commit
+type GPGSubKey struct {
+	KeyID sql.NullInt64
```
svanharmelen: This is not something I would like to add. Can you explain why you used `sql.NullInt64`?

commits.go
```diff
@@ -546,3 +547,47 @@ func (s *CommitsService) RevertCommit(pid interface{}, sha string, opt *RevertCo
...
+// GPGSignature represents a Gitlab commit's GPG Signature.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/commits.html#get-gpg-signature-of-a-commit
+type GPGSignature struct {
+	KeyID              int        `json:"gpg_key_id"`
+	KeyPrimaryKeyID    string     `json:"gpg_key_primary_keyid"`
+	KeyUserName        string     `json:"gpg_key_user_name"`
+	KeyUserEmail       string     `json:"gpg_key_user_email"`
+	VerificationStatus string     `json:"verification_status"`
+	KeySubkeyID        *GPGSubKey `json:"gpg_key_subkey_id"`
```
svanharmelen: Do you have an example output that shows how this field look in the raw response?

commits.go
```diff
@@ -546,3 +547,47 @@ func (s *CommitsService) RevertCommit(pid interface{}, sha string, opt *RevertCo
...
+
+	url := fmt.Sprintf("projects/%s/repository/commits/%s/signature", pathEscape(project), sha)
+	req, err := s.client.NewRequest("GET", url, nil, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var signature *GPGSignature
+	resp, err := s.client.Do(req, &signature)
+
```
svanharmelen: For consistency with the other code in this package, this blank line should be omitted.

commits.go
```diff
@@ -546,3 +547,47 @@ func (s *CommitsService) RevertCommit(pid interface{}, sha string, opt *RevertCo
...
+		return nil, nil, err
+	}
+
+	url := fmt.Sprintf("projects/%s/repository/commits/%s/signature", pathEscape(project), sha)
+	req, err := s.client.NewRequest("GET", url, nil, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var signature *GPGSignature
```
svanharmelen: To keep things consistent, I prefer something like this:

```suggestion
	sig := new(GPGSignature)
```

commits.go
```diff
@@ -546,3 +547,47 @@ func (s *CommitsService) RevertCommit(pid interface{}, sha string, opt *RevertCo
...
+// GetGPGSiganature gets a GPG signature of a commit.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/commits.html#get-gpg-signature-of-a-commit
+func (s *CommitsService) GetGPGSiganature(pid interface{}, sha string, options ...OptionFunc) (*GPGSignature, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	url := fmt.Sprintf("projects/%s/repository/commits/%s/signature", pathEscape(project), sha)
```
svanharmelen: In all other function `url` is called just `u` and is placed directly below the error check. So lines 579 and 578 (the blank line) should be swapped.

chrsow:
Hi @svanharmelen, just updated the changes requested.

1. the reason for importing `sql.NullInt64` is because to handle nullable integer type for `KeySubkeyID`.

2. the output of the function is something like below which I added in `testdata/get_signature.json`, I also added the test case in `commits_test.go`. 
```
{
    "gpg_key_id": 7977,
    "gpg_key_primary_keyid": "627C5F589F467F17",
    "gpg_key_user_name": "Dmitriy Zaporozhets",
    "gpg_key_user_email": "dmitriy.zaporozhets@gmail.com",
    "verification_status": "verified",
    "gpg_key_subkey_id": null
}
```

svanharmelen:
I made some changes as I didn't want to use `sql.NullInt64` for this use case and to make the PR consistent with the rest of the code in this package. Looks good to me now 👍

PR #651, opened 2019-07-12
ListTodo allows for ListOptions

ZJvandeWeg:
The GitLab API Docs are missing these options, but they are supported as
the GitLab specs show[1], and can be verified for a signed in user[2].

Including them here breaks the interface when users used assignments
without Field names.

[1]: https://gitlab.com/gitlab-org/gitlab-ce/blob/4325986f25b7523acac49cfaefb7149208ae8553/spec/requests/api/todos_spec.rb#L39
[2]: https://gitlab.com/api/v4/todos?per_page=1

### DIFF
--- a/todos_test.go
+++ b/todos_test.go
@@ -1,10 +1,10 @@
 package gitlab
 
 import (
-	"fmt"
 	"net/http"
-	"reflect"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 )
 
 func TestListTodos(t *testing.T) {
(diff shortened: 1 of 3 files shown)


todos_test.go
```diff
@@ -13,21 +13,16 @@ func TestListTodos(t *testing.T) {
 
 	mux.HandleFunc("/api/v4/todos", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
-		fmt.Fprint(w, `[{"id":1,"state": "pending","target":{"id":1,"approvals_before_merge":2}},{"id":2,"state":"pending","target":{"id":2,"approvals_before_merge":null}}]`)
+		mustWriteHTTPResponse(t, w, "testdata/list_todos.json")
 	})
 
-	opts := &ListTodosOptions{}
+	opts := &ListTodosOptions{ListOptions: ListOptions{PerPage: 2}}
```
ZJvandeWeg: I'm aware this doesn't change the fixture, nor it's output. But it does test the interface.

svanharmelen:
LGTM! Thanks @ZJvandeWeg!

PR #671, opened 2019-08-27
SSL cert checking disabled

iskanderbs:
Fix for error: x509: certificate signed by unknown authority

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/xanzy/go-gitlab
+module github.com/iskanderbs/go-gitlab
 
 require (
 	github.com/google/go-querystring v1.0.0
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/xanzy/go-gitlab
+module github.com/iskanderbs/go-gitlab
```
svanharmelen: Uuh... Please don't change this 😏

iskanderbs: Sorry for my stupid solution, thank you for your reply)

svanharmelen: No worries, all good 👍

colynn:
Defined new http.Client as below:
```
gitlabToken := token
transCfg := &http.Transport{
  TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, // ignore expired SSL certificates
}
httpClient := &http.Client{
      Transport: transCfg,
}

client := go-gitlab.NewClient(httpClient, gitlabToken)
```

PR #675, opened 2019-09-05
add Time func to get time.Time pointer as String/Int/Bool

ahuret:


svanharmelen:
LGTM! Thanks @ahuret!

PR #677, opened 2019-09-14
Project-level MR approvals: Fix Change allowed approvers HTTP-method

leominov:
Ref: https://docs.gitlab.com/ee/api/merge_request_approvals.html#change-allowed-approvers


svanharmelen:
Good catch! Thanks @leominov!

PR #667, opened 2019-08-21
Add test for merge-request parsing

alexkalderimis:
This adds new tests for merge-requests

It incidentally fixes a bug where pipeline users were not being read.

PR #678, opened 2019-09-17
Add sugessted approvers for MergeRequestApprovals request, AssigneeID…

h0ckeypl4yer:
Add sugessted approvers for MergeRequestApprovals request, AssigneeIDs for Merge Requests


merge_request_approvals.go
```diff
@@ -33,6 +33,7 @@ type MergeRequestApprovals struct {
 	ApprovedBy           []*MergeRequestApproverUser  `json:"approved_by"`
 	Approvers            []*MergeRequestApproverUser  `json:"approvers"`
 	ApproverGroups       []*MergeRequestApproverGroup `json:"approver_groups"`
+	SuggestedApprovers   []*ApproverUser              `json:"suggested_approvers"`
```
svanharmelen: I cannot find this in the docs, can you point me to a location that shows this is indeed a valid field?

h0ckeypl4yer: There is no information about this parameter in docs. But it works. Version GitLab Enterprise Edition 12.2.5-ee

merge_requests.go
```diff
@@ -476,6 +476,7 @@ type CreateMergeRequestOptions struct {
 	TargetBranch       *string `url:"target_branch,omitempty" json:"target_branch,omitempty"`
 	Labels             Labels  `url:"labels,comma,omitempty" json:"labels,omitempty"`
 	AssigneeID         *int    `url:"assignee_id,omitempty" json:"assignee_id,omitempty"`
+	AssigneeIDs        []*int  `url:"assignee_ids,omitempty" json:"assignee_ids,omitempty"`
```
svanharmelen:
```suggestion
	AssigneeIDs        []int  `url:"assignee_ids,omitempty" json:"assignee_ids,omitempty"`
```

merge_requests.go
```diff
@@ -518,6 +519,7 @@ type UpdateMergeRequestOptions struct {
 	Description        *string `url:"description,omitempty" json:"description,omitempty"`
 	TargetBranch       *string `url:"target_branch,omitempty" json:"target_branch,omitempty"`
 	AssigneeID         *int    `url:"assignee_id,omitempty" json:"assignee_id,omitempty"`
+	AssigneeIDs        []*int  `url:"assignee_ids,omitempty" json:"assignee_ids,omitempty"`
```
svanharmelen:
```suggestion
	AssigneeIDs        []int  `url:"assignee_ids,omitempty" json:"assignee_ids,omitempty"`
```

merge_request_approvals.go
```diff
@@ -60,7 +61,7 @@ type MergeRequestApproverGroup struct {
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/merge_request_approvals.html#project-level-mr-approvals
 type MergeRequestApproverUser struct {
-	User struct {
+	ApproverUser struct {
```
svanharmelen: This will break code that uses `MergeRequestApproverUser` as now the expected (`User`) field name changed (to `ApprovedUser`). So please revert this change.

merge_request_approvals.go
svanharmelen: I need to see some reference to this in order to be able to judge if this is the best way to add this field. A link to the definition in the Gitlab source code would also be fine.

itscaro: Cf. https://gitlab.com/gitlab-org/gitlab/blob/master/ee/lib/ee/api/entities.rb#L425-428

svanharmelen:
Thanks for updating the PR. Can you please rebase the PR and as a last verification (sorry for all the questions, but I don't have an instance to test against myself at the moment) provide a (censored) API response that contains this field? Thanks!!

h0ckeypl4yer:
> Thanks for updating the PR. Can you please rebase the PR and as a last verification (sorry for all the questions, but I don't have an instance to test against myself at the moment) provide a (censored) API response that contains this field?
> 
> _EDIT:_ I just merged PR #667 which adds a `BasicUser` struct which could (should) be used here as well. So could you please update hat as well? Thanks!

I updated the merge request and added a screenshot from PostMan. I hope this is enough.
Снимок экрана 2019-09-24 в 12 33 40

PR #681, opened 2019-09-20
add some fields to MergeCommentEvent

itscaro:


itscaro:
> @itscaro I cannot find these fields in the docs. Can you point me to something that shows they exist and should be added? Thanks!

I tried to find in documentation but found nothing, I saw these fields in my tests with gitlab CE, these fields are useful, to retry a pipeline, or to reply to a discussion :)

I will try to search in source code.

image

svanharmelen:
Cool, thanks for the links and thanks for your PR @itscaro! Much appreciated 👍🏻

PR #691, opened 2019-09-26
GetCommit should return err if sha is an empty string

WingT:
Change-Id: If54baf0aded1a3b4ec7dccc2e206d510a8fee833

### DIFF
--- a/commits.go
+++ b/commits.go
@@ -176,6 +176,9 @@ func (s *CommitsService) GetCommit(pid interface{}, sha string, options ...Optio
 	if err != nil {
 		return nil, nil, err
 	}
+	if sha == "" {
+		return nil, nil, fmt.Errorf("sha mustn't be empty")
+	}
 	u := fmt.Sprintf("projects/%s/repository/commits/%s", pathEscape(project), url.PathEscape(sha))
 
 	req, err := s.client.NewRequest("GET", u, nil, options)
--- a/commits.go
+++ b/commits.go
@@ -177,7 +177,7 @@ func (s *CommitsService) GetCommit(pid interface{}, sha string, options ...Optio
 		return nil, nil, err
 	}
 	if sha == "" {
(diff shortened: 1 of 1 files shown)


commits.go
```diff
@@ -176,6 +176,9 @@ func (s *CommitsService) GetCommit(pid interface{}, sha string, options ...Optio
 	if err != nil {
 		return nil, nil, err
 	}
+	if sha == "" {
+		return nil, nil, fmt.Errorf("sha mustn't be empty")
+	}
```
svanharmelen: I appreciate what you are trying to do here, but (at least until now) this package doesn't do any input validation and instead leverages the Gitlab API validation logic. So it just forwards the request and returns any errors that the API encounters while validating. This seems more robust as there are a lot of things that are tested and validated by the Gitlab API endpoints which also change over time.

So I'm inclined to close this PR instead of merge it, unless you have a very good reason why this specific check needs to be in here?

WingT: Actually in this case when sha is an empty string, the sdk will make a request to another [api](https://docs.gitlab.com/ee/api/commits.html#list-repository-commits) which gets a list of commits, quite confusing

WingT: I think we can merge this pr

commits.go
```diff
@@ -176,6 +176,9 @@ func (s *CommitsService) GetCommit(pid interface{}, sha string, options ...Optio
 	if err != nil {
 		return nil, nil, err
 	}
+	if sha == "" {
+		return nil, nil, fmt.Errorf("sha mustn't be empty")
```
svanharmelen:
```suggestion
		return nil, nil, fmt.Errorf("SHA must be a non-empty string")
```

WingT: updated

svanharmelen:
Thanks @WingT! LGTM!

PR #695, opened 2019-09-30
Added groups to RunnerDetails

bboehmke:
Runners that are assigned to a group will provide the information as a groups list inside the runner details.
Sadly this is currently not documented (or I have not found it).

But here the response from the API:
image


svanharmelen:
Thanks @bboehmke! LGTM!

Issue #696, opened 2019-09-30
Latest changes to runners.go fails


calculiTestUser:
Packages json and strings need to be imported in runners.go.
Runners.go line 87

PR #697, opened 2019-09-30
fix imports

temamagic:
#696 fix


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/xanzy/go-gitlab
+module github.com/temamagic/go-gitlab
```
svanharmelen: You should not change this path in a PR.

temamagic: Hm, strange, I'll change this after PR... anyway, thx!

PR #698, opened 2019-10-01
Add template values to CreateProjectOptions

virtualroot:
It enables the creation of a new project from a built-in or custom template.
The feature is available in GitLab Premium, GitLab.com Silver, and higher tiers.

Reference: https://docs.gitlab.com/ce/api/projects.html#create-project


svanharmelen:
Thanks @virtualroot! LGTM!

PR #701, opened 2019-10-04
Add order and sort by to runner jobs options

BimoZX:
This allows runner jobs sort order to be changed from descending to ascending.

References:
- [List runner jobs](https://docs.gitlab.com/ee/api/runners.html#list-runners-jobs)


svanharmelen:
Thanks @BimoZX! LGTM!

PR #710, opened 2019-10-09
Allow all labels to be removed from MR  

### LINKED ISSUE #709: Unable to remove all labels from a merge request
The UpdateMergeRequestOptions struct has omitempty specified in the json and url tags for the Labels field. This causes the method to error when trying to remove all labels from a MergeRequest as the api expects an empty string to be set on the field in this case.

I'm happy to submit a PR to fix it. I think the quickest fix is to change the Labels type to a pointer so it forces it to be marshaled in the cases when the array is specified but empty, although I'm open to suggestions.

ConorNevin:
Fixes #709


ConorNevin:
Thanks @svanharmelen - looking at your PR it only fixes this issue for issues not merge requests.

svanharmelen:
Oeps... Updated just now by adding a commit 👍

svanharmelen:
I had to revert this solution as it caused issues. See #715 for more details...

Issue #632, opened 2019-05-31
Most gitlab APIs support urlencoded strings and this is broken when used in go-gitlab


pschwartz:
The liberal use of the `pathEscape` function leads to issues when using the gitlab api to query things that are path based.

`group/subgroup/project` is valid

`curl --header "PRIVATE-TOKEN: ${token}" https://gitlab.com/api/v4/projects/group%2Fsubgroup%2Fproject` can be used to query it.

But using `gitlab.Client.Projects.GetProjects("group/subgroup/project")` or  `gitlab.Client.Projects.GetProjects("group%2Fsubgroup%2Fproject")` fails to find the searched for group.

These should work in many APIs, a few examples are groups, projects, repository_files

pschwartz:
Here is an example right from the gitlab api docs.

https://docs.gitlab.com/ee/api/repository_files.html#get-file-from-repository

`file_path (required) - Url encoded full path to new file. Ex. lib%2Fclass%2Erb`

pschwartz:
A hack around the issue is to enforce encoding prior to calling the go-gitlab functions. Has anyone else seen this behavior?

svanharmelen:
@pschwartz if I look at the [repository_files](https://github.com/xanzy/go-gitlab/blob/master/repository_files.go) methods, I see that `pathEscape` (like in all other methods) is only used to escape the `project` or `group`. The given `fileName` is escaped using `url.PathEscape` which should properly escape the path following the requirements of the Gitlab API.

> curl --header "PRIVATE-TOKEN: ${token}" https://gitlab.com/api/v4/projects/group%2Fsubgroup%2Fproject can be used to query it.

The fact that this works, means (at least to me) that escaping is not the issue. Can you share any addition info? Maybe a minimal example and the exact error message?

svanharmelen:
@pschwartz do you have any updates or additional details for this one?

svanharmelen:
@pschwartz  any update on this?

ahuret:
Hi @svanharmelen !  I got a case where url encoding seems to break gitlab api.

Here I want to list issues with a label filter:

```
gitlab.Issues.ListProjectIssues(projectID, &gitlab.ListProjectIssuesOptions{Labels: &gitlab.Labels{"my label"}})
```
result to => 
```
curl -H "Private-Token: XXXXXXXXXX" http://gitlab:8081/api/v4/projects/1/issues?labels=%5Bmy+label%5D
```
where
[...]

PR #715, opened 2019-10-14
rollback to ListIssues struct with Labels obj as non pointer

ahuret:
Resolve bug (discussion => #632)


svanharmelen:
Just merged this PR to rollback the changes. It does prove to be a nasty part of the API as without these changes, you cannot unset all labels. But I guess we'll have to look into finding a better solution for this. Thanks!

svanharmelen:
Additional context for people finding this PR:

```
check https://play.golang.org/p/l6tCHGMtBtZ

google/go-querystring#26 says non-nil pointer to slice are not handle properly by the go-querystring library, not exactly in our case but seems there are unexpected behaviors.
```

PR #720, opened 2019-10-24
Fix for the the merge request approvals / allowed approvers function

mvisonneau:
Always submit approver_ids and approved_group_ids for merge request approvals / allowed approvers function


mvisonneau:
According to the documentation and the behaviour of the API, we are required to provide both arguments for every call. If one of them is missing it throws a 4xx error. :(

svanharmelen:
Ok... Then I'll merge this as is... Thanks!

PR #722, opened 2019-10-25
support 'search' parameter in ListBranches method

DenKoren:
gitlab has 'search' parameter in list reposiory branches method

https://docs.gitlab.com/ce/api/branches.html#list-repository-branches

This MR adds support of this parameter to the gi-gitlab client.

### DIFF
--- a/branches.go
+++ b/branches.go
@@ -52,7 +52,7 @@ func (b Branch) String() string {
 // https://docs.gitlab.com/ce/api/branches.html#list-repository-branches
 type ListBranchesOptions struct {
 	ListOptions
-	Search string `url:"search,omitempty" json:"search,omitempty"`
+	Search *string `url:"search,omitempty" json:"search,omitempty"`
 }
 
 // ListBranches gets a list of repository branches from a project, sorted by
--- a/branches.go
+++ b/branches.go
@@ -50,7 +50,10 @@ func (b Branch) String() string {
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/branches.html#list-repository-branches
-type ListBranchesOptions ListOptions
+type ListBranchesOptions struct {
+	ListOptions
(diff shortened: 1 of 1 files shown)


branches.go
```diff
@@ -50,7 +50,10 @@ func (b Branch) String() string {
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/branches.html#list-repository-branches
-type ListBranchesOptions ListOptions
+type ListBranchesOptions struct {
+	ListOptions
+	Search string `url:"search,omitempty" json:"search,omitempty"`
```
svanharmelen:
```suggestion
	Search *string `url:"search,omitempty" json:"search,omitempty"`
```

DenKoren: I just tested the behaviour with this code:
```go
	opt := &gitlab.ListBranchesOptions{
		Search: "",
	}

	encodedJson, _ := json.Marshal(opt)
	values, _ := query.Values(opt)

	fmt.Println(string(encodedJson))
	fmt.Println(values.Encode())
```

Without the pointer it works as expected on empty string:
```text
{}

```

What kind of edge case do you see? I can't figure out circumstances that will lead to empty value appearance in query line or json.

svanharmelen: In this specific case you are right, but in general setting the field with an empty string is not the same as not setting the field at all. So to make the client be able to differentiate between those two, this need to be a pointer.

DenKoren:
One thing I have to warn about: tere is a change of initialization process here.
Before the change this worked:
```
opts := &ListBranchesOptions{
    PerPage: 100
    Page: 1
}
```

Now it will expect this for pagination initialization:
```
opts := &ListBranchesOptions{
    ListOptions: {
        PerPage: 100
        Page: 1
    }
}
```

For already initialized structures everything remains unchanged:
```
opt := &gitlab.ListBranchesOptions{}
opt.Page = 1
opt.PerPage = 100
```

PR #726, opened 2019-10-29
Add hash field in file and metadata response

biohedge:
It would be great if repository file responses have hash info. (As it shown on  https://docs.gitlab.com/ee/api/repository_files.html#get-file-from-repository).
Thanks!

### DIFF
--- a/repository_files.go
+++ b/repository_files.go
@@ -43,6 +43,7 @@ type File struct {
 	Ref      string `json:"ref"`
 	BlobID   string `json:"blob_id"`
 	CommitID string `json:"commit_id"`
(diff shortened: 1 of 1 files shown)


repository_files.go
```diff
@@ -43,6 +43,7 @@ type File struct {
 	Ref      string `json:"ref"`
 	BlobID   string `json:"blob_id"`
 	CommitID string `json:"commit_id"`
+	Sha256   string `json:"content_sha256"`
```
svanharmelen: This should be `SHA` in Go (where abbreviations are in all caps).

biohedge: Oh yes! Thanks, fixed that

PR #732, opened 2019-11-04
Expand group struct fields

### LINKED ISSUE #708: Support group field: runners_token
In order to fix https://github.com/terraform-providers/terraform-provider-gitlab/issues/181
that wants to expose the runners token one would need access to the runners_token field of group (and project but that is already provided)

p.s. group now has a lot more fields than the Group struct provides, might be fun to expose more of that.
p.s.2 If you would prefer me to prepare a merge request let me know.

jrdelarosa8:
Resolves #708

### DIFF
--- a/groups.go
+++ b/groups.go
@@ -32,21 +32,34 @@ type GroupsService struct {
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/groups.html
 type Group struct {
-	ID                   int                `json:"id"`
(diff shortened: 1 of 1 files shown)


jrdelarosa8:
Thank you for the suggestions, I have made the requested changes.

svanharmelen:
LGTM! Thanks @jrdelarosa8!

PR #735, opened 2019-11-16
Add `Locked` to `RunnerDetails` struct

rsrchboy:
It's available in the API, but not exposed until now.

### DIFF
--- a/runners.go
+++ b/runners.go
@@ -77,6 +77,7 @@ type RunnerDetails struct {
 		Name   string `json:"name"`
 		WebURL string `json:"web_url"`
 	} `json:"groups"`
+	Locked         bool `json:"locked"`
 }
 
 // ListRunnersOptions represents the available ListRunners() options.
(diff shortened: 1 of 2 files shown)


runners.go
```diff
@@ -77,6 +77,7 @@ type RunnerDetails struct {
 		Name   string `json:"name"`
 		WebURL string `json:"web_url"`
 	} `json:"groups"`
+	Locked         bool `json:"locked"`
```
svanharmelen: These fields are (more or less) ordered according to the docs, so in that case this field should be above `AccessLevel`. And could you also add the field to the create and update option structs to make support for the field complete? Thanks!

rsrchboy: I've moved it up; AFAICT the update and create structs already have `Locked`, though I might be missing something.

svanharmelen: Your absolutely right! Thanks for fixing the PR!

svanharmelen:
LGTM! Thanks @rsrchboy!

PR #743, opened 2019-12-02
Add OrderBy and Sort parameters to ListContributorsOptions

imkh:
Reference: https://docs.gitlab.com/ce/api/repositories.html#contributors

```
Parameters:

- id (required) - The ID or URL-encoded path of the project owned by the authenticated user
- order_by (optional) - Return contributors ordered by name, email, or commits (orders by commit date) fields. Default is commits
- sort (optional) - Return contributors sorted in asc or desc order. Default is asc
```


svanharmelen:
Thanks!

imkh:
Thanks @svanharmelen! Would it be possible to make a new release following this change?

svanharmelen:
Done 👍

PR #748, opened 2019-12-17
Adding the URL Options "Updated Before and Updated After"

Razz:


deployments.go
```diff
@@ -69,6 +69,8 @@ type ListProjectDeploymentsOptions struct {
 	ListOptions
 	OrderBy *string `url:"order_by,omitempty" json:"order_by,omitempty"`
 	Sort    *string `url:"sort,omitempty" json:"sort,omitempty"`
+	UpdatedBefore *string `url:"updated_before,omitempty" json:"updated_before,omitempty"`
+	UpdatedAfter *string `url:"updated_after,omitempty" json:"updated_after,omitempty"`
```
svanharmelen: Please order these using the ordering used in the docs. So that would mean moving `UpdatedAfter` above `UpdatedBefore`. Also make sure that you run `go fmt` before making a PR so your code is properly formatted.

Last thing is that you know used a `*string` type, but the docs mention these are `datetime` fields. So are you sure `*string` is the right type here?

Razz:
@svanharmelen I apologize for the delay. Updated the ordering and ran it through fmt. 

Re:`datetime` vs `string`. I opted to keep it due to some funkiness. The Gitlab API is very finicky about formatting. If `datetime` doesnt default to `time.(time.RFC3339).String()` it'll be a bad time.   

Thoughts

svanharmelen:
Since [created_at](https://github.com/xanzy/go-gitlab/blob/2da17f99f2afb0a1edf4ed93ef4d7a64d7f40a95/deployments.go#L37) is also using a `*time.Time` I think we should use it here too. And probably we should also add the `updated_at` field to the `Deployments` struct.

PR #750, opened 2019-12-20
Add function to transfer project to a namespace

DevKyleS:
This adds the TransferProject function to move projects to a namespace

### Example
```golang
if projects[p].ID == 10 {
	_, resp, _ := _gitlab.Projects.TransferProject(10, &gitlab.TransferProjectOptions{NamespaceID:gitlab.Int(5)})
	// Check status code
	if resp.StatusCode == 200 {
		//
	}
}
```

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -1517,7 +1517,7 @@ func (s *ProjectsService) StartMirroringProject(pid interface{}, options ...Opti
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/projects.html#transfer-a-project-to-a-new-namespace
 type TransferProjectOptions struct {
-	NamespaceID *int `url:"namespace,omitempty" json:"namespace,omitempty"`
+	Namespace interface{} `url:"namespace,omitempty" json:"namespace,omitempty"`
 }
 
 // TransferProject transfer a project into the specified namespace
--- a/projects.go
+++ b/projects.go
@@ -1512,3 +1512,34 @@ func (s *ProjectsService) StartMirroringProject(pid interface{}, options ...Opti
 
 	return resp, err
 }
+
+// TransferProjectOptions represents the available TransferProject() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/projects.html#transfer-a-project-to-a-new-namespace
+type TransferProjectOptions struct {
+	NamespaceID *int `url:"namespace,omitempty" json:"namespace,omitempty"`
+}
+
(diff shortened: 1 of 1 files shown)


projects.go
```diff
@@ -1512,3 +1512,34 @@ func (s *ProjectsService) StartMirroringProject(pid interface{}, options ...Opti
 
 	return resp, err
 }
+
+// TransferProjectOptions represents the available TransferProject() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/projects.html#transfer-a-project-to-a-new-namespace
+type TransferProjectOptions struct {
+	NamespaceID *int `url:"namespace,omitempty" json:"namespace,omitempty"`
```
svanharmelen:
```suggestion
	Namespace interface{} `url:"namespace,omitempty" json:"namespace,omitempty"`
```

sluongng:
I am not sure about this API. I personally tried it on our internal gitlab instance and it returned a 404 error.

My guess is that gitlab has long deprecated this endpoint in favor of the group-level transfer api instead

reference: https://docs.gitlab.com/ee/api/groups.html#transfer-project-to-group

svanharmelen:
It says `Introduced in GitLab 11.1` in the docs and doesn't seem to mention anything about the API being deprecated. So can you point me to anything that indicates it might be deprecated in favor if the AP you referred to? Thanks!

sluongng:
> It says `Introduced in GitLab 11.1` in the docs and doesn't seem to mention anything about the API being deprecated. So can you point me to anything that indicates it might be deprecated in favor if the AP you referred to? Thanks!

hmm, it seems like the API is still there https://gitlab.com/gitlab-org/gitlab/blob/master/app/controllers/projects_controller.rb#L97 but I wasnt able to use it with my self-hosted instance.

Given the poor documentation in https://docs.gitlab.com/ee/api/projects.html#transfer-a-project-to-a-new-namespace, I could be missing something.

It would be nice to test this against gitlab.com but it requires admin access 😞

DevKyleS:
It's what worked for me on v12.6.0. I'll work on the requested changes as well

projects.go
DevKyleS: Good catch, I learned from this that the interface{} type is a better solution for this. I tested this in my self-hosted instance and was able to pass both int/string types (per Gitlab docs of supported types) and it worked for me on v12.6.0.

svanharmelen:
Thanks @DevKyleS! LGTM!

PR #759, opened 2020-01-25
Corrected merging issue (PR: add system hook events)

Cosby5:
This was a quick and dirty attempt to fix the merging conflict with the previous pull request. Feel free to call me out on errors.

@itscaro is probably a better source of info than I am.


svanharmelen:
Thanks for inspiring me to have a look at that original PR. I'll update that one instead of using this new PR, so this one can be closed. Thanks!

Cosby5:
You're welcome, I don't really care which is used. I just needed that functionality and didn't want to fully develop a non-unified solution. I honestly don't even know if it compiled.

Issue #762, opened 2020-02-03
Unable to set extended group attributes on creation


justinkillen:
Gitlab API supports the following fields in group creation that aren't enabled in groups/CreateGroupOptions:

require_two_factor_authentication
two_factor_grace_period
project_creation_level
subgroup_creation_level

(https://docs.gitlab.com/ce/api/groups.html#new-group)

justinkillen:
https://github.com/xanzy/go-gitlab/pull/775

PR #763, opened 2020-02-03
Add group options

justinkillen:
For https://github.com/xanzy/go-gitlab/issues/762

### DIFF
--- a/groups.go
+++ b/groups.go
@@ -125,13 +125,17 @@ func (s *GroupsService) GetGroup(gid interface{}, options ...OptionFunc) (*Group
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/groups.html#new-group
 type CreateGroupOptions struct {
-	Name                 *string          `url:"name,omitempty" json:"name,omitempty"`
-	Path                 *string          `url:"path,omitempty" json:"path,omitempty"`
-	Description          *string          `url:"description,omitempty" json:"description,omitempty"`
-	Visibility           *VisibilityValue `url:"visibility,omitempty" json:"visibility,omitempty"`
-	LFSEnabled           *bool            `url:"lfs_enabled,omitempty" json:"lfs_enabled,omitempty"`
-	RequestAccessEnabled *bool            `url:"request_access_enabled,omitempty" json:"request_access_enabled,omitempty"`
-	ParentID             *int             `url:"parent_id,omitempty" json:"parent_id,omitempty"`
+	Name                  *string          `url:"name,omitempty" json:"name,omitempty"`
+	Path                  *string          `url:"path,omitempty" json:"path,omitempty"`
+	Description           *string          `url:"description,omitempty" json:"description,omitempty"`
+	Visibility            *VisibilityValue `url:"visibility,omitempty" json:"visibility,omitempty"`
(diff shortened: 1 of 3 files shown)


groups.go
```diff
@@ -125,13 +125,17 @@ func (s *GroupsService) GetGroup(gid interface{}, options ...OptionFunc) (*Group
...
+	Path                  *string          `url:"path,omitempty" json:"path,omitempty"`
+	Description           *string          `url:"description,omitempty" json:"description,omitempty"`
+	Visibility            *VisibilityValue `url:"visibility,omitempty" json:"visibility,omitempty"`
+    RequireTwoFactorAuth  *bool            `url:"require_two_factor_authentication,omitempty" json:"require_two_factor_authentication,omitempty"`
+    TwoFactorGracePeriod  *int             `url:"two_factor_grace_period,omitempty" json:"two_factor_grace_period,omitempty"`
+    ProjectCreationLevel  *string          `url:"project_creation_level,omitempty" json:"project_creation_level,omitempty"`
+    SubGroupCreationLevel *string          `url:"subgroup_creation_level,omitempty" json:"subgroup_creation_level,omitempty"`
+	LFSEnabled            *bool            `url:"lfs_enabled,omitempty" json:"lfs_enabled,omitempty"`
+	RequestAccessEnabled  *bool            `url:"request_access_enabled,omitempty" json:"request_access_enabled,omitempty"`
+	ParentID              *int             `url:"parent_id,omitempty" json:"parent_id,omitempty"`
```
svanharmelen: Please use `go fmt` to make sure the code is formatted properly.

groups.go
```diff
@@ -125,13 +125,17 @@ func (s *GroupsService) GetGroup(gid interface{}, options ...OptionFunc) (*Group
...
-	RequestAccessEnabled *bool            `url:"request_access_enabled,omitempty" json:"request_access_enabled,omitempty"`
-	ParentID             *int             `url:"parent_id,omitempty" json:"parent_id,omitempty"`
+	Name                  *string          `url:"name,omitempty" json:"name,omitempty"`
+	Path                  *string          `url:"path,omitempty" json:"path,omitempty"`
+	Description           *string          `url:"description,omitempty" json:"description,omitempty"`
+	Visibility            *VisibilityValue `url:"visibility,omitempty" json:"visibility,omitempty"`
+    RequireTwoFactorAuth  *bool            `url:"require_two_factor_authentication,omitempty" json:"require_two_factor_authentication,omitempty"`
+    TwoFactorGracePeriod  *int             `url:"two_factor_grace_period,omitempty" json:"two_factor_grace_period,omitempty"`
+    ProjectCreationLevel  *string          `url:"project_creation_level,omitempty" json:"project_creation_level,omitempty"`
+    SubGroupCreationLevel *string          `url:"subgroup_creation_level,omitempty" json:"subgroup_creation_level,omitempty"`
```
svanharmelen: According to the docs this should be an `*int`, but I wonder if that could be wrong in the docs?

justinkillen: Yes, the field type is wrong in the table - it's clearly a string from the description and the example responses.  I created an issue for the gitlab docs team: https://gitlab.com/gitlab-org/gitlab-docs/issues/642

groups.go
justinkillen: will do

justinkillen:
Added all fields from the CreateGroup API

PR #769, opened 2020-02-08
Add auto_ssl_enabled to Pages Domain

alexives:
The auto_ssl_enabled option was added in gitlab 12.1 and I'd like to have it here! Let me know what I need to do to get this in!

### DIFF
--- a/pages_domains.go
+++ b/pages_domains.go
@@ -18,6 +18,7 @@ type PagesDomainsService struct {
 // GitLab API docs: https://docs.gitlab.com/ce/api/pages_domains.html
 type PagesDomain struct {
 	Domain           string     `json:"domain"`
+	AutoSslEnabled   bool       `json:"auto_ssl_enabled"`
 	URL              string     `json:"url"`
 	ProjectID        int        `json:"project_id"`
 	Verified         bool       `json:"verified"`
@@ -107,11 +108,12 @@ func (s *PagesDomainsService) GetPagesDomain(pid interface{}, domain string, opt
 // CreatePagesDomainOptions represents the available CreatePagesDomain() options.
 //
 // GitLab API docs:
-// // https://docs.gitlab.com/ce/api/pages_domains.html#create-new-pages-domain
+// https://docs.gitlab.com/ce/api/pages_domains.html#create-new-pages-domain
 type CreatePagesDomainOptions struct {
-	Domain      *string `url:"domain,omitempty" json:"domain,omitempty"`
-	Certificate *string `url:"certifiate,omitempty" json:"certifiate,omitempty"`
-	Key         *string `url:"key,omitempty" json:"key,omitempty"`
+	Domain         *string `url:"domain,omitempty" json:"domain,omitempty"`
+	AutoSslEnabled *bool   `url:"auto_ssl_enabled,omitempty" json:"auto_ssl_enabled,omitempty"`
+	Certificate    *string `url:"certifiate,omitempty" json:"certifiate,omitempty"`
+	Key            *string `url:"key,omitempty" json:"key,omitempty"`
 }
 
 // CreatePagesDomain creates a new project pages domain.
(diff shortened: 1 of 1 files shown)


pages_domains.go
```diff
@@ -144,8 +146,9 @@ func (s *PagesDomainsService) CreatePagesDomain(pid interface{}, opt *CreatePage
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/pages_domains.html#update-pages-domain
 type UpdatePagesDomainOptions struct {
-	Cerificate *string `url:"certifiate" json:"certifiate"`
-	Key        *string `url:"key" json:"key"`
+	Cerificate     *string `url:"certifiate" json:"certifiate"`
```
alexives: I don't want to do it as a part of this PR (I'm worried about backwards compatibility), but I wonder if there's a good way to fix the missing `t` in `Certificate` here.

alexives:
For more details on auto ssl, see https://docs.gitlab.com/ee/user/project/pages/custom_domains_ssl_tls_certification/lets_encrypt_integration.html

pages_domains.go
```diff
@@ -109,9 +110,10 @@ func (s *PagesDomainsService) GetPagesDomain(pid interface{}, domain string, opt
 // GitLab API docs:
 // // https://docs.gitlab.com/ce/api/pages_domains.html#create-new-pages-domain
 type CreatePagesDomainOptions struct {
-	Domain      *string `url:"domain,omitempty" json:"domain,omitempty"`
-	Certificate *string `url:"certifiate,omitempty" json:"certifiate,omitempty"`
-	Key         *string `url:"key,omitempty" json:"key,omitempty"`
+	Domain         *string `url:"domain,omitempty" json:"domain,omitempty"`
+	Certificate    *string `url:"certifiate,omitempty" json:"certifiate,omitempty"`
+	Key            *string `url:"key,omitempty" json:"key,omitempty"`
+	AutoSslEnabled *bool   `url:"auto_ssl_enabled,omitempty" json:"auto_ssl_enabled,omitempty"`
```
svanharmelen: In general I try to use the ordering as it's used in the docs. So in this case the new field should be below `Domain`. Could you update that?

pages_domains.go
svanharmelen: Please do fix it in your PR. Since nobody reported this typo, I'm inclined to think that it's not being used much and it's clearly wrong.

pages_domains.go
```diff
@@ -144,8 +146,9 @@ func (s *PagesDomainsService) CreatePagesDomain(pid interface{}, opt *CreatePage
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/pages_domains.html#update-pages-domain
 type UpdatePagesDomainOptions struct {
-	Cerificate *string `url:"certifiate" json:"certifiate"`
-	Key        *string `url:"key" json:"key"`
+	Cerificate     *string `url:"certifiate" json:"certifiate"`
+	Key            *string `url:"key" json:"key"`
+	AutoSslEnabled *bool   `url:"auto_ssl_enabled,omitempty" json:"auto_ssl_enabled,omitempty"`
```
svanharmelen: Same about the ordering...

pages_domains.go
```diff
@@ -109,9 +110,10 @@ func (s *PagesDomainsService) GetPagesDomain(pid interface{}, domain string, opt
 // GitLab API docs:
 // // https://docs.gitlab.com/ce/api/pages_domains.html#create-new-pages-domain
```
svanharmelen: You didn't add it, but I just noticed the double `//` pair. Could you maybe remove the additional one while updating your PR? Thanks!

alexives:
Wow! Thanks for your fast feedback! I'll those changes are up now!

pages_domains.go
alexives: This is fixed, but for some reason it doesn't show as an `outdated` change ¯\_(ツ)_/¯

PR #773, opened 2020-02-12
Added missing group boards update and delete functions

paulvollmer:
The PR add the missing endpoint functions for the group issue board's.

### DIFF
--- a/group_boards.go
+++ b/group_boards.go
@@ -164,8 +164,8 @@ func (s *GroupIssueBoardsService) UpdateIssueBoard(gid interface{}, board int, o
 		return nil, nil, err
 	}
 
-	var gib GroupIssueBoard
-	resp, err := s.client.Do(req, gbl)
+	gib := new(GroupIssueBoard)
+	resp, err := s.client.Do(req, gib)
 	if err != nil {
 		return nil, resp, err
 	}
--- a/group_boards.go
+++ b/group_boards.go
@@ -144,7 +144,7 @@ type UpdateGroupIssueBoardOptions struct {
 	Name        *string `url:"name,omitempty" json:"name,omitempty"`
 	AssigneeID  *int    `url:"assignee_id,omitempty" json:"assignee_id,omitempty"`
 	MilestoneID *int    `url:"milestone_id,omitempty" json:"milestone_id,omitempty"`
-	Labels      *Label  `url:"labels,omitempty" json:"labels,omitempty"`
+	Labels      *Labels `url:"labels,omitempty" json:"labels,omitempty"`
 	Weight      *int    `url:"weight,omitempty" json:"weight,omitempty"`
 }
 
--- a/group_boards.go
+++ b/group_boards.go
@@ -144,7 +144,7 @@ type UpdateGroupIssueBoardOptions struct {
 	Name        *string `url:"name,omitempty" json:"name,omitempty"`
 	AssigneeID  *int    `url:"assignee_id,omitempty" json:"assignee_id,omitempty"`
 	MilestoneID *int    `url:"milestone_id,omitempty" json:"milestone_id,omitempty"`
-	Labels      *string `url:"labels,omitempty" json:"labels,omitempty"`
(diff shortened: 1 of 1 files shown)


group_boards.go
```diff
@@ -136,6 +136,66 @@ func (s *GroupIssueBoardsService) GetGroupIssueBoard(gid interface{}, board int,
 	return gib, resp, err
 }
 
+type UpdateGroupIssueBoardOptions struct {
+	ID          *string `url:"id" json:"id"`
+	BoardID     *int    `url:"board_id" json:"board_id"`
```
svanharmelen: `ID` and `BoardID` are already in the function signature (`gid` and `board`), so they should be taken out of the options struct.

group_boards.go
```diff
@@ -136,6 +136,66 @@ func (s *GroupIssueBoardsService) GetGroupIssueBoard(gid interface{}, board int,
...
+	Name        *string `url:"name" json:"name"`
+	AssigneeID  *int    `url:"assignee_id" json:"assignee_id"`
+	MilestoneID *int    `url:"milestone_id" json:"milestone_id"`
+	Labels      *string `url:"labels" json:"labels"`
```
svanharmelen: There is a `Label` type in this package which is probably a better fit for this one. Of do you have a specific reason to use a `*string` here?

svanharmelen: And these fields all need `omitempty`...

group_boards.go
```diff
@@ -136,6 +136,66 @@ func (s *GroupIssueBoardsService) GetGroupIssueBoard(gid interface{}, board int,
...
+// https://docs.gitlab.com/ce/api/group_boards.html#update-a-group-issue-board-premium
+func (s *GroupIssueBoardsService) UpdateIssueBoard(gid interface{}, board int, opt *UpdateGroupIssueBoardOptions, options ...OptionFunc) (*GroupIssueBoard, *Response, error) {
+	group, err := parseID(gid)
+	if err != nil {
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("groups/%s/boards/%d",
+		pathEscape(group),
+		board,
+	)
```
svanharmelen: While this is not wrong, this is short enough to put on one line....

group_boards.go
```diff
@@ -136,6 +136,66 @@ func (s *GroupIssueBoardsService) GetGroupIssueBoard(gid interface{}, board int,
...
+// https://docs.gitlab.com/ce/api/group_boards.html#delete-a-group-issue-board-premium
+func (s *GroupIssueBoardsService) DeleteIssueBoard(gid interface{}, board int, options ...OptionFunc) (*Response, error) {
+	group, err := parseID(gid)
+	if err != nil {
+		return nil, err
+	}
+	u := fmt.Sprintf("groups/%s/boards/%d",
+		pathEscape(group),
+		board,
+	)
```
svanharmelen: While this is not wrong, this is short enough to put on one line....

paulvollmer:
Thank you for the review. I'll fix it and let you know when I'm done.

group_boards.go
paulvollmer: i'm using `*string` because the docs say that the `labels` are "Comma-separated list of label names which the board should be scoped to". Because of this i think it should be a string.

paulvollmer:
@svanharmelen i did the fixes...

group_boards.go
svanharmelen: I know, but that is what I created the `Label` type for. It has custom (un)marshaling methods so it can be used as a slice but will be serialized as a comma separated string. See here for more info: https://github.com/xanzy/go-gitlab/blob/master/issues.go#L105

paulvollmer: if we have the `Labels` type then it's better to use it. i changed it from string to Labels now.

paulvollmer:
Great. @svanharmelen  can you create a git tag for it?

svanharmelen:
Done 👍

PR #777, opened 2020-02-19
Add 2 request parameters to list-projects-pipelines

KingCrunch:
`/projects/<id>/pipelines` has two additional parameters `updated_before` and `updated-after`.

### DIFF
--- a/pipelines.go
+++ b/pipelines.go
@@ -101,15 +101,17 @@ func (p PipelineInfo) String() string {
 // GitLab API docs: https://docs.gitlab.com/ce/api/pipelines.html#list-project-pipelines
 type ListProjectPipelinesOptions struct {
 	ListOptions
-	Scope      *string          `url:"scope,omitempty" json:"scope,omitempty"`
-	Status     *BuildStateValue `url:"status,omitempty" json:"status,omitempty"`
-	Ref        *string          `url:"ref,omitempty" json:"ref,omitempty"`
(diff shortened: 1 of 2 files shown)


pipelines.go
```diff
@@ -101,15 +101,17 @@ func (p PipelineInfo) String() string {
...
-	Sort       *string          `url:"sort,omitempty" json:"sort,omitempty"`
+	Scope         *string          `url:"scope,omitempty" json:"scope,omitempty"`
+	Status        *BuildStateValue `url:"status,omitempty" json:"status,omitempty"`
+	Ref           *string          `url:"ref,omitempty" json:"ref,omitempty"`
+	SHA           *string          `url:"sha,omitempty" json:"sha,omitempty"`
+	YamlErrors    *bool            `url:"yaml_errors,omitempty" json:"yaml_errors,omitempty"`
+	Name          *string          `url:"name,omitempty" json:"name,omitempty"`
+	Username      *string          `url:"username,omitempty" json:"username,omitempty"`
+	UpdatedBefore *time.Time       `url:"update_before,omitempty" json:"updated_before,omitempty"`
+	UpdatedAfter  *time.Time       `url:"updated_after,omitempty" json:"updated_after,omitempty"`
```
svanharmelen: Looks good, but can you order them as they are ordered in the docs? So `UpdatedAfter` should be above `UpdatedBefore`. I try to follow the ordering used in the docs so it's easy to compare and find things.

PR #778, opened 2020-02-20
Adding GetConfiguration into MergeRequestApprovals

mocnik:
Added GetConfiguration (https://docs.gitlab.com/ee/api/merge_request_approvals.html#get-configuration-1) into MergeRequestsApprovals.

### DIFF
--- a/merge_request_approvals.go
+++ b/merge_request_approvals.go
@@ -40,6 +40,31 @@ func (m MergeRequestApprovals) String() string {
 	return Stringify(m)
 }
 
+// GetConfiguration shows information about single merge request approvals
+//
+// GitLab API docs:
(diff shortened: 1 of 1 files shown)


merge_request_approvals.go
```diff
@@ -40,6 +40,31 @@ func (m MergeRequestApprovals) String() string {
...
+	}
+
+	m := new(MergeRequestApprovals)
+	resp, err := s.client.Do(req, m)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return m, resp, err
+}
```
svanharmelen: Looks good, but it would be nice if we place this new method right next to the related methods. So please move this down so it's just above the `ChangeApprovalConfiguration` method. Thanks!

svanharmelen:
Thanks again, looks great now 👍

PR #783, opened 2020-02-25
Update GroupMember api object to have group_saml_identity in the output

pschwartz:
About 2 months ago the GroupMembers was updated to allow the display of group saml identity links. This change updates the GroupMember object to include these from the API return.

### DIFF
--- a/group_members.go
+++ b/group_members.go
@@ -28,30 +28,30 @@ type GroupMembersService struct {
 	client *Client
 }
 
-// GroupMemberSamlIdentity represents the SAML Identity link for the group member.
+// GroupMemberSAMLIdentity represents the SAML Identity link for the group member.
 //
-// GitLab API docs: TBD
+// GitLab API docs: https://docs.gitlab.com/ce/api/members.html#list-all-members-of-a-group-or-project
 // Gitlab MR for API change: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/20357
 // Gitlab MR for API Doc change: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/25652
-type GroupMemberSamlIdentity struct {
+type GroupMemberSAMLIdentity struct {
 	ExternUID      string `json:"extern_uid"`
 	Provider       string `json:"provider"`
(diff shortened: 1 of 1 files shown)


group_members.go
```diff
@@ -28,18 +28,30 @@ type GroupMembersService struct {
 	client *Client
 }
 
+// GroupMemberSamlIdentity represents the SAML Identity link for the group member.
+//
+// GitLab API docs: TBD
+// Gitlab MR for API change: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/20357
+// Gitlab MR for API Doc change: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/25652
+type GroupMemberSamlIdentity struct {
```
svanharmelen: In Go initialisms or acronyms are always written in al caps: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

So for example `GroupMemberSamlIdentity` should be `GroupMemberSAMLIdentity` and `SAMLProviderID` should be `SAMLProviderID`.

group_members.go
```diff
@@ -28,18 +28,30 @@ type GroupMembersService struct {
...
-	AccessLevel AccessLevelValue `json:"access_level"`
+	ID                int                     `json:"id"`
+	Username          string                  `json:"username"`
+	Name              string                  `json:"name"`
+	State             string                  `json:"state"`
+	AvatarURL         string                  `json:"avatar_url"`
+	WebURL            string                  `json:"web_url"`
+	ExpiresAt         *ISOTime                `json:"expires_at"`
+	AccessLevel       AccessLevelValue        `json:"access_level"`
+	GroupSamlIdentity GroupMemberSamlIdentity `json:"group_saml_identity"`
```
svanharmelen: This needs to be a pointer field.

group_members.go
pschwartz: Changed.

group_members.go
pschwartz: Changed.

pschwartz:
@svanharmelen Changes made, please re-review.

PR #789, opened 2020-03-03
Support for creating and updating project deployments

marcinwyszynski:
Hey folks, here's a little PR adding two extra methods to the `DeploymentsService`. I did my best to maintain the same coding style as everywhere else in this project.

### DIFF
--- a/deployments.go
+++ b/deployments.go
@@ -183,7 +183,7 @@ func (s *DeploymentsService) UpdateProjectDeployment(pid interface{}, deployment
 	}
 	u := fmt.Sprintf("projects/%s/deployments/%d", pathEscape(project), deployment)
 
-	req, err := s.client.NewRequest("POST", u, opt, options)
+	req, err := s.client.NewRequest("PUT", u, opt, options)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/deployments.go
+++ b/deployments.go
@@ -124,20 +124,11 @@ func (s *DeploymentsService) GetProjectDeployment(pid interface{}, deployment in
 //
 // GitLab API docs: https://docs.gitlab.com/ee/api/deployments.html#create-a-deployment
 type CreateProjectDeploymentOptions struct {
-	// The name of the environment to create the deployment for
-	Environment string `json:"environment"`
-
-	// The name of the branch or tag that is deployed
-	Ref string `json:"ref"`
-
-	// The SHA of the commit that is deployed
-	SHA string `json:"sha"`
-
-	// A boolean that indicates if the deployed ref is a tag (true) or not (false)
-	Tag bool `json:"tag"`
-
-	// The status of the deployment
-	Status BuildStateValue `json:"status"`
+	Environment *string                `url:"environment,omitempty" json:"environment,omitempty"`
+	Ref         *string                `url:"ref,omitempty" json:"ref,omitempty"`
+	SHA         *string                `url:"sha,omitempty" json:"sha,omitempty"`
+	Tag         *bool                  `url:"tag,omitempty" json:"tag,omitempty"`
+	Status      *DeploymentStatusValue `url:"status,omitempty" json:"status,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


deployments.go
```diff
@@ -118,3 +118,81 @@ func (s *DeploymentsService) GetProjectDeployment(pid interface{}, deployment in
...
+
+	// The SHA of the commit that is deployed
+	SHA string `json:"sha"`
+
+	// A boolean that indicates if the deployed ref is a tag (true) or not (false)
+	Tag bool `json:"tag"`
+
+	// The status of the deployment
+	Status BuildStateValue `json:"status"`
+}
```
svanharmelen: Please have a look at the other code in this package and try to make this fit in nicely. This means that all these options need to be pointers, they need a `url` tag and they all need the `omitempty` option. Additionally the field names and the link to the docs already give enough info, so we don't comment each field individually as well.

deployments.go
```diff
@@ -118,3 +118,81 @@ func (s *DeploymentsService) GetProjectDeployment(pid interface{}, deployment in
...
+
+// UpdateProjectDeploymentOptions represents the available
+// UpdateProjectDeployment() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/deployments.html#updating-a-deployment
+type UpdateProjectDeploymentOptions struct {
+	// The status of the deployment
+	Status BuildStateValue `json:"status"`
+}
```
svanharmelen: Same comments apply as for the other option struct.

gitlab.go
```diff
@@ -83,6 +83,7 @@ type BuildStateValue string
 // These constants represent all valid build states.
 const (
 	Pending  BuildStateValue = "pending"
+	Created  BuildStateValue = "created"
```
svanharmelen: I don't think we can reuse these constants, but instead we should create dedicated `DeploymentStatus` type for this.

deployments.go
marcinwyszynski: > options need to be pointers

I mean, it's your repo and I don't want to be difficult, so if you really insist I can make this change, but my personal view is that since these fields are marked as required in the documentation, having them as pointers adds no value, just makes the API more difficult to use.

gitlab.go
marcinwyszynski: Fair enough. As an aside, and I don't think this is something that can be addressed easily, but the names of these constants should contain their type (eg. `BuildStateValuePending`) to keep the API unambiguous.

svanharmelen: Yes, that is very true! Guess the package has grown quite a bit since it was first created and this was never addressed...

deployments.go
svanharmelen: I understand that, but at least it means that all option structs can be used in the same way. If we would differentiate between required and optional fields it would mean that you would need to look at each field to check if it expects a pointer to a type or just a type.

So in the end I think it does make it easier to work with as it's at least super consistent...

marcinwyszynski: Fair enough and no problem mate, you're the boss 🤷‍♂️

gitlab.go
marcinwyszynski: I've namespaced my bit.

marcinwyszynski:
PTAL @svanharmelen 👀👁👁‍🗨

gitlab.go
```diff
@@ -92,6 +92,18 @@ const (
...
+type DeploymentStatusValue string
+
+// These constants represent all valid deployment statuses.
+const (
+	DeploymentStatusCreated  = "created"
+	DeploymentStatusRunning  = "running"
+	DeploymentStatusSuccess  = "success"
+	DeploymentStatusFailed   = "failed"
+	DeploymentStatusCanceled = "canceled"
+)
```
svanharmelen: Nice 👍

deployments.go
marcinwyszynski: Just a figure of speech :)

Issue #782, opened 2020-02-25
Gitlab introduced soft delete of projects and groups


ringods:
Filing this ticket not as a bug report, more as an awareness.

The test suites for the Gitlab Terraform & Pulumi providers became instable since Gitlab 12.6. After some investigation, it seems that soft-delete of Projects was introduced in that version by **changing the behaviour of `DELETE /projects/{id}`**.

More context here: 
https://github.com/terraform-providers/terraform-provider-gitlab/issues/263
which links to even more gitlab tickets.

So be aware that a delete is not necessarily a hard delete.

svanharmelen:
If I read it correctly, there now is a `pending_delete` field in the returned `project` struct? If so, we probably need to add it so it can (maybe) be used to update the logic in the TF provider to be able to cope with this...

ringods:
I asked for some help from Gitlab:

https://gitlab.com/gitlab-org/gitlab/issues/33257#note_296585615

svanharmelen:
Maybe good to have the discussion there, but not what kind of help you mean? Adding the `pending_delete` fields should be a tiny PR right? Not sure what else you are looking for?

ringods:
@svanharmelen if it is clear to you what the response structure and fields are for a delete request of a project or group, then no additional help is needed. If not, then it is mainly my request to the Gitlab people to help us out to clarify how we can detect the `pending delete` status programmatically.

Nevertheless, are you able to implement the `pending delete` status to be interpreted as "the repo is deleted"?

svanharmelen:
@ringods can you confirm that the fields in PR #791 are filled correctly? If so, I can merge the request and it can be used in the TF provider...

ringods:
@svanharmelen I finally came around integrating the soft-delete changes in Terraform. What I notice from debug logs is that the server *never* sets the `pending_delete` field in the response JSON of groups and repositories.

Here is a link to a testsuite run with DEBUG logs enabled:

https://github.com/terraform-providers/terraform-provider-gitlab/runs/552631042?check_suite_focus=true

*Note:* download the logs from the menu in the 3 dots in the upper right of the black logs screen for faster search. 😉 

If you search for `marked_for_deletion_at` in the logs, you will see the response given back from the server. When a non-`null` value is set, there is no `pending_delete` field in the same response JSON. The `marked_for_deletion_at` is only added to the `Project` type, not to `Group` so I can't detect a soft delete on groups.

[...]

PR #791, opened 2020-03-03
Add a `pending_delete` field to projects and groups

svanharmelen:


svanharmelen:
I think this adds the missing bits in order to be able to better support soft deleted project in the TF provider.

I'm not in a position to test this, so please let me know if the fields are properly populated so they can be used to determine if a project create failed because it already exists, or because it is currently soft deleted.

Fixes #782

projects.go
```diff
@@ -94,9 +95,10 @@ type Project struct {
 	} `json:"shared_with_groups"`
 	Statistics        *ProjectStatistics `json:"statistics"`
 	Links             *Links             `json:"_links,omitempty"`
-	CIConfigPath      *string            `json:"ci_config_path"`
+	CIConfigPath      string             `json:"ci_config_path"`
```
ringods: @svanharmelen Is this an intentional change as I don't think it is related to the "pending delete"

svanharmelen: It’s not related but it is intentional. Could make a dedicated PR for it, but thought to just add it when I noticed it...

ringods: You don't have to put this in a dedicated PR. I just want to rule out any accidental mistake. ;-)

svanharmelen: Thanks for bringing it up 👍🏻

PR #796, opened 2020-03-09
Ability to use any HTTP Client implementations that follow net/http i…

### LINKED ISSUE #630: Does not handle retries or backoff for Gitlab.com API endpoints
For those of us using Gitlab SaaS, we're quite frequently hitting the Gitlab rate limits.

Gitlab returns the following "standard" rate limit retry headers in 429 Too Many Request responses:

```
< RateLimit-Limit: 600
< RateLimit-Observed: 29
< RateLimit-Remaining: 569
< RateLimit-Reset: 1558720990
```

It would be ideal if `go-gitlab` properly interpreted these and implemented standard backoff/retry logic for Gitlab requests, as [node-gitlab does](https://github.com/jdalrymple/node-gitlab/issues/73)

Obviously it _can_ be reimplemented every time in the surrounding code that invokes `go-gitlab`, but I think logic this basic belongs in this library.

Am going to start work on a PR for this, however if there is disagreement about whether this functionality belongs in `go-gitlab` upstream or not, I'd like to hear it.

totobest:
Fix #630 

Allow usage of any HTTP Client implementations like https://github.com/hashicorp/go-retryablehttp.
The only requirement is that it has a Do(*http.Request) (*http.Response, error) method.

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -318,10 +318,15 @@ const (
 	UserEventTargetType         EventTargetTypeValue = "user"
 )
 
+// HTTPClient interface
+type HTTPClient interface {
+	Do(*http.Request) (*http.Response, error)
+}
+
 // A Client manages communication with the GitLab API.
 type Client struct {
 	// HTTP client used to communicate with the API.
-	client *http.Client
+	client HTTPClient
 
 	// Base URL for API requests. Defaults to the public GitLab API, but can be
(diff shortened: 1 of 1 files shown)


svanharmelen:
Thanks for your efforts @totobest, but this is not something I would like to implement. Instead I created a new PR (#800) which adds rate limit and retry logic, so if you could take that one for a test that would be great! Thanks again 👍

PR #812, opened 2020-03-30
WIP Terraform endpoint support

ash2k:
Hello there, this is a WIP. I'd like to get early feedback on the general direction of this work. I tried to be consistent with the existing codebase. I intent to add more tests, links to docs, etc and actually test against a real GitLab instance once the support is available (it's being worked on).

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,8 +97,21 @@ access different parts of the GitLab API. For example, to list all
 users:
 
 ```go
-git := gitlab.NewClient(nil, "yourtokengoeshere")
-//git.SetBaseURL("https://git.mydomain.com/api/v4")
+git, err := gitlab.NewClient("yourtokengoeshere")
+if err != nil {
+  log.Fatalf("Failed to create client: %v", err)
+}
+users, _, err := git.Users.ListUsers(&gitlab.ListUsersOptions{})
+```
+
+There are a few `With...` option functions that can be used to customize
+the API client. For example, to set a custom base URL:
+
+```go
+git, err := gitlab.NewClient("yourtokengoeshere", WithBaseURL("https://git.mydomain.com/api/v4"))
+if err != nil {
+  log.Fatalf("Failed to create client: %v", err)
+}
 users, _, err := git.Users.ListUsers(&gitlab.ListUsersOptions{})
 ```
 
@@ -106,7 +119,7 @@ Some API methods have optional parameters that can be passed. For example,
 to list all projects for user "svanharmelen":
 
 ```go
-git := gitlab.NewClient(nil)
+git := gitlab.NewClient("yourtokengoeshere")
 opt := &ListProjectsOptions{Search: gitlab.String("svanharmelen")}
 projects, _, err := git.Projects.ListProjects(opt)
 ```
@@ -126,7 +139,10 @@ import (
 )
 
 func main() {
-	git := gitlab.NewClient(nil, "yourtokengoeshere")
+	git, err := gitlab.NewClient("yourtokengoeshere")
+	if err != nil {
+		log.Fatalf("Failed to create client: %v", err)
+	}
 
 	// Create new project
 	p := &gitlab.CreateProjectOptions{
--- a/access_requests.go
+++ b/access_requests.go
@@ -39,7 +39,7 @@ type ListAccessRequestsOptions ListOptions
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/access_requests.html#list-access-requests-for-a-group-or-project
-func (s *AccessRequestsService) ListProjectAccessRequests(pid interface{}, opt *ListAccessRequestsOptions, options ...OptionFunc) ([]*AccessRequest, *Response, error) {
+func (s *AccessRequestsService) ListProjectAccessRequests(pid interface{}, opt *ListAccessRequestsOptions, options ...RequestOptionFunc) ([]*AccessRequest, *Response, error) {
 	project, err := parseID(pid)
 	if err != nil {
 		return nil, nil, err
@@ -65,7 +65,7 @@ func (s *AccessRequestsService) ListProjectAccessRequests(pid interface{}, opt *
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/access_requests.html#list-access-requests-for-a-group-or-project
-func (s *AccessRequestsService) ListGroupAccessRequests(gid interface{}, opt *ListAccessRequestsOptions, options ...OptionFunc) ([]*AccessRequest, *Response, error) {
+func (s *AccessRequestsService) ListGroupAccessRequests(gid interface{}, opt *ListAccessRequestsOptions, options ...RequestOptionFunc) ([]*AccessRequest, *Response, error) {
 	group, err := parseID(gid)
 	if err != nil {
 		return nil, nil, err
@@ -91,7 +91,7 @@ func (s *AccessRequestsService) ListGroupAccessRequests(gid interface{}, opt *Li
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/access_requests.html#request-access-to-a-group-or-project
-func (s *AccessRequestsService) RequestProjectAccess(pid interface{}, options ...OptionFunc) (*AccessRequest, *Response, error) {
+func (s *AccessRequestsService) RequestProjectAccess(pid interface{}, options ...RequestOptionFunc) (*AccessRequest, *Response, error) {
 	project, err := parseID(pid)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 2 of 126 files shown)


svanharmelen:
@ash2k I took the liberty to rebase your branch on PR #813 as that includes a few of your changes already and caused a couple of merge conflicts.

PR #813 contains a backwards incompatible change, but it introduces a pattern to have a lot of flexibility (without breaking backwards compatibility) going forward. I plan on merging this PR in the coming days, so it should be in before this PR will be ready to be merged.

I didn't yet look in detail at the content of `terraform.go`, but high level it seems reasonable. I'll have a closer look tomorrow and will come back with any feedback or suggestions.

ash2k:
@svanharmelen thanks for rebasing it! I've left some comments on that PR. It's good stuff! Also, I've made some progress here (but it's not done yet), please have a look. I think you might borrow some changes from this PR OR we can keep them here and just merge it if/when it's good to merge.

More specifically, please see `NewRequestRaw()` and how I rewrote `NewRequest()` using this new method and helpers. And also how it's used in `terraform.go`. It was a hassle to set both the request body and the query parameters before but now it's trivial. Also it matches the general pattern. WDYT?

Feel free to make any breaking changes without looking how it will impact this PR, I'm fine with rebasing. Don't want to slow the progress down :)

ash2k:
I've also opened https://github.com/hashicorp/go-retryablehttp/pull/88 that introduces some handy helper methods to clean up/optimize the code in this library later.

svanharmelen:
> More specifically, please see NewRequestRaw() and how I rewrote NewRequest() using this new method and helpers. And also how it's used in terraform.go. It was a hassle to set both the request body and the query parameters before but now it's trivial. Also it matches the general pattern. WDYT?

Super nice as I had the exact same idea, but then calling the new function `RawRequest` 😄 I'm pretty busy today (and do all this in my spare time), but will make sure to have a look at how you implemented things end of my day (CET). I think I'm going to create a commit with some suggestions/tweaks, but I like the direction this is going.

Also thanks for the feedback on the other PR, there are a few things in there that make sense and I needed to have another look at anyway. So will probably update a few things there as well.

svanharmelen:
My day is long, long gone by now and I just manage to catch up on all things that were going on. So I'm sorry, but my review on this one has to wait until tomorrow...

I did have a quick look and I think I do want to change things (mainly in `gitlab.go`) just a bit (I'm not a big fan of using (abusing?) the options functions for everything). But let me spend some time on it tomorrow and get back at you with a suggestion on how to move forward.

The essence of what you tried to do with the NewRequestRaw` matches what I was also thinking about, so that will stay (more or less) the same. Cheers!

ash2k:
Rebased on top of your branch again and temporarily set the PR to target it to make it easier to see only the commits I made.

svanharmelen:
I just had a closer look at the code and I cannot help but wonder about quite a few things. I think it would be extremely helpful for me to better understand what the API is going to look like. Is there maybe a (preliminary) design that you can share?

I assumed that the new API endpoints would (more or less) follow the same patterns and style of the existing API endpoints. Maybe with a few little tweaks or additions (like adding the additional MD5 header), but I now wonder if that assumption is correct? For example, it looks like the request and response body are the bare Terraform state data, so without any custom GitLap struct holding the state (and maybe the MD5 hash) in their own struct fields.

Also, it looks like (without having all context, so I could be wrong) you are trying to put (IMO) too much of the required logic in this package instead of in the Terraform backend. In my opinion the only code in this package should be related to making the actually calls to the GitLap API's. All other stuff (e.g. the logic to handle any errors in a custom way or creating a custom response like the `LockData` struct) are probably better placed within the context of the Terraform backend instead of the API client.

Specifically for the `LockData` struct, it could actually make sense to put it in this package if the API call to lock the state would return such an object. I don't know if that can still be added to the API design? You have all the info in GitLap, so why not respond to a lock call with the appropriate data versus manually creating that object after the call returned, based on the return codes of the call?

And of course you could also do some additional error checking in here, but that would mean that everyone who would want to (re)use these calls for a project of their own, are bound to the responses that work well for a Terraform state backend, but maybe not for their specific use.

I've made some changes to show what I think the code could look like. I understand that this will not work yet, but it shows what could be a nice separation of concerns by offloading all the Terraform backed specific logic and knowledge to the backend package and only keep the API calling logic in here. Of course it can also serve as a starting point for discussing the options.

I made quite a few assumptions and suggestion in the solution, but even if those don't hold up I still believe this sort of layout should work with just a few minor adjustments (for example if the body remains the base Terraform state). I've put in some examples of how the additional things (lock ID query param and MD5 header) can be added without requiring to change a lot of the core logic for just these 2 or 3 API calls that need to alter the request just a bit before sending it.

I pushed my changes to this branch (https://github.com/xanzy/go-gitlab/tree/svh/f-tfstate) to not bother you too much on your own branch 😉 

Last but not least, I'm also up for a face-2-face as a means to discuss things as that could maybe speed up things. Curious what you thoughts are! Cheers!

ash2k:
>  I think it would be extremely helpful for me to better understand what the API is going to look like. Is there maybe a (preliminary) design that you can share?
>
> I assumed that the new API endpoints would (more or less) follow the same patterns and style of the existing API endpoints. Maybe with a few little tweaks or additions (like adding the additional MD5 header), but I now wonder if that assumption is correct? For example, it looks like the request and response body are the bare Terraform state data, so without any custom GitLap struct holding the state (and maybe the MD5 hash) in their own struct fields.

Very good questions! I'm sorry, I should have provided the context earlier.

We are building a state storage for Terraform that is compatible with the existing Terraform's [http backend](https://github.com/hashicorp/terraform/tree/master/backend/remote-state/http). So GitLab's API is designed to match what that backend code expects to talk to.

This PR mimicks that http backend so that this library can be used to build a new native GitLab backend for Terraform still talking to the same GitLab's API endpoints. That is why the API looks slightly different.

> Also, it looks like (without having all context, so I could be wrong) you are trying to put (IMO) too much of the required logic in this package instead of in the Terraform backend. In my opinion the only code in this package should be related to making the actually calls to the GitLap API's. All other stuff (e.g. the logic to handle any errors in a custom way or creating a custom response like the LockData struct) are probably better placed within the context of the Terraform backend instead of the API client.

I'm open to discussing this, however I believe that the value that this library provides quickly vanishes if the caller needs to know about the underlying transport details. The caller should be abstracted from parsing http response body or headers or anything like that. That is why I put the logic into the library and not into the backend. Here is a draft of how I'm planning to use the library https://github.com/hashicorp/terraform/compare/master...ash2k:ash2k-gitlab-backend

> Last but not least, I'm also up for a face-2-face as a means to discuss things as that could maybe speed up things. Curious what you thoughts are! Cheers!

Yeah, I think that'd be great! I'll send you an email and we can arrange that.

svanharmelen:
> We are building a state storage for Terraform that is compatible with the existing Terraform's http backend. So GitLab's API is designed to match what that backend code expects to talk to.

> This PR mimicks that http backend so that this library can be used to build a new native GitLab backend for Terraform still talking to the same GitLab's API endpoints. That is why the API looks slightly different.

OK... That's a pretty unexpected (to me at least) design choice, so yeah that does explain a few things 😏 Is there a specific reason why you choose to implement it like that instead of following any of the other backends (which all use an API client to call API's that follow their own specs)?

Also, if you are making your backend compatible with the HTTP backend then what is the added value of also creating a native client (which will then more or less be a copy of the HTTP backend)? Wouldn't it then be easier to just use the HTTP backend instead?

Not trying to judge (as that's impossible without context) any of your choices, but I'm just curious as it feels questionable to try and implement two different API specs into one API client. Which is actually what you are trying to do now, right? We have the GitLab API specs and now additionally the API specs that the HTTP backend requires.

I was fully under the impression that you were going to implement a native solution using new API endpoints that comply to the GitLap API specs. In that case, using this client to make those calls sounds like a very logical and good plan. But if instead you are trying to rework this whole client for the purpose of adding 5 new API endpoints that follow a different spec, then that maybe isn't the most logical thing to do 🤔

Guess we have something to think about this weekend and then we'll have a chat about this early next week to see how to move this forward...

svanharmelen:
Thought about it a bit more and one reason I can think of (for making your API compatible with the HTTP backend) could be to have instant support for almost all Terraform versions (as the HTTP backend was among the first backends that were added).

While I think that people using Terraform typically upgrade quite fast and often, I must admit that could be an interesting reason 😏

If time permits I want to have another look at an idea I had before we meet. Maybe we can add the required bit without too much changes...

svanharmelen:
Hey, this one shouldn't be closed. Sorry for that! I deleted the `svh/f-client-options` branch as it was merged, which automatically closed this PR. So closing this one was not intentional...

PR #813, opened 2020-03-30
Introduce client options

svanharmelen:


access_requests_test.go
```diff
@@ -11,7 +11,10 @@ import (
 )
 
 func TestListProjectAccessRequests(t *testing.T) {
-	mux, server, client := setup()
+	mux, server, client, err := setup()
```
ash2k: Wouldn't it be better to just pass `t` here and do a `t.Fatalf()` inside to unwind the stack without returning an error and changing the signature?

examples/basic_auth.go
```diff
@@ -8,7 +8,12 @@ import (
 
 // This example shows how to create a client with username and password.
 func basicAuthExample() {
-	git, err := gitlab.NewBasicAuthClient(nil, "https://gitlab.company.com", "svanharmelen", "password")
```
ash2k: Why not turn everything into options and have a single constructor function? :) Passing `nil` is not that nice and is begging to be an optional parameter. Same with auth - pick an option that works for you.

gitlab.go
```diff
@@ -335,6 +334,9 @@ type Client struct {
 	// should always be specified with a trailing slash.
 	baseURL *url.URL
 
+	// disableRetries is used to disable the default retry logic.
+	disableRetries bool
```
ash2k: It might be more future proof to represent retries as an injectable strategy object. I.e. define an interface for backoff that tells the caller if it needs to wait and for how long. It can then be used in a loop for retries.

Some things that may be useful in some way that I found:
https://github.com/kubernetes/client-go/blob/master/util/flowcontrol/backoff.go
https://github.com/cenkalti/backoff
Can just reuse the backoff from the client that is being used here https://github.com/hashicorp/go-retryablehttp/blob/master/client.go#L322

gitlab.go
```diff
@@ -505,8 +516,20 @@ func newClient(httpClient *http.Client) *Client {
 		RetryMax:     30,
 	}
 
-	// Set the default base URL.
-	_ = c.SetBaseURL(defaultBaseURL)
+	// Set the default base URL and apply any given client options.
+	for _, fn := range append([]ClientOptionFunc{WithBaseURL(defaultBaseURL)}, options...) {
+		if fn == nil {
```
ash2k: I'm wondering what is the reason to allow for this?

gitlab.go
```diff
@@ -505,8 +516,20 @@ func newClient(httpClient *http.Client) *Client {
...
+			continue
+		}
+		if err := fn(c); err != nil {
+			return nil, err
+		}
+	}
+
+	// Configure the rate limiter.
+	if err := c.configureLimiter(); err != nil {
```
ash2k: Perhaps make it an option too?

Also, I think longer term it'd be better to make the request to get the limits on first client use. I'd say it's quite unexpected to have I/O happen as part of client construction. [`sync.Once`](https://golang.org/pkg/sync/#Once) is a perfect fit for the job. Or even better - configure with no limiter initially and then after the first successful request initialize the limiter properly. On each subsequent request check that the limiter is initialized to the correct values - they may change during the lifetime of a client.

gitlab.go
```diff
@@ -716,8 +709,14 @@ func (c *Client) configureLimiter() error {
 	return nil
 }
 
+// BaseURL return a copy of the baseURL.
+func (c *Client) BaseURL() *url.URL {
```
ash2k: I wonder why not make it return a value, not a pointer? Same with the field in the `Client` struct - it could be just a value.

gitlab.go
```diff
@@ -993,10 +989,42 @@ func parseError(raw interface{}) string {
 	}
 }
 
-// OptionFunc can be passed to all API requests to make the API call as if you were
-// another user, provided your private token is from an administrator account.
-//
-// GitLab docs: https://docs.gitlab.com/ce/api/README.html#sudo
+// ClientOptionFunc can be used customize a new GitLab API client.
+type ClientOptionFunc func(*Client) error
```
ash2k: When extending and using the library I found it hard to find all the functions that match this signature. Would be nice to move them all to a separate file. And same with `OptionFunc`. Will also make this file smaller and easier to navigate.

gitlab.go
ash2k: Another thing to consider - currently an I/O error to get the limit values will prevent the client from being constructed and this is not great. The remote side may be unavailable but the client for it should be constructable. In my  experience applications are typically prepared to handle I/O errors when using clients but treat client construction errors as fatal.

I once was dealing with a DB client library that would throw an exception when a client was constructed while the DB was unavailable and it was a huge pain point. I had to build a wrapper around the library to handle this. On each wrapper client use it'd check if the real client was already constructed and construct one if not. Was a PITA.

access_requests_test.go
svanharmelen: I think it's better, but mainly because it saves a few lines of code. The signature will still change of course 😏 But will change this one...

examples/basic_auth.go
svanharmelen: I agree that putting the optional HTTP client into an option indeed makes sense, so I'll adjust that one. But I'll keep the dedicated new client functions to make it more clear what type of clients you can create.

gitlab.go
svanharmelen: I'm not sure if it's more future proof. With the newly introduced option functions it should be a backwards compatible change to add this anywhere in the future. For now I'm not going to change this one.

gitlab.go
svanharmelen: This is actually not needed anymore as it's solved in a different way now. So this one is removed 👍

gitlab.go
svanharmelen: Not sure if I agree, but sure 😉 Moved the option functions to `client_options.go` and `request_options.go`.

gitlab.go
svanharmelen: I partly agree... Especially on the point that creating the client should not fail with an I/O error, so I changed the behavior a little to make sure that will not happen.

I initially intended `configureLimiter()` to be able to fail silently. So the error can actually be ignored as then the limiter will be configured as unlimited (i.e. disabled). That got lost in the changes somehow, but it's now used as such again.

I recognize that this could maybe be improved in the future, but let's keep it like this for now and first get some feedback for a few months (the limiter logic is super new).

gitlab.go
svanharmelen: Looks like there actually was a reason for it: https://github.com/xanzy/go-gitlab/commit/b8b6e4b888142f936ba215f02fe7712c56e29de0

ash2k: I see. Go disallows calling pointer methods on return values that are not pointers. But assigning to an intermediate variable and then calling methods on it should work. Not a big deal, just though that since a copy is being returned anyway, why not avoid extra pointer.

svanharmelen: Yeah, but it's debatable if that makes things easier or not. In the stdlib they pass URL as a pointer everywhere, so I probably (it's 5 years ago) though it would make sense to be consistent with that...

gitlab.go
```diff
@@ -437,31 +439,50 @@ type ListOptions struct {
...
-	err := client.requestOAuthToken(context.TODO())
+	err = client.requestOAuthToken(context.Background())
 	if err != nil {
 		return nil, err
 	}
 
 	return client, nil
 }
 
+// NewOAuthClient returns a new GitLab API client. If a nil httpClient is
```
ash2k: There is no `httpClient` argument anymore :)

gitlab.go
```diff
@@ -437,31 +439,50 @@ type ListOptions struct {
...
+	if err != nil {
+		return nil, err
+	}
 	client.authType = privateToken
 	client.token = token
-	return client
+	return client, nil
 }
 
 // NewBasicAuthClient returns a new GitLab API client. If a nil httpClient is
```
ash2k: There is no `httpClient` argument anymore :)

gitlab.go
```diff
@@ -437,31 +439,50 @@ type ListOptions struct {
 // NewClient returns a new GitLab API client. If a nil httpClient is
```
ash2k: There is no `httpClient` argument anymore.

README.md
```diff
@@ -97,8 +97,21 @@ access different parts of the GitLab API. For example, to list all
 users:
 
 ```go
-git := gitlab.NewClient(nil, "yourtokengoeshere")
-//git.SetBaseURL("https://git.mydomain.com/api/v4")
+git, err := gitlab.NewClient(nil, "yourtokengoeshere")
```
ash2k: Here and in a few other places there are still examples of passing `nil` http client to this function.

gitlab.go
ash2k: Sorry, it's still here. Just reminding in case you wanted to remove the `if` but forgot.

gitlab.go
```diff
@@ -779,7 +792,6 @@ func (c *Client) NewRequest(method, path string, opt interface{}, options []Opti
 		if fn == nil {
```
ash2k: Is this check still needed?

gitlab.go
```diff
@@ -856,6 +868,12 @@ func (r *Response) populatePageValues() {
 // interface, the raw response body will be written to v, without attempting to
 // first decode it.
 func (c *Client) Do(req *retryablehttp.Request, v interface{}) (*Response, error) {
+	// If not yet configured, try to configure the rate limiter. Fail
+	// silently as the limiter will be disabled in case of an error.
+	if c.limiter == nil {
+		c.configureLimiter()
```
ash2k: This is a data race: the `Do()` method might be called concurrently from multiple goroutines and the `c.limiter` read and write operations are not synchronized properly. `sync.Once` might be a good way to perform the synchronization here.

svanharmelen:
Good findings, thanks 👍🏻 I was going to have a closer look tomorrow as this was done just a bit too fast without a proper self review, but thanks for having another look.

README.md
svanharmelen: Thanks, updated...

gitlab.go
svanharmelen: Updated the comment 👍

gitlab.go
svanharmelen: Updated the comment 👍

gitlab.go
svanharmelen: Updated the comment 👍

gitlab.go
svanharmelen: That `if` check needs to stay to make sure we don't cause a panic if someone actively sets `nil` as an `OptionFunc` param value.

See this small example which uses a simplified but similar call tree: https://play.golang.org/p/aT7OXvRt8eT

ash2k: Hm, ok. I'm wondering why this is not considered a bug? If you pass `nil` where a function is expected... normally code just panics when misused.

svanharmelen: Maybe it has to do with the fact it's a variadic argument. Not sure, but I do know the compiler accepts it so this needs to stay in here.

svanharmelen: Thinking about it a bit more, I guess the zero value of a function is actually `nil`. So this gives you the same issue: https://play.golang.org/p/PstTInuOrMt

gitlab.go
svanharmelen: Yep, good on... I didn't know if `sync.Once` also blocked any other calls until the one execution has finished, but looking at the implementation is indeed does that (by using a lock), so I'll use a `sync.Once` instead 👍

svanharmelen:
In between meetings I think I resolved all feedback. Still want to do a good self review after the workday ended, but I think it should be more or less good to go like this.

gitlab.go
svanharmelen: I got another request related to being able to customize the behavior, so I added `WithCustomRetry` and `WithCustomBackoff` option funcs to be able to provide your own respective policies... That, together with `WithoutRetries` which I already added before, should offer enough flexibility around this part for the time being...

svanharmelen:
For those who are looking at this PR, I think it should be good in it's current state. So I'm planning to merge this one EOD tomorrow. Let me know any concerns (if any) with moving this forward. Thanks!

mvisonneau:
nice one @svanharmelen, thanks a lot 👌 it seems to work like a charm!

PR #819, opened 2020-03-31
Add applications support

prytoegrian:
Fix #764 

Admin applications weren't supported yet so I added this feature.
One can test these new functionnalities with a [specific example](https://github.com/prytoegrian/go-gitlab/blob/pry/applications/examples/applications.go) and [a docker image](https://docs.gitlab.com/omnibus/docker/).

PR #827, opened 2020-04-02
Break cyclic dependency

prytoegrian:
Hi, and thanks for your tool !

During my work on #819, I noted a cyclic dependency: the client `gitlab.Client` contains services fields such as `AccessRequests     *AccessRequestsService` and in return `AccessRequestsService` contains the `gitlab.Client` field. In my sense this is a design flaw so I dug into this problem.
I also use this pretext to try to split the big one directory `go-gitlab` into semantic ones.

So, this PR proposes an alternative representation of all the responsibilities and try to make them unique. As an example, I built a `template` package to separate all the business `template` stuff, hoping it would give a good idea to readers.

## In a nutshell
I think we can precise the dependency relation between the HTTP client and a service by creating the client and then injecting it into the service :
```go
git := gitlab.NewClient(nil, "")
ci := templates.NewCITemplate(*git)

c, _, err := ci.ListAllTemplates(&templates.ListCIYMLTemplatesOptions{})
```
Cf. [examples/templates.go](https://github.com/prytoegrian/go-gitlab/blob/pry/tree/examples/templates.go)

## Tree view
In the future, we could imagine a tree view based on Gitlab API : 
```
- groups/
- merge_requests/
- projects/
- users/
- broadcast_messages.go
- ...
```

## Compatibility
Since the package is already consumed, it would be inappropriate to merge this PR as is and thus create compatibility breaks for consumers. I propose using the [go versioning system](https://blog.golang.org/using-go-modules#TOC_5.) for exploiting in parallel the current version and the future one ; theses changes would therefore be in a `go-gitlab/v2` directory.

In term, it will facilitate the creation of tests.

Regards.


svanharmelen:
Thank you for opening this PR @prytoegrian, but unfortunately I don't agree with your statement that the current design is flawed. In fact the API of this client was very intentionally designed like this. 

In my opinion it gives users of the client a very nice and ergonomic API to work with and to be honest, I don't really see what actual problem (i.e. what breaks or doesn't work for you?) you are trying to solve.

Again thank you for your efforts, but unless there is a real problem (e.g. a bug) that you are trying to solve, I will not merge this PR.

PR #828, opened 2020-04-03
Fix token field data race

ash2k:
This PR solves these issues:

1. `Do()` can be called concurrently from multiple goroutines. If that happens with basic auth then the method both reads and writes the `token` field concurrently without synchronization. It's a data race.
1. When token expires, multiple concurrent `Do()` invocations can start refreshing it. This is unnecessary. More IO/cpu resources used both client and server side. Chances to run into rate limiting because oAuth2 client is not using the same rate limiter.
1. It moves the request to acquire the initial token from the client construction function into `Do()` to ensure client construction does not depend on network/remote server availability. As discussed in my other PR, it's nicer for users when client construction is deterministic. This also consolidates all locking into a single method which makes it easier to understand IMO.

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -91,6 +91,10 @@ type Client struct {
 	// Token used to make authenticated API calls.
 	token string
 
+	// Protects the token field from concurrent read/write accesses. Only used with basic auth when token field
+	// access needs to be synchronized.
+	tokenLock sync.Mutex
+
 	// User agent used when communicating with the GitLab API.
 	UserAgent string
 
@@ -203,11 +207,6 @@ func NewBasicAuthClient(username, password string, options ...ClientOptionFunc) 
 	client.username = username
 	client.password = password
 
-	err = client.requestOAuthToken(context.Background())
-	if err != nil {
-		return nil, err
-	}
-
 	return client, nil
 }
 
@@ -223,7 +222,7 @@ func NewOAuthClient(token string, options ...ClientOptionFunc) (*Client, error) 
 	return client, nil
 }
 
-func (c *Client) requestOAuthToken(ctx context.Context) error {
+func (c *Client) requestOAuthToken(ctx context.Context) (string, error) {
 	config := &oauth2.Config{
 		Endpoint: oauth2.Endpoint{
 			AuthURL:  fmt.Sprintf("%s://%s/oauth/authorize", c.BaseURL().Scheme, c.BaseURL().Host),
@@ -233,10 +232,9 @@ func (c *Client) requestOAuthToken(ctx context.Context) error {
 	ctx = context.WithValue(ctx, oauth2.HTTPClient, c.client.HTTPClient)
 	t, err := config.PasswordCredentialsToken(ctx, c.username, c.password)
 	if err != nil {
-		return err
+		return "", err
 	}
-	c.token = t.AccessToken
-	return nil
+	return t.AccessToken, nil
 }
 
 func newClient(options ...ClientOptionFunc) (*Client, error) {
@@ -494,13 +492,6 @@ func (c *Client) NewRequest(method, path string, opt interface{}, options []Requ
 	reqHeaders := make(http.Header)
 	reqHeaders.Set("Accept", "application/json")
 
-	switch c.authType {
-	case basicAuth, oAuthToken:
-		reqHeaders.Set("Authorization", "Bearer "+c.token)
-	case privateToken:
-		reqHeaders.Set("PRIVATE-TOKEN", c.token)
-	}
-
 	if c.UserAgent != "" {
 		reqHeaders.Set("User-Agent", c.UserAgent)
 	}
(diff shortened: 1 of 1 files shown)


ash2k:
Rebased to fix the merge conflict.

ash2k:
I'm sorry, but the updated fix does not fix the bug. Only writes are guarded by the mutex now, not all accesses. From https://golang.org/doc/articles/race_detector.html:

> A data race occurs when two goroutines access the same variable concurrently and at least one of the accesses is a write.

This is exactly what is happening - multiple writes cannot happen concurrently (because they are ordered by the mutex) but reads are not ordered with each other or writes so one or more read can happen concurrently with a write.

svanharmelen:
While I don't really see the danger in reading an outdated token value in this particular use case, you are right that this doesn't protect against it properly yet. So I added a few extra lines here: https://github.com/xanzy/go-gitlab/commit/c080e7aa2ebb3c8ae708ae86f9a57b533bfb4b34

The main difference is using a RWLock to not fully block in the common case where a call only needs to get the current token. To make sure the initial case is also handled properly I always send a hard coded `""` as the token while no token was set yet. This makes it function more or less like a `once.Do` call.

So with this I think it should guard for reads as well and should fully protect against any kind of race conditions right? Thanks!

ash2k:
> While I don't really see the danger in reading an outdated token value in this particular use case, 

I think this page from Dmitry Vyukov explains it very nicely https://software.intel.com/content/www/us/en/develop/blogs/benign-data-races-what-could-possibly-go-wrong.html Just say "No" to data races :)

> The main difference is using a RWLock to not fully block in the common case where a call only needs to get the current token. To make sure the initial case is also handled properly I always send a hard coded "" as the token while no token was set yet. This makes it function more or less like a once.Do call.

This line still reads the token without a lock. https://github.com/xanzy/go-gitlab/blob/master/gitlab.go#L624

I'm wondering what was wrong with my patch?

svanharmelen:
> I think this page from Dmitry Vyukov explains it very nicely https://software.intel.com/content/www/us/en/develop/blogs/benign-data-races-what-could-possibly-go-wrong.html Just say "No" to data races :)

OK, fair enough...

> This line still reads the token without a lock. https://github.com/xanzy/go-gitlab/blob/master/gitlab.go#L624

I know, but since we don't actually care _what_ value is set here only _if_ a value was set I can't really see how this could cause any harm. But in light of "just say no to data races", I'll update the code a bit more to fix this one as well 😏 

> I'm wondering what was wrong with my patch?

Nothing really, but I didn't really like having the two inline functions inside the `Do` method and the full lock on each and every call API call.

I just made another small update (https://github.com/xanzy/go-gitlab/commit/5207be848ffd9342e72ad9f97e5afa7b6b9acc7a) with now guards around all reads and writes and is actually pretty much inline with you earlier solution. Just without having the inline functions and with a RWLock.

Thanks for sticking with me and keep giving feedback 👍

ash2k:
> Thanks for sticking with me and keep giving feedback 👍

I hope I'm not annoying you too much :) Just trying to help to improve the library because I care.

> But in light of "just say no to data races", I'll update the code a bit more to fix this one as well 😏

Apart from just being correct this is also cool because if someone runs a unit test with this library and the race detector turned on they will not get warnings coming from the library code. 👍 

> Nothing really, but I didn't really like having the two inline functions inside the Do method and the full lock on each and every call API call.

Fair enough, it didn't look pretty. However performance-wise the code was just doing a [lock, read, compare, unlock](https://github.com/xanzy/go-gitlab/pull/828/commits/10f2827d8282bbc5dd1f05992c4c15ef8ddd41f9#diff-3d26bb4c03811c684c3ebb6529e0e5c0R617-R626) almost always ("token is set" case). I'm pretty sure this method is I/O bound 100% of time, so I wouldn't worry about locking performance at all. We should not use [`atomic.Value`](https://golang.org/pkg/sync/atomic/#Value) here because a concurrent request may then send a non-empty token that has expired already which we may be refreshing in the current request. We do want this mutex to block in this case to prevent sending a request with an expired token.

The code may look better if those two inline functions are promoted to methods with descriptive names like `getNonEmptyToken()` for the [first one](https://github.com/xanzy/go-gitlab/pull/828/commits/10f2827d8282bbc5dd1f05992c4c15ef8ddd41f9#diff-3d26bb4c03811c684c3ebb6529e0e5c0R616-R627) and `maybeRefreshToken()` for [the second](https://github.com/xanzy/go-gitlab/pull/828/commits/10f2827d8282bbc5dd1f05992c4c15ef8ddd41f9#diff-3d26bb4c03811c684c3ebb6529e0e5c0R645-R659). What do you think?

Looking at https://github.com/xanzy/go-gitlab/commit/5207be848ffd9342e72ad9f97e5afa7b6b9acc7a I see that `Do()` can call `authenticateAndDo()` which calls `Do()` holding the lock. This is problematic for two reasons:

- Holding an exclusive lock means no concurrency is possible at all, even the `RLock()` will block.
- Mutexes in Go are not re-entrant so this code simply does not work. This illustrates the problem - it panics because of the deadlock (no runnable goroutines) https://play.golang.org/p/b8VARwNjKtL

svanharmelen:
> I hope I'm not annoying you too much :)

Nope, you aren't I'm mostly getting frustrated with myself trying to fix this too quickly in between other stuff causing me to miss such obvious issues as you just pointed out.

I'll have another (less quick 😏) look and will make a proper PR this time (which I normally always do, but didn't make time for this time).

svanharmelen:
OK, found a bit of time to review it and reverted it to a solution I had earlier today (just before I thought to make it a bit cleaner by putting the next call to `Do` in a separate method 🙈 #862 

Think this solves it, but would appreciate a double check to make sure. Thanks...

svanharmelen:
Thanks again for your feedback @ash2k! Much appreciated!

PR #832, opened 2020-04-15
Update EditProjectOptions struct

cakoolen:
Since the last change to this struct, quite some options have been added and some have been deprecated. 

I moved the deprecated fields to the end of the struct with a comment to notify the user that these fields are deprecated. I added the new fields in a separate block such that the diff is more easy to read. If you want, I could rewrite this to use the same order as the fields in the gitlab documentation.


cakoolen:
Is there something I can do to at least get the new fields in? I could for example remove the deprecation of the fields that are deprecated according to the documentation. Un-deprecating them, will at least make the code backwards compatible.

svanharmelen:
Sorry for not responding earlier on this one! I will get back and see that we get this one sorted coming week!

cakoolen:
Thanks for merging the changes. I would have been happy to help you out with the additional changes you did that were not in my original PR. 

About the time it took, there are only so many things one can do in a week. I'm just happy it made it into the repo so we can continue using the upstream repo.

PR #834, opened 2020-04-19
Fixed missing d in pipeline api param

sluongng:
UpdatedBefore in pipeline api was not url encoded correctly and missed a
`d`. This could be very bad as there is no error from Gitlab to prompt
user that this is an invalid param.


sluongng:
> Good catch! Thanks @sluongng!

Production data was affected because of this 😢

svanharmelen:
Ai, sorry to hear that!

PR #851, opened 2020-05-10
Add last_commit_id field to create commit action

terev:
This field seems to be missing from the struct. Adding it should be safe. https://docs.gitlab.com/ee/api/commits.html#create-a-commit-with-multiple-files-and-actions


commits.go
```diff
@@ -124,6 +124,7 @@ type CommitAction struct {
 	PreviousPath string     `url:"previous_path,omitempty" json:"previous_path,omitempty"`
 	Content      string     `url:"content,omitempty" json:"content,omitempty"`
 	Encoding     string     `url:"encoding,omitempty" json:"encoding,omitempty"`
+	LastCommitID string     `url:"last_commit_id,omitempty" json:"last_commit_id,omitempty"`
```
svanharmelen, line 127: Maybe add `execute_filemode` as well to make the list complete?

PR #863, opened 2020-05-27
Improve connection handling

ash2k:
Two optimizations:

- drain connections before closing to ensure they can be reused. See docs for the `Body` field https://golang.org/pkg/net/http/#Response.

- Eagerly close connection before requesting a refreshed auth token so that the same connection can be potentially used for that.

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -626,15 +626,16 @@ func (c *Client) Do(req *retryablehttp.Request, v interface{}) (*Response, error
 	if err != nil {
 		return nil, err
 	}
-	defer resp.Body.Close()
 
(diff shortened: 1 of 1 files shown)


mhutter:
I believe [Go already does this for you](https://github.com/golang/go/blob/go1.14.6/src/net/http/transfer.go#L1008-L1010)

ash2k:
It's weird that this is not documented on the `Response` type. Updated the code, thanks.

mhutter:
True... I only discovered it because I've never seen the "drain before close" and it made me curious

svanharmelen (approved):
Thanks @ash2k! LGTM!

PR #866, opened 2020-06-03
add example to fetch language metrics

falcucci:


svanharmelen:
@falcucci thanks for this PR, but to be honest I don't think this is a good example to add. It doesn't show any new or specific uses of the package that are not already captured by other examples and it also looks there is quite some logic that isn't related to this package.

Again thanks for you efforts, but I prefer not to add this one. Sorry!

PR #868, opened 2020-06-05
Added RemoveLabels and AddLabels fields to UpdateIssueOptions struct

skast96:
I have added AddLabels and RemoveLabels as fields to the UpdateIssueOptions struct as this is supported by the gitlab REST API.

### DIFF
--- a/issues.go
+++ b/issues.go
@@ -401,8 +401,8 @@ type UpdateIssueOptions struct {
 	AssigneeIDs      []int      `url:"assignee_ids,omitempty" json:"assignee_ids,omitempty"`
 	MilestoneID      *int       `url:"milestone_id,omitempty" json:"milestone_id,omitempty"`
 	Labels           *Labels    `url:"labels,comma,omitempty" json:"labels,omitempty"`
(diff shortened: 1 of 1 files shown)


issues.go
```diff
@@ -401,6 +401,8 @@ type UpdateIssueOptions struct {
 	AssigneeIDs      []int      `url:"assignee_ids,omitempty" json:"assignee_ids,omitempty"`
 	MilestoneID      *int       `url:"milestone_id,omitempty" json:"milestone_id,omitempty"`
 	Labels           *Labels    `url:"labels,comma,omitempty" json:"labels,omitempty"`
+	AddLabels        Labels     `url:"labels,comma,omitempty" json:"add_labels,omitempty"`
+	RemoveLabels     Labels     `url:"labels,comma,omitempty" json:"remove_labels,omitempty"`
```
svanharmelen, line 405: These fields should be pointers and the tags are wrong...

DavidMBusey: @svanharmelen I'll make the edits.

svanharmelen (approved):

issues.go
svanharmelen: Thanks @DavidMBusey! But I just fixed it so we can merge this one 👍

PR #869, opened 2020-06-08
Add support for 'external' attribute in list-users

maxenced:
Gitlab provides support for searching external users (different than the external auth), as per : 

https://docs.gitlab.com/ee/api/users.html#for-admins

`You can search for users who are external with: /users?external=true `

This PR simply adds the `external` attribute in the ListUsers options


maxenced:
Sorry, missed it, ran fmt + rebased so we get one clean commit

PR #872, opened 2020-06-19
[license] Update addon codes

leominov:
AddOns looks like `map[string]int`, but I'm not sure, so I've adds new codes in backward compatibility maneer.

### DIFF
--- a/license.go
+++ b/license.go
@@ -48,9 +48,14 @@ type License struct {
 		Company string `json:"Company"`
 		Email   string `json:"Email"`
 	} `json:"licensee"`
+	// Add on codes that may occur in legacy licenses that don't have a plan yet.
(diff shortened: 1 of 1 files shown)


license.go
```diff
@@ -48,9 +48,14 @@ type License struct {
 		Company string `json:"Company"`
 		Email   string `json:"Email"`
 	} `json:"licensee"`
+	// Add on codes that may occur in legacy licenses that don't have a plan yet.
+	// https://gitlab.com/gitlab-org/gitlab/-/blob/master/ee/app/models/license.rb
```
svanharmelen, line 52: Think we can leave out the comments...

leominov: First line is from original code, second - one and only source of addon codes. Are you sure? Is it some overhead with it or something else?

svanharmelen: Ok... In general I try to prevent these kind of inline comments, bit it seems this one makes good sense 👍🏻

PR #879, opened 2020-07-01
Remote mirrors

lukemgriffith:
Adding remote mirrors service, as per the remote mirrors documentation on gitlab.

https://docs.gitlab.com/ce/api/remote_mirrors.html

### DIFF
--- a/project_mirror.go
+++ b/project_mirror.go
@@ -24,27 +24,25 @@ import (
 // ProjectMirrorService handles communication with the project mirror
 // related methods of the GitLab API.
 //
-// GitLAb API docs:
-// https://docs.gitlab.com/ce/api/remote_mirrors.html
+// GitLAb API docs: https://docs.gitlab.com/ce/api/remote_mirrors.html
 type ProjectMirrorService struct {
 	client *Client
 }
 
 // ProjectMirror represents a project mirror configuration.
 //
-// GitLAb API docs:
-// https://docs.gitlab.com/ce/api/remote_mirrors.html
+// GitLAb API docs: https://docs.gitlab.com/ce/api/remote_mirrors.html
 type ProjectMirror struct {
-	Enabled                *bool      `url:"enabled" json:"enabled"`
-	ID                     *int       `url:"id" json:"id"`
-	LastError              *string    `url:"last_error" json:"last_error"`
-	LastSuccessfulUpdateAt *time.Time `url:"last_successful_update_at" json:"last_successful_update_at"`
-	LastUpdateAt           *time.Time `url:"last_update_at" json:"last_update_at"`
-	LastUpdateStartedAt    *time.Time `url:"last_update_started_at" json:"last_update_started_at"`
-	OnlyProtectedBranches  *bool      `url:"only_protected_branches" json:"only_protected_branches"`
-	KeepDivergentRefs      *bool      `url:"keep_divergent_refs" json:"keep_divergent_refs"`
-	UpdateStatus           *string    `url:"update_status" json:"update_status"`
-	URL                    *string    `url:"url" json:"url"`
+	Enabled                bool       `json:"enabled"`
+	ID                     int        `json:"id"`
+	LastError              string     `json:"last_error"`
+	LastSuccessfulUpdateAt *time.Time `json:"last_successful_update_at"`
+	LastUpdateAt           *time.Time `json:"last_update_at"`
+	LastUpdateStartedAt    *time.Time `json:"last_update_started_at"`
+	OnlyProtectedBranches  bool       `json:"only_protected_branches"`
+	KeepDivergentRefs      bool       `json:"keep_divergent_refs"`
+	UpdateStatus           string     `json:"update_status"`
+	URL                    string     `json:"url"`
 }
 
 // ListProjectMirror gets a list of mirrors configured on the project.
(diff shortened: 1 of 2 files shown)


project_mirror.go
```diff
@@ -0,0 +1,162 @@
...
+
+package gitlab
+
+import (
+	"fmt"
+	"time"
+)
+
+// ProjectMirrorService handles communication with the project mirror
+// related methods of the GitLab API.
```
svanharmelen, line 25:
```suggestion
// related methods of the GitLab API.
//
// GitLAb API docs:
// https://docs.gitlab.com/ce/api/remote_mirrors.html
```

project_mirror.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// ProjectMirror represents a project mirror configuration.
+//
+// GitLAb API docs:
+// https://docs.gitlab.com/ce/api/remote_mirrors.html
+type ProjectMirror struct {
+	Enabled                bool        `json:"enabled"`
+	ID                     int         `json:"id"`
+	LastError              interface{} `json:"last_error"`
```
svanharmelen, line 37: Why is this an `interface{}`? This should be a specific type.

project_mirror.go
```diff
@@ -0,0 +1,162 @@
...
+	LastSuccessfulUpdateAt time.Time   `json:"last_successful_update_at"`
+	LastUpdateAt           time.Time   `json:"last_update_at"`
+	LastUpdateStartedAt    time.Time   `json:"last_update_started_at"`
```
svanharmelen, line 40: Are all these times always returned by the API?

project_mirror.go
```diff
@@ -0,0 +1,162 @@
...
+	OnlyProtectedBranches  bool        `json:"only_protected_branches"`
+	KeepDivergentRefs      bool        `json:"keep_divergent_refs"`
+	UpdateStatus           string      `json:"update_status"`
+	URL                    string      `json:"url"`
+}
+
+// ListProjectMirror gets a list of mirrors configured on the project.
+// these are copies of the repositories to external version control repositories
```
svanharmelen, line 48:
```suggestion
```

project_mirror.go
```diff
@@ -0,0 +1,162 @@
...
+	u := fmt.Sprintf("projects/%s/remote_mirrors", pathEscape(project))
+
+	req, err := s.client.NewRequest("GET", u, nil, options)
+
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var pm []*ProjectMirror
+
```
svanharmelen, line 66:
```suggestion
```

project_mirror.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// AddProjectMirrorOptions contains the properties requires to create
+// a new project mirror.
+type AddProjectMirrorOptions struct {
+	URL                   string `json:"url"`
+	Enabled               bool   `json:"enabled"`
+	OnlyProtectedBranches bool   `json:"only_protected_branches"`
+	KeepDivergentRefs     bool   `json:"keep_divergent_refs"`
```
svanharmelen, line 82: These all need to be pointers and also have `url` struct tags.

project_mirror.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return pm, resp, err
+
+}
+
+// AddProjectMirrorOptions contains the properties requires to create
+// a new project mirror.
```
svanharmelen, line 77:
```suggestion
// a new project mirror.
//
// GitLab API docs:
// https://docs.gitlab.com/ce/api/remote_mirrors.html#create-a-remote-mirror
```

project_mirror.go
```diff
@@ -0,0 +1,162 @@
...
+// https://docs.gitlab.com/ce/api/remote_mirrors.html#list-a-projects-remote-mirrors
+func (s *ProjectMirrorService) ListProjectMirror(pid interface{}, options ...RequestOptionFunc) ([]*ProjectMirror, *Response, error) {
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/remote_mirrors", pathEscape(project))
+
+	req, err := s.client.NewRequest("GET", u, nil, options)
+
```
svanharmelen, line 60:
```suggestion
```

project_mirror.go
```diff
@@ -0,0 +1,162 @@
...
+	resp, err := s.client.Do(req, pm)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return pm, resp, err
+}
+
+// EditProjectMirrorOptions contains the properties requires to edit
+// an existing project mirror.
```
svanharmelen, line 111:
```suggestion
// an existing project mirror.
//
// GitLab API docs:
// https://docs.gitlab.com/ce/api/remote_mirrors.html#update-a-remote-mirrors-attributes
```

project_mirror.go
```diff
@@ -0,0 +1,162 @@
...
+type EditProjectMirrorOptions struct {
+	MirrorID              int    `json:"mirror_id"`
+	URL                   string `json:"url"`
+	Enabled               bool   `json:"enabled"`
+	OnlyProtectedBranches bool   `json:"only_protected_branches"`
+	KeepDivergentRefs     bool   `json:"keep_divergent_refs"`
```
svanharmelen, line 117: These all need to be pointers and also have url struct tags.

project_mirror.go
```diff
@@ -0,0 +1,162 @@
...
+		return nil, resp, err
+	}
+
+	return pm, resp, err
+}
+
+// EditProjectMirrorOptions contains the properties requires to edit
+// an existing project mirror.
+type EditProjectMirrorOptions struct {
+	MirrorID              int    `json:"mirror_id"`
```
svanharmelen, line 113: This one should not be here as it's already in the function signature.

project_mirror.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	u := fmt.Sprintf("projects/%s/members/%d", pathEscape(project), user)
+
+	req, err := s.client.NewRequest("DELETE", u, nil, options)
+	if err != nil {
+		return nil, err
+	}
+
+	return s.client.Do(req, nil)
+}
```
svanharmelen, line 162: This seems to be moved from the project members service, but I don't think we should do that. Or was this meant as a copy/paste/adjust which went bad?

project_members.go
```diff
@@ -188,22 +190,3 @@ func (s *ProjectMembersService) EditProjectMember(pid interface{}, user int, opt
...
-	}
-	u := fmt.Sprintf("projects/%s/members/%d", pathEscape(project), user)
-
-	req, err := s.client.NewRequest("DELETE", u, nil, options)
-	if err != nil {
-		return nil, err
-	}
-
-	return s.client.Do(req, nil)
-}
```
svanharmelen, line 209: I don't think this should be removed, right?

project_members.go
```diff
@@ -14,6 +14,8 @@
 // limitations under the License.
 //
 
+// N.B Documentation does not support remove functionality, alternative is to disable an existing mirror.
```
svanharmelen, line 17: I would not add this comment here. If needed it should be placed in the comment of the related method.

project_mirror.go
lukemgriffith: Documentation doesn't state it, but testing myself looks like these properties are always returned.

lukemgriffith:
Fixed up where requested, there isn't any delete action on the api spec, so I've not included a delete method.

lukemgriffith:
Excellent, thanks.

PR #886, opened 2020-07-10
Add function for sharing groups with groups

DonEmil:
As far as I could tell, sharing groups with groups is not implemented: https://docs.gitlab.com/ce/api/groups.html#share-groups-with-groups

This PR is a suggestion to add it, along with an example of how that could look.

The type for "Share-Groups-With-Groups"-options seems identical to "ShareWithGroupOptions", so that should be able to be reused.

I did not see any tests for group_members.go

### DIFF
--- a/group_members.go
+++ b/group_members.go
@@ -176,6 +176,31 @@ func (s *GroupMembersService) AddGroupMember(gid interface{}, opt *AddGroupMembe
 	return gm, resp, err
 }
 
+// ShareWithGroup shares a group with the group.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/groups.html#share-groups-with-groups
+func (s *GroupMembersService) ShareWithGroup(gid interface{}, opt *ShareWithGroupOptions, options ...RequestOptionFunc) (*Group, *Response, error) {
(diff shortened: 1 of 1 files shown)


group_members.go
```diff
@@ -176,6 +176,31 @@ func (s *GroupMembersService) AddGroupMember(gid interface{}, opt *AddGroupMembe
 	return gm, resp, err
 }
 
+// ShareWithGroup shares a group with the group.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/groups.html#share-groups-with-groups
+func (s *GroupMembersService) ShareWithGroup(gid interface{}, opt *ShareWithGroupOptions, options ...RequestOptionFunc) (*Group, *Response, error) {
```
svanharmelen, line 183: Is `ShareWithGroupOptions` defined anywhere already? Or did you forgot to commit that?

DonEmil: ShareWithGroupOptions already exists for projects: https://github.com/xanzy/go-gitlab/blob/f2b3cfead604678820f486a362a1b264895589f3/projects.go#L791

It should be able to be reused, or duplicated for sharing groups with groups

PR #889, opened 2020-07-26
add list-group-runners

Starttoaster:
Regarding the API function for listing runners specific to a group: https://docs.gitlab.com/ee/api/runners.html#list-groups-runners


Starttoaster:
Here's a simple script in case you'd like to test it, but I verified that it works :smile: 

```
func main() {
	client, err := gitlab.NewClient(os.Getenv("GITLAB_TOKEN"))
	if err != nil {
		log.Println(err)
	}

        //Change these how you like
	groupID := 1234567890
	runnerType := "group_type"
	runnerStatus := "offline"
        runnerTags := []string{"medium"}

	runners, err := getGroupRunners(client, &groupID, &runnerType, &runnerStatus, runnerTags)
	if err != nil {
		log.Println(err)
	}

	for _, runner := range runners {
		log.Printf("%+v", runner)
	}
}

func getGroupRunners(client *gitlab.Client, gid *int, runnerType *string, runnerStatus *string, runnerTags []string) ([]*gitlab.Runner, error) {
	runners, _, err := client.Runners.ListGroupsRunners(*gid, &gitlab.ListGroupsRunnersOptions{
		Type:    runnerType,
		Status:  runnerStatus,
		TagList: runnerTags,
	})
	if err != nil {
		return nil, err
	}

	return runners, err
}
```

runners.go
```diff
@@ -414,3 +414,39 @@ func (s *RunnersService) VerifyRegisteredRunner(opt *VerifyRegisteredRunnerOptio
...
+	Type    *string  `url:"type,omitempty" json:"type,omitempty"`
+	Status  *string  `url:"status,omitempty" json:"status,omitempty"`
+	TagList []string `url:"tag_list,comma,omitempty" json:"tag_list,omitempty"`
+}
+
+// ListGroupsRunners List all runners (specific and shared) available in the group as well it’s ancestor groups.
+// Shared runners are listed if at least one shared runner is defined.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/runners.html#list-groups-runners
+func (s *RunnersService) ListGroupsRunners(gid interface{}, opt *ListGroupsRunnersOptions, options ...RequestOptionFunc) ([]*Runner, *Response, error) {
```
svanharmelen, line 433: The methods in this file are ordered using the order in which they are documented. So following that logic, this new code should be placed below the `DisableProjectRunner` method.

Starttoaster: Thanks!

Starttoaster:
Thanks for the review! Accepted your inline comment suggestions, and moved the new function underneath DisableProjectRunner().

Starttoaster:
@svanharmelen did you happen to get a chance to check this PR out after the suggestions you made were implemented?

Starttoaster:
👀 I made a mess out of a rebase. One sec

PR #890, opened 2020-07-26
Add Projects.ServiceDesk Api / AccessLevel to Project struct

ax1036:
Sample:
```
func createProject(client *gitlab.Client) {
	const (
		projectName        = "created_by_api"
		projectPath        = "sandbox/created_by_api"
		projectNamespaceID = 42
	)
	project, _, err := client.Projects.GetProject(projectPath, nil)
	if err == nil {
		log.Println(project)
		resp, err := client.Projects.DeleteProject(project.ID)
		if err != nil {
			log.Println(resp.Response)
			log.Fatalln(err)
		}
	}

	opt := &gitlab.CreateProjectOptions{
		Name:               gitlab.String(projectName),
		NamespaceID:        gitlab.Int(projectNamespaceID),
		ServiceDeskEnabled: gitlab.Bool(false),
	}
	project, resp, err := client.Projects.CreateProject(opt)
	if err != nil {
		log.Println(project)
		log.Println(resp.Response)
		log.Fatalln(err)
	}
}
```


ax1036:
Add missing access level to project struc.
```
IssuesAccessLevel       
RepositoryAccessLevel   
MergeRequestsAccessLevel
ForkingAccessLevel      
WikiAccessLevel         
BuildsAccessLevel       
SnippetsAccessLevel     
PagesAccessLevel        
```

(tell me if need separated PR)

projects.go
```diff
@@ -38,57 +38,67 @@ type ProjectsService struct {
...
+	ServiceDeskEnabled                        bool                `json:"service_desk_enabled"`
+	ServiceDeskAddress                        string              `json:"service_desk_address"`
+	IssuesAccessLevel                         *AccessControlValue `json:"issues_access_level"`
+	RepositoryAccessLevel                     *AccessControlValue `json:"repository_access_level"`
+	MergeRequestsAccessLevel                  *AccessControlValue `json:"merge_requests_access_level"`
+	ForkingAccessLevel                        *AccessControlValue `json:"forking_access_level"`
+	WikiAccessLevel                           *AccessControlValue `json:"wiki_access_level"`
+	BuildsAccessLevel                         *AccessControlValue `json:"builds_access_level"`
+	SnippetsAccessLevel                       *AccessControlValue `json:"snippets_access_level"`
+	PagesAccessLevel                          *AccessControlValue `json:"pages_access_level"`
```
svanharmelen, line 101: I don't see all these access level fields in the docs, are you sure they exist? Can you point me to a location that confirms that?

ax1036: https://gitlab.com/api/v4/projects/19184008 / https://gitlab.com/gitlab_ap_test/root_project

image

request access to project if you will. (switched to public)

svanharmelen: Check, thanks for pointing me to this. So then these should be good, but it shouldn't be pointers. If you can update that, we should be good I guess.

ax1036: all `*AccessControlValue` changed to `AccessControlValue` :+1:

PR #892, opened 2020-07-28
Adds GetUserByUsername method

veverkap:
This is "hidden" feature (see https://gitlab.com/gitlab-org/gitlab-foss/-/issues/43843)


svanharmelen:
Thanks @veverkap, but this logic is already in this client. If you look at the `ListUsers` method, you can see that one of the options allows you to set the username: https://github.com/xanzy/go-gitlab/blob/master/users.go#L108

So can we close this PR? Or is there something I missed that requires this PR?

veverkap:
Yeah, you're right.  I saw the note regarding admins and thought it was a different endpoint.

Turns out that the ?username= trick will work for non-admins some.

PR #893, opened 2020-07-30
Add instance cluster API support

sfang97:
[Instance cluster API support was added to GitLab ](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/36001) in 13.2, so this PR adds support for it. This is similar to the project/group cluster integration, but unlike the project/group implementation, there is no instance ID value since any instance cluster applies to the entire GitLab instance.

[Instance cluster API documentation](https://docs.gitlab.com/ee/api/instance_clusters.html#add-existing-instance-cluster)


instance_clusters.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	return ics, resp, err
+}
+
+// GetCluster gets an instance cluster.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/instance_clusters.html#get-a-single-instance-cluster
+func (s *InstanceClustersService) GetCluster(clusterID int, options ...RequestOptionFunc) (*InstanceCluster, *Response, error) {
```
svanharmelen, line 88: For consistency, I prefer to call this param `cluster` instead of `clusterID` (while I agree this would probably be a good, or maybe even better, name as well).

instance_clusters.go
```diff
@@ -0,0 +1,206 @@
...
+	Domain              *string                               `url:"domain,omitempty" json:"domain,omitempty"`
+	Enabled             *bool                                 `url:"enabled,omitempty" json:"enabled,omitempty"`
+	Managed             *bool                                 `url:"managed,omitempty" json:"managed,omitempty"`
+	EnvironmentScope    *string                               `url:"environment_scope,omitempty" json:"environment_scope,omitempty"`
+	PlatformKubernetes  *AddInstancePlatformKubernetesOptions `url:"platform_kubernetes_attributes,omitempty" json:"platform_kubernetes_attributes,omitempty"`
+	ManagementProjectID *string                               `url:"management_project_id,omitempty" json:"management_project_id,omitempty"`
+}
+
+// AddInstancePlatformKubernetesOptions represents the available PlatformKubernetes options for adding.
+type AddInstancePlatformKubernetesOptions struct {
```
svanharmelen, line 120:
```suggestion
type PlatformKubernetesOptions struct {
```

instance_clusters.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	return ic, resp, err
+}
+
+// AddInstanceClusterOptions represents the available AddCluster() options.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/instance_clusters.html#add-existing-cluster-to-instance
+type AddInstanceClusterOptions struct {
```
svanharmelen, line 109:
```suggestion
type AddClusterOptions struct {
```

instance_clusters.go
```diff
@@ -0,0 +1,206 @@
...
+	PlatformKubernetes *PlatformKubernetes `json:"platform_kubernetes"`
+	ManagementProject  *ManagementProject  `json:"management_project"`
+}
+
+func (v InstanceCluster) String() string {
+	return Stringify(v)
+}
+
+// PlatformKubernetes represents a GitLab Instance Cluster PlatformKubernetes.
+type InstanceClusterPlatformKubernetes struct {
```
svanharmelen, line 55:
```suggestion
type PlatformKubernetes struct {
```

instance_clusters.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	return ic, resp, err
+}
+
+// EditInstanceClusterOptions represents the available EditCluster() options.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/instance_clusters.html#edit-instance-cluster
+type EditInstanceClusterOptions struct {
```
svanharmelen, line 153:
```suggestion
type EditClusterOptions struct {
```

instance_clusters.go
```diff
@@ -0,0 +1,206 @@
...
+	Name                *string                                `url:"name,omitempty" json:"name,omitempty"`
+	Domain              *string                                `url:"domain,omitempty" json:"domain,omitempty"`
+	EnvironmentScope    *string                                `url:"environment_scope,omitempty" json:"environment_scope,omitempty"`
+	ManagementProjectID *string                                `url:"management_project_id,omitempty" json:"management_project_id,omitempty"`
+	PlatformKubernetes  *EditInstancePlatformKubernetesOptions `url:"platform_kubernetes_attributes,omitempty" json:"platform_kubernetes_attributes,omitempty"`
```
svanharmelen, line 158:
```suggestion
	PlatformKubernetes  *PlatformKubernetesOptions `url:"platform_kubernetes_attributes,omitempty" json:"platform_kubernetes_attributes,omitempty"`
```

instance_clusters.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// EditInstancePlatformKubernetesOptions represents the available PlatformKubernetes options for editing.
+type EditInstancePlatformKubernetesOptions struct {
+	APIURL    *string `url:"api_url,omitempty" json:"api_url,omitempty"`
+	Token     *string `url:"token,omitempty" json:"token,omitempty"`
+	CaCert    *string `url:"ca_cert,omitempty" json:"ca_cert,omitempty"`
+	Namespace *string `url:"namespace,omitempty" json:"namespace,omitempty"`
+}
```
svanharmelen, line 167: I know and understand that the `AuthorizationType` cannot be set here, but I prefer to reuse the `PlatformKubernetesOptions` here and let the API return an error if you so try to update that particular field.

instance_clusters.go
```diff
@@ -0,0 +1,206 @@
...
+
+// EditCluster updates an existing instance cluster.
+// Note: name, api_url, ca_cert and token can only be updated if the cluster
+// was added through the 'Add existing Kubernetes' cluster option or through the
+// 'Add existing instance' cluster endpoint.
```
svanharmelen, line 172:
```suggestion
```

As we already link to the docs, it's a bit redundant to also add this note here. Not an issue, but I try to point to the docs as much as possible as things change pretty often in the Gitlab API.

instance_clusters.go
sfang97: Hi @svanharmelen , thanks for the suggestion. I tried applying it but the type `PlatformKubernetes` is already declared in `https://github.com/xanzy/go-gitlab/blob/master/project_clusters.go:56:6: PlatformKubernetes redeclared in this block`

instance_clusters.go
sfang97: Same thing here, `EditClusterOptions` is already declared in `https://github.com/xanzy/go-gitlab/blob/d3946d4f8c62bfae945062be289f13ea5e6ecfe4/project_clusters.go#L177`

instance_clusters.go
sfang97: `AddClusterOptions` is already declared in `https://github.com/xanzy/go-gitlab/blob/d3946d4f8c62bfae945062be289f13ea5e6ecfe4/project_clusters.go#L129`

instance_clusters.go
sfang97: I think `AddPlatformKubernetesOptions` would be more explicit here, since these are the options for adding a cluster. But then `AddPlatformKubernetesOptions` is already declared in https://github.com/xanzy/go-gitlab/blob/d3946d4f8c62bfae945062be289f13ea5e6ecfe4/project_clusters.go#L140 so the compiler throws a redeclaration error

instance_clusters.go
sfang97: Makes sense, I'll make the change

instance_clusters.go
sfang97: Ah, I just saw your note on reusing `PlatformKubernetesOptions` for both Add and Edit cluster. In that case, we don't name this `AddPlatformKubernetesOptions` and there is not a redeclaration. Making the change!

sfang97:
@svanharmelen Thanks for the review! I applied the non-conflicting suggestions. As for the others, the compiler complains that the new name is `redeclared in this block` so I didn't make those changes. Take another look and lmk if it's ready to merge! :)

instance_clusters.go
svanharmelen: So why not just omit the struct here and use the existing one?

instance_clusters.go
svanharmelen: Same for this one, if the struct already exists, why not reuse it here versus declaring the same struct twice?

instance_clusters.go
sfang97: Gotcha I see what you mean, pushed these changes!

sfang97:
@svanharmelen Thanks for the review, I removed the unused structs. Ready to merge? 🚀

PR #894, opened 2020-07-31
Support SharedWithGroups in groups

### LINKED ISSUE #888: Add support for new share group with group feature : `/groups/:id/share`
MR https://gitlab.com/gitlab-org/gitlab/-/merge_requests/32008 has been deployed on gitlab.com and should be available for on-prem soon.

This MR add support for the group sharing with group feature, available through WebUi since a few months.

It would be really nice to get it supported. Does not look like much of work is needed.

horjulf:
Fixes #888, the sharing part was previously added at https://github.com/xanzy/go-gitlab/pull/886.


group_members.go
```diff
@@ -201,6 +201,24 @@ func (s *GroupMembersService) ShareWithGroup(gid interface{}, opt *ShareWithGrou
 	return g, resp, err
 }
 
+// RemoveShareWithGroup allows to unshare a group from a group.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/groups.html#delete-link-sharing-group-with-another-group
+func (s *GroupMembersService) RemoveShareWithGroup(gid interface{}, groupID int, options ...RequestOptionFunc) (*Response, error) {
```
horjulf, line 207: `RemoveShareWithGroup` or `DeleteShareWithGroup` ?
On the project side, its implemented in the `Project` struct instead of `ProjectMembers` and is named `DeleteSharedProjectFromGroup`.

horjulf: Project side all are named `Delete*`, however here we have `RemoveGroupMember`. 🤔

group_members.go
```diff
@@ -201,6 +201,24 @@ func (s *GroupMembersService) ShareWithGroup(gid interface{}, opt *ShareWithGrou
 	return g, resp, err
 }
 
+// RemoveShareWithGroup allows to unshare a group from a group.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/groups.html#delete-link-sharing-group-with-another-group
```
svanharmelen, line 206:
```suggestion
// GitLab API docs:
// https://docs.gitlab.com/ce/api/groups.html#delete-link-sharing-group-with-another-group
```

group_members.go
svanharmelen: I prefer to follow the docs (and their layout) as much as possible. So that would mean to use the name `DeleteShareWithGroup`. As for the `Project` vs `ProjectMembers`, I think this has changes over time and the docs layout (and API) was different when this was created.

While I'm all for consistency, in this case I would not try to fix it as it probably causes more harm then good. So let's keep it here, but update the name. Thanks @horjulf!

horjulf: Sounds good, updated.

svanharmelen:
Thanks @horjulf! LGTM!

PR #895, opened 2020-08-01
Types compare helpers

ax1036:
It should be usefull to compare types easily

if #890 merged, a POC updating a project settings, only if needed:
```
	opt = &gitlab.EditProjectOptions{}

	aProject, _, _ := Client.Projects.GetProject(42, nil, nil)

	if !aProject.ForkingAccessLevel.IsDisabledAccessControl() {
		opt.ForkingAccessLevel = gitlab.AccessControl(gitlab.DisabledAccessControl)
		needUpdate = true
	}

	if needUpdate {
		client.Projects.EditProject(aProject.ID, opt)
	}
```

@svanharmelen Need some review if it's ok, before doing it to other types :wink:

### DIFF
--- a/types.go
+++ b/types.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"reflect"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


svanharmelen:
Not too big a fan of these kind of methods. And when comparing a whole struct, it isn't needed right? Only when you want to check single individual values, but I don't see the need for that in this package.

ax1036:
> Not too big a fan of these kind of methods. And when comparing a whole struct, it isn't needed right? Only when you want to check single individual values, but I don't see the need for that in this package.

I'm currently using this:
`if !compareAccessControl(aProject.ForkingAccessLevel, gitlab.DisabledAccessControl) {`
instead of purpose this PR:
`if !aProject.ForkingAccessLevel.IsDisabledAccessControl() {`

For me, it's nicest with "IsDisabledAccessControl" directly on type. But anyway, thx for feedback.

*closing*

PR #903, opened 2020-08-06
Pipeline emails

jeremad:
Add a missing field in the Pipelines Email properties. The API is not well documented: https://docs.gitlab.com/ee/api/services.html#get-pipeline-emails-service-settings but from testing, the field is indeed present.


jeremad:
You're very welcome, I should be thanking you for this work! Any chance to have a quick 0.34.1 ?

svanharmelen:
Done 👍

PR #906, opened 2020-08-07
ListGroupEpicsOptions Params Change

DavidMBusey:
Added more parameters for ListGroupEpicsOptions

### DIFF
--- a/epics.go
+++ b/epics.go
@@ -65,7 +65,6 @@ type ListGroupEpicsOptions struct {
 	Milestone        *string    `url:"milestone,omitempty" json:"milestone,omitempty"`
 	Scope            *string    `url:"scope,omitempty" json:"scope,omitempty"`
 	AuthorID         *int       `url:"author_id,omitempty" json:"author_id,omitempty"`
-	AssigneeID       *int       `url:"assignee_id,omitempty" json:"assignee_id,omitempty"`
 	MyReactionEmoji  *string    `url:"my_reaction_emoji,omitempty" json:"my_reaction_emoji,omitempty"`
 	IIDs             []int      `url:"iids[],omitempty" json:"iids,omitempty"`
 	OrderBy          *string    `url:"order_by,omitempty" json:"order_by,omitempty"`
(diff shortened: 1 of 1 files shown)


DavidMBusey:
```
AssigneeID       *int       `url:"assignee_id,omitempty" json:"assignee_id,omitempty"
```
should be removed actually.

DavidMBusey:
The reason I committed this is because the `CreatedAfter` param is very useful.

epics.go
```diff
@@ -59,12 +59,22 @@ func (e Epic) String() string {
...
+	MyReactionEmoji  *string    `url:"my_reaction_emoji,omitempty" json:"my_reaction_emoji,omitempty"`
+	IIDs             []int      `url:"iids[],omitempty" json:"iids,omitempty"`
+	OrderBy          *string    `url:"order_by,omitempty" json:"order_by,omitempty"`
+	Sort             *string    `url:"sort,omitempty" json:"sort,omitempty"`
+	Search           *string    `url:"search,omitempty" json:"search,omitempty"`
+	CreatedAfter     *time.Time `url:"created_after,omitempty" json:"created_after,omitempty"`
+	CreatedBefore    *time.Time `url:"created_before,omitempty" json:"created_before,omitempty"`
+	UpdatedAfter     *time.Time `url:"updated_after,omitempty" json:"updated_after,omitempty"`
+	UpdatedBefore    *time.Time `url:"updated_before,omitempty" json:"updated_before,omitempty"`
+	Confidential     *bool      `url:"confidential,omitempty" json:"confidential,omitempty"`
```
svanharmelen, line 77: I don't see `Milestone`, `Scope`, IIDs` and `Confidential` in the docs. Are you sure that are valid options?

And if we want to make this list complete, we probably need to add `include_ancestor_groups` and `include_descendant_groups` as well.

Last but not least, I try to keep the ordering inline with the ordering in de docs. So please make so that is fixed. Thanks!

DavidMBusey: Sorry, I just rushed it. I'll commit a fix for those issues today.

DavidMBusey:
@svanharmelen I fixed those issues that you mentioned. The desc/asc groups was not tested, but I know the Time-based ones work.

PR #910, opened 2020-08-10
Swap the concrete limiter to interface and add WithCustomLimiter

### LINKED ISSUE #909: Add support for custom rate limiter
Sort of a feature request, use cases may include private gitlab instance not doing the rate limiting itself and not returning rate limit headers and / or when someone wishes to limit load to lower than what's automatically configured. 

Can make a pull request to add it, wanted some thoughts and considerations that I'm not thinking about :)

Thanks.

2785:
closes #909

### DIFF
--- a/client_options.go
+++ b/client_options.go
@@ -47,3 +47,13 @@ func WithoutRetries() ClientOptionFunc {
 		return nil
 	}
 }
+
+// WithCustomLimiter injects a custom rate limiter to the client.
+func WithCustomLimiter(limiter RateLimiter) ClientOptionFunc {
+	return func(c *Client) error {
+		c.configureLimiterOnce.Do(func() {
+			c.limiter = limiter
+		})
+		return nil
+	}
+}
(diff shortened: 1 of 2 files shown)


client_options.go
```diff
@@ -47,3 +47,13 @@ func WithoutRetries() ClientOptionFunc {
 		return nil
 	}
 }
+
+// WithCustomLimiter injects a custom rate limiter to the client.
+func WithCustomLimiter(limiter RateLimiter) ClientOptionFunc {
```
svanharmelen, line 52: I try to keep these options ordered in alphabetical order, so if you could update that...

gitlab.go
```diff
@@ -185,6 +185,11 @@ type ListOptions struct {
 	PerPage int `url:"per_page,omitempty" json:"per_page,omitempty"`
 }
 
+// RateLimiter specifies a dependency that can wait on a context.
```
svanharmelen, line 188:
```suggestion
// RateLimiter describes the interface that all (custom) rate limiters must implement.
```

client_options.go
2785: o yeah I can do that.

2785: done.

PR #916, opened 2020-08-15
Add pipeline schedules trigger

### LINKED ISSUE #870: allow to run project schedule
Hi, 
i could not find a client method that would execute project schedule.
Would like to do that thanks

2785:
closes #870


2785:
trynna add tests. Any reason to not use testify assert?

PR #923, opened 2020-08-28
Add ListServices method

joonas:
I was looking for this method to use in some internal code I am working on, and since I couldn't find it I thought I'd add it.


svanharmelen (approved):
LGTM! Thanks @joonas!

PR #929, opened 2020-08-30
Support NotLabels for MergeRequestOptions and New Get Inherited Member Function

ekristen:
Adds the ability to query Merge Requests by Not Labels and adds a function to get inherited members of a project.

### DIFF
--- a/group_milestones.go
+++ b/group_milestones.go
@@ -59,7 +59,7 @@ type ListGroupMilestonesOptions struct {
 	IIDs                    []int  `url:"iids,omitempty" json:"iids,omitempty"`
 	State                   string `url:"state,omitempty" json:"state,omitempty"`
 	Search                  string `url:"search,omitempty" json:"search,omitempty"`
-	IncludeParentMilestones bool   `url:"include_parent_milestones,omitempty" json:"include_parent_milestones,omitempty"
+	IncludeParentMilestones bool   `url:"include_parent_milestones,omitempty" json:"include_parent_milestones,omitempty"`
 }
 
 // ListGroupMilestones returns a list of group milestones.
--- a/group_milestones.go
+++ b/group_milestones.go
@@ -41,9 +41,9 @@ type GroupMilestone struct {
 	StartDate   *ISOTime   `json:"start_date"`
 	DueDate     *ISOTime   `json:"due_date"`
 	State       string     `json:"state"`
-	Expired     bool       `json:"expired"`
 	UpdatedAt   *time.Time `json:"updated_at"`
 	CreatedAt   *time.Time `json:"created_at"`
+	Expired     *bool      `json:"expired"`
 }
 
 func (m GroupMilestone) String() string {
@@ -57,10 +57,11 @@ func (m GroupMilestone) String() string {
 // https://docs.gitlab.com/ce/api/group_milestones.html#list-group-milestones
 type ListGroupMilestonesOptions struct {
 	ListOptions
-	IIDs                    []int  `url:"iids,omitempty" json:"iids,omitempty"`
-	State                   string `url:"state,omitempty" json:"state,omitempty"`
-	Search                  string `url:"search,omitempty" json:"search,omitempty"`
-	IncludeParentMilestones bool   `url:"include_parent_milestones,omitempty" json:"include_parent_milestones,omitempty"`
+	IIDs                    []int   `url:"iids,omitempty" json:"iids,omitempty"`
+	State                   *string `url:"state,omitempty" json:"state,omitempty"`
+	Title                   *string `url:"title,omitempty" json:"title,omitempty"`
+	Search                  *string `url:"search,omitempty" json:"search,omitempty"`
+	IncludeParentMilestones *bool   `url:"include_parent_milestones,omitempty" json:"include_parent_milestones,omitempty"`
 }
 
 // ListGroupMilestones returns a list of group milestones.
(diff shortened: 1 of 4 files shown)


project_members.go
```diff
@@ -117,6 +117,31 @@ func (s *ProjectMembersService) GetProjectMember(pid interface{}, user int, opti
 	return pm, resp, err
 }
 
+// GetAllProjectMember gets a project team member, including inherited
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/members.html#get-a-member-of-a-group-or-project-including-inherited-members
+func (s *ProjectMembersService) GetAllProjectMember(pid interface{}, user int, options ...RequestOptionFunc) (*ProjectMember, *Response, error) {
```
svanharmelen, line 124: I think the `All` part in this name is a little misleading. How about something like:
```suggestion
func (s *ProjectMembersService) GetInheritedProjectMember(pid interface{}, user int, options ...RequestOptionFunc) (*ProjectMember, *Response, error) {
```

svanharmelen (changes requested):

merge_requests.go
```diff
@@ -189,6 +190,7 @@ type ListGroupMergeRequestsOptions struct {
 	Milestone       *string    `url:"milestone,omitempty" json:"milestone,omitempty"`
 	View            *string    `url:"view,omitempty" json:"view,omitempty"`
 	Labels          *Labels    `url:"labels,omitempty" json:"labels,omitempty"`
+	NotLabels       *Labels    `url:"not[labels],comma,omitempty" json:"not[labels],omitempty"`
```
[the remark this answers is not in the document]
ekristen: The Labels was already a pointer so I just follow suite? It seems like it's mixed throughout the library, some places they are pointers, some places they are not.

project_members.go
ekristen: Sounds good to me. I wasn't really sure either, I was just following the ListAll example since the API has the `all` in it. Happy to accept the suggestion.

project_members.go
```diff
@@ -117,6 +117,31 @@ func (s *ProjectMembersService) GetProjectMember(pid interface{}, user int, opti
 	return pm, resp, err
 }
 
+// GetAllProjectMember gets a project team member, including inherited
```
svanharmelen, line 120: This needs to change as well now:
```suggestion
// GetInheritedProjectMember gets a project team member, including inherited.
```

group_milestones.go
```diff
@@ -56,9 +57,10 @@ func (m GroupMilestone) String() string {
 // https://docs.gitlab.com/ce/api/group_milestones.html#list-group-milestones
 type ListGroupMilestonesOptions struct {
 	ListOptions
-	IIDs   []int  `url:"iids,omitempty" json:"iids,omitempty"`
-	State  string `url:"state,omitempty" json:"state,omitempty"`
-	Search string `url:"search,omitempty" json:"search,omitempty"`
+	IIDs                    []int  `url:"iids,omitempty" json:"iids,omitempty"`
+	State                   string `url:"state,omitempty" json:"state,omitempty"`
+	Search                  string `url:"search,omitempty" json:"search,omitempty"`
+	IncludeParentMilestones bool   `url:"include_parent_milestones,omitempty" json:"include_parent_milestones,omitempty"`
```
svanharmelen, line 63: I know you did not add all of these, but the `string` and `bool` fields in this struct need to be pointers. And maybe we should add `title` as well to make the list complete?

group_milestones.go
```diff
@@ -41,6 +41,7 @@ type GroupMilestone struct {
 	StartDate   *ISOTime   `json:"start_date"`
 	DueDate     *ISOTime   `json:"due_date"`
 	State       string     `json:"state"`
+	Expired     bool       `json:"expired"`
```
svanharmelen, line 44: Following the docs, this should be the last field of the struct.

merge_requests.go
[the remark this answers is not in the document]
svanharmelen: `Labels` should never be pointers. I fixed it for all other fields, so if you rebase on master it should be good (except for this one you now added).

milestones.go
```diff
@@ -41,6 +41,7 @@ type Milestone struct {
 	StartDate   *ISOTime   `json:"start_date"`
 	DueDate     *ISOTime   `json:"due_date"`
 	State       string     `json:"state"`
+	Expired     bool       `json:"expired"`
```
svanharmelen, line 44: Following the docs, this should be the last field of the struct.

merge_requests.go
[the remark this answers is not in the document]
ekristen: Sounds good, I can take care of it after a rebase.

project_members.go
ekristen: Yup, missed it, thanks.

group_milestones.go
ekristen: What's the guidance for when a pointer should or should not be used?

ekristen:
It's not clear to me when pointers should and should not be used on this library, there seems to be a lot of inconsistencies. Any guidance would be appreciated.

group_milestones.go
ekristen: Add title, fixed pointers.

svanharmelen (approved):
Thanks for sticking with me... LGTM!

svanharmelen:
>It's not clear to me when pointers should and should not be used on this library, there seems to be a lot of inconsistencies. Any guidance would be appreciated.

In all option structs it's needed to have types the can indicate if that are set or not. The fact that Go always initializes values to their "zero" value means that if you do not set a `bool` in an option struct, it will actually become `false`. So to prevent that you need to have a way to distinguish from a field not being set, and a field being set with the zero value by the user.

As the zero of a pointer and of referenced types is `nil`, we can use that to solve this issue. So that means that in the option structs all non-referenced types need to be pointers.

ekristen:
Thanks for the explanation. It makes sense.

PR #932, opened 2020-09-05
Adds group level push rules API

Starttoaster:
This merge adds the push rules API added for groups in GitLab 13.4. Currently available for gitlab.com and soon to be available to self-hosted gitlab instances.

https://docs.gitlab.com/ee/api/groups.html#push-rules-starter

Also took a look at the project level push rules and added newly available push rules.

### DIFF
--- a/groups.go
+++ b/groups.go
@@ -605,9 +605,7 @@ func (s *GroupsService) EditGroupPushRule(gid interface{}, opt *EditGroupPushRul
 	return gpr, resp, err
 }
 
-// DeleteGroupPushRule removes a push rule from a group. This is an
-// idempotent method and can be called multiple times. Either the push rule is
-// available or not.
+// DeleteGroupPushRule deletes the push rules of a group.
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/groups.html#delete-group-push-rule-starter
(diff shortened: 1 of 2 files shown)


groups.go
```diff
@@ -471,3 +471,155 @@ func (s *GroupsService) DeleteGroupLDAPLinkForProvider(gid interface{}, provider
 
 	return s.client.Do(req, nil)
 }
+
+// GroupPushRules represents a group push rule.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/groups.html#get-group-push-rules-starter
```
svanharmelen, line 478:
```suggestion
// https://docs.gitlab.com/ee/api/groups.html#get-group-push-rules
```

svanharmelen: I suggest to remove `-starter` for all URLs as they don't resolve for me.

svanharmelen (changes requested):
Only one small comment about the doc URLs, rest look perfect 👍

Starttoaster:
Thanks @svanharmelen ! Went ahead and removed those '-starter' suffixes.

svanharmelen (approved):
Thanks @Starttoaster! LGTM!

PR #936, opened 2020-09-08
Add WebURL to users.

ruimiguelsantos:
Adds the missing field to User.

https://docs.gitlab.com/ee/api/users.html#single-user

### DIFF
--- a/users.go
+++ b/users.go
@@ -87,6 +87,7 @@ type User struct {
 	PrivateProfile            bool               `json:"private_profile"`
 	SharedRunnersMinutesLimit int                `json:"shared_runners_minutes_limit"`
 	CustomAttributes          []*CustomAttribute `json:"custom_attributes"`
+	WebURL                    string             `json:"web_url"`
 }
 
 // UserIdentity represents a user identity.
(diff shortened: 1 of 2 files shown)


users.go
```diff
@@ -87,6 +87,7 @@ type User struct {
 	PrivateProfile            bool               `json:"private_profile"`
 	SharedRunnersMinutesLimit int                `json:"shared_runners_minutes_limit"`
 	CustomAttributes          []*CustomAttribute `json:"custom_attributes"`
+	WebURL                    string             `json:"web_url"`
```
svanharmelen, line 90: Please place this field above `CreateAt` so it follows the ordering in the docs (which I try to use as guideline here). Thanks!

svanharmelen: And it seems some tests also need to be updated to make Travis pass...

ruimiguelsantos: Sure, sorry about that. I've moved the field to honour the docs ordering and updated the failing tests.

svanharmelen (approved):
Thanks @ruimiguelsantos! LGTM!

PR #939, opened 2020-09-11
Add enabled attribute

sfang97:
Enabled is an editable cluster attribute-- will also update the GItlab docs to reflect this

### DIFF
--- a/project_clusters.go
+++ b/project_clusters.go
@@ -178,7 +178,7 @@ type EditClusterOptions struct {
 	Name                *string                        `url:"name,omitempty" json:"name,omitempty"`
 	Domain              *string                        `url:"domain,omitempty" json:"domain,omitempty"`
 	EnvironmentScope    *string                        `url:"environment_scope,omitempty" json:"environment_scope,omitempty"`
-	Enabled							*bool													 `url:"enabled,omitempty" json:"enabled,omitempty"`
+	Enabled             *bool                          `url:"enabled,omitempty" json:"enabled,omitempty"`
 	ManagementProjectID *string                        `url:"management_project_id,omitempty" json:"management_project_id,omitempty"`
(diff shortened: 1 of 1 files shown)


sfang97:
@svanharmelen Mind taking a look at this little change? :)

project_clusters.go
```diff
@@ -178,6 +178,7 @@ type EditClusterOptions struct {
 	Name                *string                        `url:"name,omitempty" json:"name,omitempty"`
 	Domain              *string                        `url:"domain,omitempty" json:"domain,omitempty"`
 	EnvironmentScope    *string                        `url:"environment_scope,omitempty" json:"environment_scope,omitempty"`
+	Enabled             *bool                          `url:"enabled,omitempty" json:"enabled,omitempty"`
```
svanharmelen, line 181: I don't see this field in the docs, can you show me something that confirms this is a valid option?

Starttoaster: I was curious and took a look. Best I can tell, `enabled` is not an available attribute in the EditCluster function.

It does appear to be valid for the AddCluster function, however. https://docs.gitlab.com/ee/api/project_clusters.html#add-existing-cluster-to-project

Starttoaster: The curl I used to test this:

`curl --header "Private-Token: <token>" -X PUT https://gitlab.com/api/v4/projects/<pid>/clusters/<cluster-id>?enabled=false`

It returned the cluster's object details, but didn't seem to toggle whether or not the cluster was enabled.

sfang97: Huh okay, you can edit whether or not the cluster is enabled via the GUI but maybe you can't through the API. I'll close this PR since it probably doesn't work as intended.

svanharmelen: Thanks for checking @Starttoaster!

PR #940, opened 2020-09-11
Add managed attribute to instance cluster

sfang97:
Add managed attribute to instance cluster

### DIFF
--- a/instance_clusters.go
+++ b/instance_clusters.go
@@ -37,7 +37,7 @@ type InstanceCluster struct {
 	ID                 int                 `json:"id"`
 	Name               string              `json:"name"`
 	Domain             string              `json:"domain"`
-	Managed						 bool								 `json:"managed"`
+	Managed            bool                `json:"managed"`
 	CreatedAt          *time.Time          `json:"created_at"`
 	ProviderType       string              `json:"provider_type"`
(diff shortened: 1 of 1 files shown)


svanharmelen (approved):
Thanks @sfang97 👍

instance_clusters.go
```diff
@@ -37,6 +37,7 @@ type InstanceCluster struct {
 	ID                 int                 `json:"id"`
 	Name               string              `json:"name"`
 	Domain             string              `json:"domain"`
+	Managed            bool                `json:"managed"`
```
svanharmelen, line 40: @Starttoaster could you maybe check this one as well? I noticed the docs mention this as an option when adding a cluster (https://docs.gitlab.com/ee/api/instance_clusters.html#add-existing-instance-cluster) but not sure if it actually is a field in the response...

Starttoaster: Absolutely! This looks like a self-hosted feature, and I'm personally a user of GitLab.com, but to be honest I've been meaning to deploy my own instance one of these days anyway. I'll take a look and get back to this.

svanharmelen: Thanks! But I don't want to put a burden on your time, so please only have a look if it fits your personal schedule and interest!

PR #947, opened 2020-09-25
Update Pipeline test report

avanloo:
Gitlab documentation wasn't entirly as expected:

- Remove system_output and strack_trace field aren't consitant (see test data)
- Updated type of time fields to float
- Added example how to retrive testreport

### DIFF
--- a/pipelines.go
+++ b/pipelines.go
@@ -82,7 +82,7 @@ func (p Pipeline) String() string {
 
 // PipelineTestReport contains a detailed report of a test run.
 type PipelineTestReport struct {
-	TotalTime    int                  `json:"total_time"`
+	TotalTime    float64              `json:"total_time"`
 	TotalCount   int                  `json:"total_count"`
 	SuccessCount int                  `json:"success_count"`
 	FailedCount  int                  `json:"failed_count"`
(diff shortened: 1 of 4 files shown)


pipelines.go
```diff
@@ -105,12 +105,10 @@ type PipelineTestSuites struct {
 
 // PipelineTestCases contains test cases details.
 type PipelineTestCases struct {
-	Status        string `json:"status"`
-	Name          string `json:"name"`
-	Classname     string `json:"classname"`
-	ExecutionTime int    `json:"execution_time"`
-	SystemOutput  string `json:"system_output"`
-	StackTrace    string `json:"stack_trace"`
```
svanharmelen, line 113: Why are you removing these two? The docs seem to mention they are actually returned.

avanloo: Netto change is removal of SystemOutput / StackTrace; this due that response could be nill, empty arry, or single object. and which one you get is unpredicateble. 

Wasn't able to come up with a good sollution

svanharmelen (approved):
Thanks @avanloo!

pipelines.go
svanharmelen: OK... Then let's merge this for now and properly sort it later when someone needs the fields 👍🏻

PR #949, opened 2020-09-29
Add in to ListIssuesOptions properties with tests

### LINKED ISSUE #948: ListIssuesOptions missing 'in' property
I'm looking to be able to search issues by title only (as the descriptions can be quite long), but finding that the search is returning results that match the text in the description as well. 

I found that the search can be limited to only the title with the `in` parameter. However, that is not yet provided in this client.

GitLab documentation: https://docs.gitlab.com/ce/api/issues.html#list-issues
ListIssuesOptions source: https://github.com/xanzy/go-gitlab/blob/v0.38.1/issues.go#L201

Not sure if that was intentionally excluded, or if it was accidentally missed. I don't see when it was added, so it may be new.

Happy to put together a PR if it's something that would be welcomed.

burtonr:
As discussed in Issue #948, this changeset adds an `In` property to the `ListIssuesOptions` as documented in the GitLab API: https://docs.gitlab.com/ce/api/issues.html#list-issues

I've also added 2 tests to confirm that the parameter is added properly with the expected values. All tests pass. 

I've run `go fmt` as well to ensure that there are no formatting issues.

Resolves #948


svanharmelen (approved):
Thanks @burtonr! LGTM!

PR #950, opened 2020-10-01
(groups) support listing descendant groups

zaquestion:
Ref: https://docs.gitlab.com/ce/api/groups.html#list-a-groups-descendant-groups


svanharmelen (approved):
Thanks @zaquestion! 👍🏻

Issue #944, opened 2020-09-22
Add support for User Memberships


edshelton:
https://docs.gitlab.com/ee/api/users.html#user-memberships-admin-only

This would be useful to be able to get the lowest `access_level` for a given user in order to determine whether that user is using a license seat. Any GitLab user with an `access_level` of `20` or greater and a `state` that != `blocked` consumes a seat, `20` being the `Developer` role.

cpackingham:
Mind if I take this one?

svanharmelen:
By all means :)

PR #953, opened 2020-10-05
User memberships

cpackingham:
Resolves https://github.com/xanzy/go-gitlab/issues/944

- Added user memberships method
- Added a unit test with some test data

Tried to test this locally but I don't have access to an admin account. It looks like you might need to create a self-hosted gitlab instance, is that right?


testdata/get_user_memberships.json
```diff
@@ -0,0 +1,14 @@
...
+    "source_type": "Project",
+    "access_level": "20"
+  },
+  {
+    "source_id": 3,
+    "source_name": "Group three",
+    "source_type": "Namespace",
+    "access_level": "20"
+  }
+]
```
svanharmelen, line 14: Please make sure the file ends with a newline.

users_test.go
```diff
@@ -234,3 +236,23 @@ func TestActivateUser_UserNotFound(t *testing.T) {
...
+	path := fmt.Sprintf("/%susers/1/memberships", apiVersionPath)
+	mux.HandleFunc(path, func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "GET")
+		mustWriteHTTPResponse(t, w, "testdata/get_user_memberships.json")
+	})
+
+	opt := new(GetUserMembershipOptions)
+
+	memberships, _, err := client.Users.GetUserMemberships(1, opt)
+
```
svanharmelen, line 253:
```suggestion
```

users_test.go
```diff
@@ -234,3 +236,23 @@ func TestActivateUser_UserNotFound(t *testing.T) {
...
+	require.NoError(t, err)
+
+	want := []*UserMembership{{SourceID: 1, SourceName: "Project one", SourceType: "Project", AccessLevel: "20"}, {SourceID: 3, SourceName: "Group three", SourceType: "Namespace", AccessLevel: "20"}}
+	require.Equal(t, want, memberships)
```
svanharmelen, line 257: While require is fine, assert would probably also be a good option here.

svanharmelen (changes requested):
Thanks @cpackingham! Few super minor comments, looks good otherwise 👍🏻

cpackingham:
Okay I think I got all of the suggestions in, how does it look?

svanharmelen (approved):
Thanks @cpackingham! LGTM!

PR #954, opened 2020-10-12
allow to update file with empty content

Murphylu1993:


svanharmelen:
Good find @Murphylu1993! Yet the correct solution is to make the struct follow the option struct pattern used in the rest of this package. So I created #955 for that in order to solve this. Please let me know if the fix in that PR also works for you. Thanks!

svanharmelen:
Would be nice if you can verify the solution I proposed in PR #955 so that we can merge that one. Thanks!

Murphylu1993:
> Would be nice if you can verify the solution I proposed in PR #955 so that we can merge that one. Thanks!

sure, merge plz

svanharmelen:
Check.. Thanks!

PR #956, opened 2020-10-13
Add User to MergeRequest struct

profclems:
The GitLab merge request endpoint has `"user" : {"can_merge" : false},` as part of the payload which indicates whether the current authenticated user has permission to merge a particular merge request. 
This PR adds it to the MergeRequest struct so it can be accessed with `User.CanMerge`

Reference: https://docs.gitlab.com/ee/api/merge_requests.html#get-single-mr


svanharmelen (approved):

svanharmelen:
Thanks!

PR #958, opened 2020-10-15
Add method to delete artifacts for a job

ferrastas:
As described in [GitLab's documentation](https://docs.gitlab.com/ce/api/job_artifacts.html#delete-artifacts) since GitLab 11.9 there is an endpoint available to remove the artifacts for a job, a new method has been introduced in `JobsService` to add support for this missing API call.


svanharmelen (approved):
Thanks @ferrastas 👍🏻

PR #963, opened 2020-10-26
Extend ProjectsNamespace with avatar_url and web_url

vnarek:
These fields were present in the Gitlab v4 api since 11.0, but without proper documentation. The only place where we can find some information about these fields is in the *Note* under https://docs.gitlab.com/13.3/ee/api/projects.html#get-single-project API description.

I tested the code on /projects/{:id} and /projects.


svanharmelen (approved):
Thanks!

PR #965, opened 2020-10-30
feat: add list pipeline bridges

Oyabi:
Hi,

It seems that [List pipeline bridges](https://docs.gitlab.com/ee/api/jobs.html#list-pipeline-bridges) was not implemented.
I need it for a project so here the code. 
I'm a beginner in golang so there might be some optimizations between Bridges and Jobs structures but unfortunately I don't know how to do it.
The function `ListPipelineBridges()` is just a copy of `ListPipelineJobs()` with adaptation.

Hope everything is ok ;)

### DIFF
--- a/jobs.go
+++ b/jobs.go
@@ -76,6 +76,9 @@ type Job struct {
 	User   *User  `json:"user"`
 }
 
+// Bridge represents a pipeline bridge.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/jobs.html#list-pipeline-bridges
 type Bridge struct {
 	Commit       *Commit    `json:"commit"`
 	Coverage     float64    `json:"coverage"`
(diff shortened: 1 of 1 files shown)


jobs.go
```diff
@@ -76,6 +76,42 @@ type Job struct {
 	User   *User  `json:"user"`
 }
 
+type Bridge struct {
```
svanharmelen, line 79: Please add a comment with a link...

jobs.go
```diff
@@ -136,6 +172,32 @@ func (s *JobsService) ListPipelineJobs(pid interface{}, pipelineID int, opts *Li
 	return jobs, resp, err
 }
 
+// ListPipelinebridges gets a list of bridges for specific pipeline in a
+// project. If the pipeline ID is not found, it will respond with 404.
```
svanharmelen, line 176:
```suggestion
// project.
```

svanharmelen (changes requested):
Two super small comments, rest looks good 👍🏻

svanharmelen (approved):
Thanks @Oyabi! LGTM!

PR #967, opened 2020-11-04
Update event webhooks

joerocklin:
This is not an exhaustive set of updates to these structs, but they were needed for some work I was doing. Added fields include:

- **JobEvent**
  - Add `BuildFailureReason` string
  - Add `Runner` struct
- **PipelineEvent**
  - Add `Source` string

Example structs were pulled from the [gitlab documentation](https://docs.gitlab.com/ee/user/project/integrations/webhooks.html).


svanharmelen (approved):
Thanks @joerocklin! LGTM 👍🏻

PR #364, opened 2018-04-01
Update ListPipelineJobs to return array of pointers

jamesramsay:
Most list functions return array of pointers, not values. Updated for
consistency.

PR #968, opened 2020-11-09
Update ListProjectJobs to return array of pointers

firelizzard18:
This is essentially identical to #364:

> Most list functions return array of pointers, not values. Updated for
consistency.

Right now, `JobsService.ListPipelineJobs` returns `[]*Job` and `JobsService.ListProjectJobs` returns `[]Job`. This MR fixes the situation by modifying `ListProjectJobs` to return `[]*Job`


svanharmelen (approved):
Good catch! Thanks @firelizzard18 👍🏻

PR #969, opened 2020-11-09
Add header-based auth methods

firelizzard18:
This change adds support for using CI tokens and deploy tokens to authenticate requests, by adding a constructor, `NewHeaderAuthClient`, that accepts a header and token.

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -60,6 +60,7 @@ const (
 	basicAuth authType = iota
 	oAuthToken
 	privateToken
+	jobToken
 )
 
 // A Client manages communication with the GitLab API.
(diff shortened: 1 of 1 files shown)


firelizzard18:
If #970 is accepted, I will close this PR in favor of the change I suggest in that issue.

svanharmelen:
Thanks and sorry for leaving this open for so long without giving any update. I think that by now we support all the tokens you tried to support using this PR. I understand it's a bit less flexible and might have a tiny bit of duplicate code, but for now I'm not really willing to make any big changes around this point. So sorry, but I will not be merging this PR...

PR #971, opened 2020-11-13
feat(commit status): add AllowFailure in CommitStatus struct.

likaiwen123:
#972 

When checking the status of a commit, the allow_failure option should be checked. This key is added in the commit status API in GitLab 12, so the user can combine that with the status to get the final decision.


svanharmelen (approved):
Thanks @likaiwen123!

PR #974, opened 2020-11-20
AddProjectMemberOptions.UserId shoud be *string

itscaro:
https://docs.gitlab.com/ee/api/members.html#add-a-member-to-a-group-or-project

We can pass a comma-separated string to do bulk.


svanharmelen (approved):
Thanks @itscaro! LGTM!

PR #977, opened 2020-11-24
Added ExpiresAt field to AddSSHKeyOptions according to Gitlab API spec

mrahbar:
The Gitlab API now supports the optional field expires_at when adding a new SSH key to the user account. This has been previously only possible via UI. This PR adds the struct field ExpiresAt in the same way as already present in ImpersonationToken.


svanharmelen (changes requested):
Looks good @mrahbar, but please run `go fmt` on your code...

svanharmelen (approved):
Thanks! LGTM!

Issue #980, opened 2020-11-24
Add fields CreatedAt and UpdatedAt to Pipeline within a Deployment


sk000f:
In line with https://docs.gitlab.com/ee/api/deployments.html

Add the created_at and updated_at fields within the Pipeline object, within the Deployable, as part of a Deployment.

PR #981, opened 2020-11-24
Fix: #980 add CreatedAt and UpdatedAt fields to a Pipeline within a Deployment

sk000f:


deployments.go
```diff
@@ -53,10 +53,12 @@ type Deployment struct {
...
-			ID     int    `json:"id"`
-			SHA    string `json:"sha"`
-			Ref    string `json:"ref"`
-			Status string `json:"status"`
+			ID        int    `json:"id"`
+			SHA       string `json:"sha"`
+			Ref       string `json:"ref"`
+			Status    string `json:"status"`
+			CreatedAt string `json:"created_at"`
+			UpdatedAt string `json:"updated_at"`
```
svanharmelen, line 61: These should be of type `*time.Time` I guess...

svanharmelen (changes requested):

sk000f:
Yes good point. Updated.

svanharmelen (approved):
Thanks!

PR #982, opened 2020-11-25
Fix travis build

ferrastas:
Checking the library with a colleague we discovered that the lint step in Travis is not executed. Seems that I'm a beta user for https://docs.travis-ci.com/user/build-config-validation/ and it's reporting some incompatibilities.

Last build for master:

This PR:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,21 @@
 language: go
+
 arch:
-    - amd64
-    - ppc64le
+  - amd64
+  - ppc64le
 
 go:
   - 1.13.x
   - 1.14.x
+  - 1.x
   - master
 
-stages:
-  - lint
-  - test
-
-jobs:
-  include:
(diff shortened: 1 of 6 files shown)


svanharmelen (approved):
Thanks @ferrastas! 👍🏻

ferrastas:
@svanharmelen I'm checking why there is only one lint job (when we should have one per go version and arch)

ferrastas:
@svanharmelen I slightly changed the original approach. I tried several approaches using the [TravisCI linter](https://config.travis-ci.com/explore) and have more than one job using matrix is not possible. So in that case linters are moved to the `before_script` section. Also reduces the number of jobs to execute per build.

Feel free to merge it if you like that new approach :D

svanharmelen (approved):
Looks good to me. Thanks again @ferrastas 👍🏻

PR #983, opened 2020-11-25
Add file, system_output, stack_trace, attachment_url

ruudvanderweijde:
The fields `file`, `system_output`, `stack_trace`, `attachment_url` were
missing. As you can see in the api spec below, the fields can be string
on null.

Besides that, in `testdata/get_pipeline_testreport.json` the `classname`
had the value `null`, but that should not be possible according to the spec.

---

The api spec: https://gitlab.com/gitlab-org/gitlab-foss/-/blob/13-6-stable/spec/fixtures/api/schemas/entities/test_case.json
```
{
  "type": "object",
  "required" : [
    "status",
    "name"
  ],
  "properties": {
    "status": { "type": "string" },
    "name": { "type": "string" },
    "classname": { "type": "string" },
    "file": { "type": ["string", "null"] },
    "execution_time": { "type": "float" },
    "system_output": { "type": ["string", "null"] },
    "stack_trace":  { "type": ["string", "null"] },
    "attachment_url":  { "type": ["string", "null"] },
    "recent_failures": {
      "oneOf": [
        { "type": "null" },
        { "$ref": "test_case/recent_failures.json" }
      ]
    }
  },
  "additionalProperties": false
}
```

Related MR in which the `system_output` and `stack_trace` were removed: https://github.com/xanzy/go-gitlab/pull/947


ruudvanderweijde:
@svanharmelen could you have a look?

svanharmelen (approved):
Thanks @ruudvanderweijde!

ferrastas:
Just a small remark, and for future references. JSON schemas are located in https://gitlab.com/gitlab-org/gitlab/-/blob/master/spec/fixtures/api/schemas/entities/test_case.json.

PR #985, opened 2020-11-25
Add recent_failures to test case

ruudvanderweijde:
As of v13.6 there was a change to include recent failures data in test
reports response

Ref: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/45187


svanharmelen (approved):
Thanks 👍🏻

PR #986, opened 2020-11-26
fix(merge_requests/issues): fix Issue.ID type to handle Jira issues

QuentinBrosse:
This MR fixes the `Issue.ID` type field that can be an integer or a string to support third parties issue systems (like Jira).

The Gitlab documentation page is available [here](https://docs.gitlab.com/ee/api/merge_requests.html#list-issues-that-will-close-on-merge).

### DIFF
--- a/issues.go
+++ b/issues.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
+	"reflect"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


QuentinBrosse:
Hi @svanharmelen, can you give a look at this MR, please? :)

svanharmelen:
Sorry for the delay, will try to find some time...

svanharmelen (commented):
Sorry for the late response @QuentinBrosse, but I needed to think about this one... I think that the impact of changing the ID type from `int` to `interface{}` could be very impactful. So I would like to suggest an alternative solution which makes the purpose of the fields my explicit.

What do you think about a solution like this?

QuentinBrosse:
@svanharmelen don't worry it's okay.

Your suggestion is fine but, Gitlab supports other external issue trackers, not only Jira. Maybe we should rename the field `JiraID` to a more generic name? Something like `StringID`?

Keep in mind that the `id` field of the Gitlab API can be an `int` or a `string` ([the doc is available here](https://docs.gitlab.com/ee/api/merge_requests.html#list-issues-that-close-on-merge)).

What do you think about it? Thanks for your reply.

svanharmelen:
What about this suggestion `ExternalID`? It's still not the perfect solution, but I think this will satisfy most use cases...

QuentinBrosse (commented):
LGTM

PR #987, opened 2020-11-28
Adds IssueLink LinkType field

leominov:
ref: https://docs.gitlab.com/ee/api/issue_links.html#create-an-issue-link

### DIFF
--- a/issue_links.go
+++ b/issue_links.go
@@ -69,9 +69,9 @@ func (s *IssueLinksService) ListIssueRelations(pid interface{}, issueIID int, op
 //
 // GitLab API docs: https://docs.gitlab.com/ee/api/issue_links.html
 type CreateIssueLinkOptions struct {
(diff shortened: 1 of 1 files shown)


issue_links.go
```diff
@@ -68,6 +69,7 @@ func (s *IssueLinksService) ListIssueRelations(pid interface{}, issueIID int, op
 //
 // GitLab API docs: https://docs.gitlab.com/ee/api/issue_links.html
 type CreateIssueLinkOptions struct {
+	LinkType        *string `json:"link_type"`
```
svanharmelen, line 72: Please place this field below the target fields (so following the order in the docs). Thanks!

svanharmelen (changes requested):

svanharmelen (approved):
Thanks @leominov! LGTM!

PR #989, opened 2020-11-30
Update users.go

JuwanXu:
add note


users.go
```diff
@@ -88,6 +88,7 @@ type User struct {
 	PrivateProfile            bool               `json:"private_profile"`
 	SharedRunnersMinutesLimit int                `json:"shared_runners_minutes_limit"`
 	CustomAttributes          []*CustomAttribute `json:"custom_attributes"`
+	Note                      string             `json:"note"`
```
svanharmelen, line 91: Can you place this above `Identities` so it matches the ordering of the docs? Thanks!

svanharmelen (approved):
Thanks @JuwanXu!

PR #990, opened 2020-12-01
Fix missing InstanceVariables service

horjulf:
Fixes a crash due to missing InstanceVariables service in the client.


svanharmelen (approved):
Good catch @horjulf!

PR #996, opened 2020-12-01
Added PerPage and Page Options

mikeangstadt:
Added PerPage and Page pagination options to ListGroupsOptions & ListGroupProjectsOptions


svanharmelen:
Please make sure your code is always formatted using `go fmt`.

But in this case these fields are actually already added by the embedded `ListOptions` struct: https://github.com/xanzy/go-gitlab/blob/master/gitlab.go#L181-L187

So I'll close this one again...

PR #998, opened 2020-12-09
Use *bool for Silver/Premium push rules

jtdoepke:
The `commit_committer_check` and `reject_unsigned_commits` project and group push rules are only available in GitLab Silver/Premium and above.

See: https://docs.gitlab.com/ee/api/projects.html#get-project-push-rules

This commit changes the push rules response struct to use type `*bool` instead of `bool` for those rules, since they aren't returned by the API if you don't have the right plan. Leaving the type as `bool` makes it hard to tell if trying to set those rules will succeed or return a `403 Forbidden` error.

### DIFF
--- a/groups.go
+++ b/groups.go
@@ -521,8 +521,10 @@ type GroupPushRules struct {
 	AuthorEmailRegex           string     `json:"author_email_regex"`
 	FileNameRegex              string     `json:"file_name_regex"`
 	MaxFileSize                int        `json:"max_file_size"`
-	CommitCommitterCheck       bool       `json:"commit_committer_check"`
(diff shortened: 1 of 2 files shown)


svanharmelen:
@jtdoepke I can see why you created this PR, but I will not be merging this one. The current assumption is that the users of this package (i.e. the GitLab API) know what they are doing and understand which fields are applicable for them.

Currently the response structs are consistent in their configuration, which makes if very easy to use them. Making them have both pointer and non-pointer fields for the same types (so a `bool` and ` *bool` for example) isn't an improvement and would still require special handling of the fields (do a nil check first). So I don't see the benefit and will close this one. Sorry!

PR #999, opened 2020-12-09
Add exclude_internal option to ListUsersOptions

### LINKED ISSUE #997: Add support for exclude bot users
From the GitLab documentation:
GitLab supports bot users such as the alert bot or the support bot. To exclude these users from the users’ list, you can use the parameter exclude_internal=true (introduced in GitLab 13.4).

`GET /users?exclude_internal=true`

GitLab doc xref: https://gitlab.com/gitlab-org/gitlab/-/issues/241144

alex1989hu:
GitLab supports bot users such as the alert bot or the support bot.

To exclude these users from the users’ list, you can use the parameter exclude_internal=true.

Feature is introduced in GitLab 13.4.

Fixes #997

Note:

```
As of now, GitLab 13.6.2-ee (98aab73cbd5) Users API does not expose the field exclude_internal
```

### DIFF
--- a/users.go
+++ b/users.go
@@ -102,8 +102,9 @@ type UserIdentity struct {
 // GitLab API docs: https://docs.gitlab.com/ce/api/users.html#list-users
 type ListUsersOptions struct {
 	ListOptions
-	Active  *bool `url:"active,omitempty" json:"active,omitempty"`
-	Blocked *bool `url:"blocked,omitempty" json:"blocked,omitempty"`
+	Active          *bool `url:"active,omitempty" json:"active,omitempty"`
+	Blocked         *bool `url:"blocked,omitempty" json:"blocked,omitempty"`
+	ExcludeInternal *bool `url:"exclude_internal,omitempty" json:"exclude_internal,omitempty"`
 
 	// The options below are only available for admins.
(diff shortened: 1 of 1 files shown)


users.go
```diff
@@ -115,6 +115,7 @@ type ListUsersOptions struct {
 	OrderBy              *string    `url:"order_by,omitempty" json:"order_by,omitempty"`
 	Sort                 *string    `url:"sort,omitempty" json:"sort,omitempty"`
 	External             *bool      `url:"external,omitempty" json:"external,omitempty"`
+	ExcludeInternal      *bool      `url:"exclude_internal,omitempty" json:"exclude_internal,omitempty"`
```
svanharmelen, line 118: This looks good, but it's an option that is also available for normal users. So this should be added below line 106 I guess?

svanharmelen (changes requested):

svanharmelen (approved):
Thanks quick! Thanks @alex1989hu, LGTM!

users.go
alex1989hu: You are right, I can verify that the GitLab documentation also says it is also available for https://docs.gitlab.com/ee/api/users.html#for-normal-users not just for admins. Force pushed the branch

PR #1001, opened 2020-12-09
Add support for using_license_seat field in users

### LINKED ISSUE #1000: Add support for using_license_seat field in users
From GitLab documentation:
Users on GitLab Starter, Bronze, or higher also see the shared_runners_minutes_limit, extra_shared_runners_minutes_limit, and **using_license_seat parameters**.

alex1989hu:
Use type *bool instead of bool because this field is not returned by GitLab API.

This field is only available in GitLab Starter, Bronze, or higher.

Fixes #1000 


users.go
```diff
@@ -89,6 +89,9 @@ type User struct {
 	PrivateProfile            bool               `json:"private_profile"`
 	SharedRunnersMinutesLimit int                `json:"shared_runners_minutes_limit"`
 	CustomAttributes          []*CustomAttribute `json:"custom_attributes"`
+
+	// UsingLicenseSeat is only available in GitLab Starter, Bronze, or higher.
+	UsingLicenseSeat *bool `json:"using_license_seat,omitempty"`
```
svanharmelen, line 94: Maybe also add `extra_shared_runners_minutes_limit` while you are at it?

But also, please do not make a distinction for certain fields by making them a pointer field. See my comment here #998 as well. So please make it normal fields and place them below `SharedRunnersMinutesLimit` following the order in the docs.

Thanks!

svanharmelen (changes requested):
Next to the inline comment also a small FYI; no need to create an issue first. A descriptive PR is just fine for me 👍🏻

users.go
alex1989hu: I see your point, to be honest am little bit confused: some codes use pointer field. Sure, I can update this.

Do we want to have `omitempty` at `UsingLicenseSeat` field?

```go
	SharedRunnersMinutesLimit      int                `json:"shared_runners_minutes_limit"`
	ExtraSharedRunnersMinutesLimit int                `json:"extra_shared_runners_minutes_limit"`
	UsingLicenseSeat               bool               `json:"using_license_seat,omitempty"`
```

svanharmelen: Nope, no `omitempty` for those please.

As for what should be a pointer and what should not. In options structs everything should be a pointer (so we can use them properly when making API calls) and in response structs only struct fields are pointers. Hope that helps...

alex1989hu: Gotcha, thank you for your explanation. Updated the code as you rested - force pushed the branch.

svanharmelen (approved):
Thanks! LGTM 👍🏻

PR #1004, opened 2020-12-14
Add a new type of API authentication: multiple

christophedcpm:
This mode allow to use oauth and private token at the same time.
Both the `Authorization: Bearer ` and `PRIVATE-TOKEN` header are set in
the requests sent to the server


svanharmelen (changes requested):
Not sure if I understand the usecase for this one. When would you be in a situation that you don't know which client to use?

Additionally I noticed that you added a lot of import rewrites, I'm not going to merge those of course 😉

christophedcpm:
Woops sorry, my bad, this PR was not meant for repo just yet. I might open an issue to disscuss the use case later on. Right now I just needed a quick hack to test something.

PR #1010, opened 2020-12-18
Add run_untagged field to RunnerDetails

### LINKED ISSUE #1009: run_untagged field missing in runner details response
When we issue the call from here: https://docs.gitlab.com/ee/api/runners.html#get-runners-details , we go through

```go
func (s *RunnersService) GetRunnerDetails(rid interface{}, options ...RequestOptionFunc) (*RunnerDetails, *Response, error) {
```

In this, there's no `RunUntagged` field. However, if we issue the same request directly via curl:

``` console
$ curl --header "PRIVATE-TOKEN: …" http://gitlab.…/api/v4/runners/175 | jq | grep run_untagged
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1958  100  1958    0     0   2971      0 --:--:-- --:--:-- --:--:--  2971
  "run_untagged": true,
```

I appreciate that this field is not documented on the GitLab page but it's part of the response and should probably be exposed via the SDK.

If we look at the GitLab source code, we can see what the actual fields are https://gitlab.com/gitlab-org/gitlab/-/blob/1cf69f5dd38398a3ee7689400272cfbcd45bb56d/lib/api/entities/runner_details.rb#L7

Fuuzetsu:
Fixes #1009

### DIFF
--- a/runners.go
+++ b/runners.go
@@ -69,6 +69,7 @@ type RunnerDetails struct {
 	Token          string   `json:"token"`
 	Revision       string   `json:"revision"`
 	TagList        []string `json:"tag_list"`
+	RunUntagged    bool     `json:"run_untagged"`
 	Version        string   `json:"version"`
 	Locked         bool     `json:"locked"`
 	AccessLevel    string   `json:"access_level"`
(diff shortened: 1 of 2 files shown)


svanharmelen (approved):
Thanks @Fuuzetsu! LGTM 👍🏻

PR #1013, opened 2020-12-21
Support ID field in contribution events

hansmi:
GitLab events also have a unique ID.

https://gitlab.com/gitlab-org/gitlab/-/blob/a523866d205c89f8a7fe94ca45be407f224c56e7/lib/api/entities/event.rb#L6


svanharmelen (approved):
Thanks @hansmi 👍🏻

PR #1014, opened 2020-12-22
Add info struct to runner registration.

### LINKED ISSUE #1003: Should RegisterNewRunnerOptions.Info not be String?
We have https://github.com/xanzy/go-gitlab/blob/d6dd0aaafbf3c25a877842511a52e5022d44d249/runners.go#L385

But in the API this is a `hash` with some specific fields: https://docs.gitlab.com/ee/api/runners.html#register-a-new-runner

The field type is either wrong or I am failing to figure out what kind of encoding is expected to create the correct value for this field.

The expected form encoding is shown here https://docs.gitlab.com/ee/api/README.html#hash . For some other structs I looked at, for cases where objects where expected, separate structs were created.

Fuuzetsu:
Fixes #1003

### DIFF
--- a/runners.go
+++ b/runners.go
@@ -381,14 +381,27 @@ func (s *RunnersService) ListGroupsRunners(gid interface{}, opt *ListGroupsRunne
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/runners.html#register-a-new-runner
 type RegisterNewRunnerOptions struct {
-	Token          *string  `url:"token" json:"token"`
-	Description    *string  `url:"description,omitempty" json:"description,omitempty"`
-	Info           *string  `url:"info,omitempty" json:"info,omitempty"`
-	Active         *bool    `url:"active,omitempty" json:"active,omitempty"`
-	Locked         *bool    `url:"locked,omitempty" json:"locked,omitempty"`
-	RunUntagged    *bool    `url:"run_untagged,omitempty" json:"run_untagged,omitempty"`
-	TagList        []string `url:"tag_list[],omitempty" json:"tag_list,omitempty"`
-	MaximumTimeout *int     `url:"maximum_timeout,omitempty" json:"maximum_timeout,omitempty"`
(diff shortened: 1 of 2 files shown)


runners.go
```diff
@@ -375,20 +375,33 @@ func (s *RunnersService) ListGroupsRunners(gid interface{}, opt *ListGroupsRunne
 	return rs, resp, err
 }
 
+// RegisterNewRunnerInfo represents the info hashmap parameter in
+// RegisterNewRunnerOptions.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/runners.html#register-a-new-runner
+type RegisterNewRunnerInfo struct {
```
svanharmelen, line 383: All option struct end with the word `Options`, so this one should too. Not sure what the best name would be in this case, but maybe something like `NewRunnerMetadataOptions`? Of `NewRunnerInfoOptions`?

runners.go
```diff
@@ -375,20 +375,33 @@ func (s *RunnersService) ListGroupsRunners(gid interface{}, opt *ListGroupsRunne
...
+	Name         *string `url:"name,omitempty" json:"name,omitempty"`
+	Version      *string `url:"version,omitempty" json:"version,omitempty"`
+	Revision     *string `url:"revision,omitempty" json:"revision,omitempty"`
+	Platform     *string `url:"platform,omitempty" json:"platform,omitempty"`
+	Architecture *string `url:"architecture,omitempty" json:"architecture,omitempty"`
+}
+
```
svanharmelen, line 390: Can you please place this struct below the `RegisterNewRunnerOptions` struct declaration? I general we always start with the main struct and the list any addition structs. Thanks!

svanharmelen (changes requested):
Looks good @Fuuzetsu! Just two small comments about naming and styling (what else 😏)...

svanharmelen (approved):
Thanks @Fuuzetsu! LGTM!

PR #1015, opened 2020-12-22
Add golangci-lint

alex1989hu:
Add golangci-lint

Starting with minimimal, permissive configuration
to do not fail CI build and let us able to follow
code changes

These are the most minimalistic linters which are good candidate to introduce `golangci-lint`:
- Fix govet errors reported by golangci-lint
- Fix goconst errors reported by golangci-lint
- Fix staticcheck errors reported by golangci-lint
- Fix misspell errors reported by golangci-lint
- Fix whitespace errors reported by golangci-lint


svanharmelen (changes requested):
Before continuing my review, lets exclude the `errcheck` and with that all the `//nolint:errcheck` comments. I'm open to adding the rest of the checks...

Also lets define all the `exampleXXX` constants in one location. I would suggest adding them just below the imports in `gitlab_test.go`.

access_requests_test.go
```diff
@@ -57,6 +57,7 @@ func TestListProjectAccessRequests(t *testing.T) {
 	}
 
 	requests, resp, err := client.AccessRequests.ListProjectAccessRequests(1, nil)
+	assert.NotNil(t, resp)
```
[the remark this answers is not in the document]
alex1989hu: Good catch, rewrote the order 👍

alex1989hu:
> Before continuing my review, lets exclude the `errcheck` and with that all the `//nolint:errcheck` comments. I'm open to adding the rest of the checks...
> 
> Also lets define all the `exampleXXX` constants in one location. I would suggest adding them just below the imports in `gitlab_test.go`.

To be honest, I do not clearly understand what you asked me to do. Please tell me which one you thought
- disable linter OR
- modify the code to check if any error

For the second part, I moved all constant - I have just pushed commit: **Move constants for test to common place**

svanharmelen:
I mean to disable (remove) the `errcheck` linter. And when we don't use the `errcheck` linter, we can also remove all the `//nolint:errcheck` comments again.

gitlab_test.go
```diff
@@ -16,6 +16,228 @@ import (
...
+	"access_level": "ref_protected",
+	"maximum_timeout": 3600,
+	"locked": false
+}`
+
+	exampleRegisterNewRunner = `{
+	"id": 12345,
+	"token": "6337ff461c94fd3fa32ba3b1ff4125"
+}`
+)
```
svanharmelen, line 239: Ow, this is quite a list. Didn't expect that to be honest 😊 In this case then maybe we should move this into it's own file? `fixtures_test.go` or something like that?

alex1989hu: Yes, it is. To be honest, I would keep them where they are and drop my last commit. It is better than moving all of them into one file. Thoughts?

svanharmelen: I think its better to group them in one location as otherwise you have to do a search each time you want to look at one of them. And finding them in somewhat random locations could be confusing...

alex1989hu: Lets see now, please

alex1989hu:
> I mean to disable (remove) the `errcheck` linter. And when we don't use the `errcheck` linter, we can also remove all the `//nolint:errcheck` comments again.

In the meantime I fixed them and write a summary. There are only two items which needs rework but it requires to have deeper knowledge of this library.

If you are still unhappy with the change, I can create an issue and put reference after `nolint`, but on the other side I fixed most of them.

Updated the PR and fixed them, here is some description:
- ./gitlab.go:392:			if reset, _ := strconv.ParseInt(v, 10, 64); reset > 0 { //nolint:errcheck
   - **This requires rework.**, because it will end up in `func newClient(options ...ClientOptionFunc) (*Client, error)`: `Backoff:      c.retryHTTPBackoff,`
- ./gitlab.go:627:	c.configureLimiterOnce.Do(func() { c.configureLimiter() }) //nolint:errcheck
   - *Intended, see the already existing comment in the code*
- ./gitlab.go:674:	defer resp.Body.Close() //nolint:errcheck
   - *Intended, no other option.*
- ./gitlab.go:754:	path, _ := url.QueryUnescape(e.Response.Request.URL.Path) //nolint:errcheck
   - **This requires rework.**
- ./projects_test.go:298:	defer os.Remove(tf.Name()) //nolint:errcheck
   - *Intended, no other option.*
- ./access_requests_test.go:251:		defer r.Body.Close() //nolint:errcheck
   - *Intended, no other option.*
- ./access_requests_test.go:311:		defer r.Body.Close() //nolint:errcheck
   - *Intended, no other option.*
- ./projects.go:1141:	defer f.Close() //nolint:errcheck
   - *Intended, no other option.*

svanharmelen:
Sorry @alex1989hu, but I don't like this approach. I appreciate that you want to add some checks to make the package better tested, but let's not change the way this package is written.

I think the only places were the errors are ignored, is where they can actually be ignored without any danger of it causing issues for users of this package. So please just remove the error check and revert all related changes.

As for the location of the constants, lets see how it looks when the rest is reverted and the diff is a bit cleaner again.

Thanks!

alex1989hu:
> Sorry @alex1989hu, but I don't like this approach. I appreciate that you want to add some checks to make the package better tested, but let's not change the way this package is written.
> 
> I think the only places were the errors are ignored, is where they can actually be ignored without any danger of it causing issues for users of this package. So please just remove the error check and revert all related changes.
> 
> As for the location of the constants, lets see how it looks when the rest is reverted and the diff is a bit cleaner again.
> 
> Thanks!

Done 👍

alex1989hu:
Hello @svanharmelen: resolved merge conflict, rebased the branch. Is there anything I could do as an enhancement of this PR?

svanharmelen (approved):
Thanks @alex1989hu! Sorry for the delay on this one... I made a few final tweaks and am happy with how things look now 👍🏻

PR #1016, opened 2020-12-23
Bridge: Make Downstream Pipeline optional

sluongng:
In case pipelines with a bridge(trigger) job but the downstream pipeline
has yet been scheduled, the downstream object will be 'null'.

Make DownstreamPipeline in Bridge struct a pointer instead of value to
allow null to be detected.

### DIFF
--- a/jobs.go
+++ b/jobs.go
@@ -98,21 +98,13 @@ type Bridge struct {
 		UpdateAt  *time.Time `json:"updated_at"`
 		WebURL    string     `json:"web_url"`
 	} `json:"pipeline"`
-	Ref                string `json:"ref"`
-	Stage              string `json:"stage"`
-	Status             string `json:"status"`
-	Tag                bool   `json:"tag"`
-	WebURL             string `json:"web_url"`
-	User               *User  `json:"user"`
-	DownstreamPipeline *struct {
-		ID        int        `json:"id"`
-		Sha       string     `json:"sha"`
-		Ref       string     `json:"ref"`
-		Status    string     `json:"status"`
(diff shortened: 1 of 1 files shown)


jobs.go
```diff
@@ -104,7 +104,7 @@ type Bridge struct {
 	Tag                bool   `json:"tag"`
 	WebURL             string `json:"web_url"`
 	User               *User  `json:"user"`
-	DownstreamPipeline struct {
+	DownstreamPipeline *struct {
```
svanharmelen, line 107: If we make this a pointer, we should make a named struct for this (to keep the code consistent). Make define:

```
type PipelineInfo struct {
	ID        int        `json:"id"`
	Ref       string     `json:"ref"`
	Sha       string     `json:"sha"`
	Status    string     `json:"status"`
	UpdateAt  *time.Time `json:"updated_at"`
	WebURL    string     `json:"web_url"`
}
```

Which can then be used for the `Pipeline` and `DownstreamPipeline` fields?

sluongng: Thats a great suggestion. I have pushed an additional commit to make use of `PipelineInfo` in pipeline.go

svanharmelen: Cool... I even forgot that it already existed 😂 Could you also use the same struct for the `Pipeline` field to keep it consistent? Thanks!

sluongng: Added another commit do replace `Pipeline` field, but I kept it as value instead of switching it to pointer.

svanharmelen (approved):
Thanks @sluongng! LGTM!

PR #1017, opened 2020-12-25
Validate the CI YAML configuration with a namespace

madest92:


validate.go
```diff
@@ -81,3 +81,35 @@ func (s *ValidateService) ProjectLint(pid interface{}, opt *ProjectLintOptions,
...
+	}
+
+	l := new(ProjectLintResult)
+	resp, err := s.client.Do(req, l)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return l, resp, nil
+}
```
svanharmelen, line 115: Please place the new `ProjectNamespaceLint` blocks above the `ProjectLint` blocks so the order in inline with the docs (I try to use those to keep some kind of logic in the ordering).

svanharmelen (changes requested):
Looks good @madest92, just one tiny request to get it ready to merge...

validate.go
```diff
@@ -81,3 +81,35 @@ func (s *ValidateService) ProjectLint(pid interface{}, opt *ProjectLintOptions,
 
 	return l, resp, nil
 }
+
+// ProjectNamespaceLintOptions represents the available ProjectNamespaceLint() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/lint.html#validate-a-ci-yaml-configuration-with-a-namespace
```
svanharmelen, line 87:
```suggestion
// ProjectNamespaceLintOptions represents the available ProjectNamespaceLint()
// options.
//
// GitLab API docs:
// https://docs.gitlab.com/ee/api/lint.html#validate-a-ci-yaml-configuration-with-a-namespace
```

validate.go
```diff
@@ -81,3 +81,35 @@ func (s *ValidateService) ProjectLint(pid interface{}, opt *ProjectLintOptions,
...
+type ProjectNamespaceLintOptions struct {
+	Content *string `url:"content,omitempty" json:"content,omitempty"`
+	DryRun  *bool   `url:"dry_run,omitempty" json:"dry_run,omitempty"`
+}
+
+// ProjectNamespaceLint validates .gitlab-ci.yml content by project.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/lint.html#validate-a-ci-yaml-configuration-with-a-namespace
```
svanharmelen, line 95:
```suggestion
// GitLab API docs:
// https://docs.gitlab.com/ee/api/lint.html#validate-a-ci-yaml-configuration-with-a-namespace
```

svanharmelen (changes requested):
Ow, and these two tweaks 😏

madest92:
@svanharmelen  fixed :)

validate.go
```diff
@@ -5,22 +5,25 @@ import "fmt"
 // ValidateService handles communication with the validation related methods of
 // the GitLab API.
 //
-// GitLab API docs: https://docs.gitlab.com/ce/api/lint.html
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/lint.html
```
svanharmelen, line 9: Thanks for trying to make things consistent! Yet they do not all need to be formatted like this. It's bases on line length. The target for doc lines is to keep it below 80 chars... Of course a URL can still be longer, but we should not break in the middle of a URL so that is fine as long as these longer URLs start on their own line.

Yet this one is short enough, so this can just be placed on one line.

validate.go
```diff
@@ -5,22 +5,25 @@ import "fmt"
...
 type ValidateService struct {
 	client *Client
 }
 
 // LintResult represents the linting results.
 //
-// GitLab API docs: https://docs.gitlab.com/ce/api/lint.html
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/lint.html
```
svanharmelen, line 17: As above, please revert this one.

validate.go
```diff
@@ -30,7 +33,8 @@ type ProjectLintResult struct {
 
 // Lint validates .gitlab-ci.yml content.
 //
-// GitLab API docs: https://docs.gitlab.com/ce/api/lint.html
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/lint.html
```
svanharmelen, line 37: This one can also be reverted...

svanharmelen (commented):
Almost there 😉

validate.go
madest92: Ha-ha :) Yes, I understand. Okay, I'll fix it.

svanharmelen (approved):
Perfect, thanks @madest92!

PR #1018, opened 2020-12-25
Add restore group

markuslackner:
Adds restore group (only available in silver/premium) to groups API


groups.go
```diff
@@ -656,3 +656,28 @@ func (s *GroupsService) DeleteGroupPushRule(gid interface{}, options ...RequestO
...
+	}
+
+	g := new(Group)
+	resp, err := s.client.Do(req, g)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return g, resp, nil
+}
```
svanharmelen, line 683: Please place this new method in the order following the order in the docs. So that would be somewhere on line 268. And also please make sure the file ends with a newline. Thanks!

svanharmelen (changes requested):
Looks good @markuslackner, just needs to be reordering and then it can be merged.

groups.go
markuslackner: Reordered  `RestoreGroup` and added fixed missing newline in `groups_test.go`

svanharmelen (approved):
That was quick 😃 Thanks @markuslackner , LGTM!

PR #1021, opened 2021-01-04
Added missing FreezePeriodService

zachkknowbe4:
Went to use the freeze period api and noticed the service was missing, this adds the missing service to gitlab.go.

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -117,7 +117,7 @@ type Client struct {
 	Epics                 *EpicsService
 	Events                *EventsService
 	Features              *FeaturesService
-	FreezePeriod          *FreezePeriodsService
+	FreezePeriods         *FreezePeriodsService
 	GitIgnoreTemplates    *GitIgnoreTemplatesService
 	GroupBadges           *GroupBadgesService
 	GroupCluster          *GroupClustersService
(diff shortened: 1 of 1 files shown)


gitlab.go
```diff
@@ -117,6 +117,7 @@ type Client struct {
 	Epics                 *EpicsService
 	Events                *EventsService
 	Features              *FeaturesService
+	FreezePeriod          *FreezePeriodsService
```
svanharmelen, line 120: In order for this to work as expected, you also need to add this to line ~284:

```go
c.FreezePeriod = &FreezePeriodsService{client: c}
```

svanharmelen (changes requested):
Thanks @zachkknowbe4... But it needs another line for it all to work as expected.

gitlab.go
zachkknowbe4: I've updated the file with the correct naming and added the requested line.

svanharmelen (approved):
Thanks @zachkknowbe4! LGTM 👍🏻

zachkknowbe4:
@svanharmelen Is it possible this could get released today? We need it for https://github.com/gitlabhq/terraform-provider-gitlab/pull/516 to be completed and merged.

Thanks.

PR #1024, opened 2021-01-06
fix(users.go): fix json unmarshal error of usermembership accesslevel

feiguoL:
Fix json unmarshal error of usermembership accesslevel，the gitlab api response accesslevel int type, when unmarshal response will return 400

### DIFF
--- a/users.go
+++ b/users.go
@@ -942,10 +942,10 @@ func (s *UsersService) SetUserStatus(opt *UserStatusOptions, options ...RequestO
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/users.html#user-memberships-admin-only
 type UserMembership struct {
-	SourceID    int    `json:"source_id"`
-	SourceName  string `json:"source_name"`
-	SourceType  string `json:"source_type"`
-	AccessLevel string `json:"access_level"`
+	SourceID    int              `json:"source_id"`
+	SourceName  string           `json:"source_name"`
+	SourceType  string           `json:"source_type"`
+	AccessLevel AccessLevelValue `json:"access_level"`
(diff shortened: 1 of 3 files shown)


users.go
```diff
@@ -945,7 +945,7 @@ type UserMembership struct {
 	SourceID    int    `json:"source_id"`
 	SourceName  string `json:"source_name"`
 	SourceType  string `json:"source_type"`
-	AccessLevel string `json:"access_level"`
+	AccessLevel int    `json:"access_level"`
```
svanharmelen, line 948: The docs say this is a string, are you sure this is an integer? If it really is an integer, we should probably use the [`AccessLevelValue`](https://github.com/xanzy/go-gitlab/blob/master/types.go#L38) type here?

feiguoL: you can check this type, when i use AccessLevel with string the response is 400 by 'can not unmarshal num to string'

when i query gitlab code, the struct is define with number type
[project_member.rb](https://gitlab.com/gitlab-org/gitlab-foss/-/blob/master/qa/qa/resource/project_member.rb)

svanharmelen: Check... I'm cool with it, but the docs also give an example response that clearly shows a number returned as a string:

image

I'm not using GitLab myself that much lately, but I trust your test results. Which version of GitLab are you using? And in that case I guess we should use the `AccountLevelValue` as the type like I suggested. Agree?

feiguoL: The version of GitLab that show v13.7.0. I agree with your suggestion， thanks!

svanharmelen (approved):
Thanks @feiguoL! LGTM!

PR #1026, opened 2021-01-09
Return a *bytes.Reader instead of an io.Reader

hartzell:
Change GetJobArtifacts, DownloadArtifactsFile, DownloadSingleArtifactsFile, and GetTraceFile to return a *bytes.Reader instead of an io.Reader.

This change makes it simple to e.g. unzip archive files without needing to first write the data to a file.

I don't see any applicable tests to touch up, I've tested it in a personal application that uses GetJobArtifacts to download and unpack artifact archive files and it works as expected.

Closes: #1023

Issue #970, opened 2020-11-11
Refactor authentication?


firelizzard18:
Would the team be open to a PR refactoring the request authentication system?

Currently, there are 3 authentication types: private-token header token, HTTP bearer token (oauth), and HTTP basic -> bearer. These represent a small subset of the mechanisms supported by GitLab, and more importantly, none of them work with GitLab CI tokens: CI tokens can be passed via 1) the `JOB-TOKEN` header, 2) the `job_token` query variable, or 3) HTTP basic auth. However, CI tokens cannot be used to create an OAuth token, so they cannot be used with `NewBasicAuthClient`.

I propose a generic authentication callback that does whatever it needs to the request:

```golang
package gitlab

import (
	"context"
	"encoding/base64"
	"errors"
	"net/http"
	"strings"
	"sync"

	"github.com/hashicorp/go-retryablehttp"
	"golang.org/x/oauth2"
)

var ErrNoRetry = errors.New("do not retry")

func NewClientWithAuthenticator(auth Authenticator, options ...ClientOptionFunc) (*Client, error) {
	client, err := newClient(options...)
	if err != nil {
		return nil, err
	}

	client.authenticator = auth
	return client, nil
}

type Authenticator interface {
	Apply(*Client, *retryablehttp.Request) error
	Retry(*Client, *http.Response) error
}

func HeaderTokenAuth(header, token string) Authenticator {
	return &headerAuthenticator{header, token}
}

func PrivateTokenAuth(token string) Authenticator {
	return HeaderTokenAuth("PRIVATE-TOKEN", token)
}

func JobTokenAuth(token string) Authenticator {
	return HeaderTokenAuth("JOB-TOKEN", token)
}

func DeployTokenAuth(token string) Authenticator {
	return HeaderTokenAuth("DEPLOY-TOKEN", token)
}

func OAuth(token string) Authenticator {
	return HeaderTokenAuth("Authorization", "Bearer "+token)
}

func BasicAuth(username, password string) Authenticator {
	return HeaderTokenAuth("Authorization", "Basic "+base64.StdEncoding.EncodeToString([]byte(username+":"+password)))
}

func BasicPlusOAuth(username, password string) Authenticator {
	return &oauthPlusBasicAuthenticator{
		username: username,
		password: password,
	}
}

type headerAuthenticator struct {
	header, token string
}

func (a *headerAuthenticator) Apply(_ *Client, req *retryablehttp.Request) error {
[...]

PR #1027, opened 2021-01-09
A more flexible solution for configuring authentication

svanharmelen:


svanharmelen:
Inspired by both #969, #970 (@firelizzard18) and #1026 (@RangerCD), would this be a good way forward?

Still in doubt if (for backwards compatibility) we should leave the old `NewClient` function in tact and replace only the others with a single new client function. Yet personally I prefer to refactor them all into one new function. Makes it pretty clean I think and uses a little bit of both suggested solutions.

firelizzard18:
@svanharmelen This is definitely a step forward, but I would still prefer something that allows me to take control of the authentication process.

RangerCD:
@svanharmelen This looks good enough to me :+1: .

svanharmelen:
I think that by now we support all the tokens you tried to support using this PR. I understand it's a bit less flexible and might have a tiny bit of duplicate code, but for now I'm not really willing to make any big changes around this point so I will not be merging this PR...

PR #1033, opened 2021-01-14
Add ContainerExpirationPolicyAttributes

cakoolen:
The project API contains settings for changing the container expiration policy. 

This PR implements these settings for both the create / edit project as well as for retrieving the project settings.

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -38,71 +38,72 @@ type ProjectsService struct {
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/projects.html
 type Project struct {
-	ID                                        int                `json:"id"`
-	Description                               string             `json:"description"`
-	DefaultBranch                             string             `json:"default_branch"`
-	Public                                    bool               `json:"public"`
-	Visibility                                VisibilityValue    `json:"visibility"`
-	SSHURLToRepo                              string             `json:"ssh_url_to_repo"`
-	HTTPURLToRepo                             string             `json:"http_url_to_repo"`
-	WebURL                                    string             `json:"web_url"`
-	ReadmeURL                                 string             `json:"readme_url"`
-	TagList                                   []string           `json:"tag_list"`
-	Owner                                     *User              `json:"owner"`
-	Name                                      string             `json:"name"`
-	NameWithNamespace                         string             `json:"name_with_namespace"`
-	Path                                      string             `json:"path"`
-	PathWithNamespace                         string             `json:"path_with_namespace"`
-	IssuesEnabled                             bool               `json:"issues_enabled"`
-	OpenIssuesCount                           int                `json:"open_issues_count"`
-	MergeRequestsEnabled                      bool               `json:"merge_requests_enabled"`
-	ApprovalsBeforeMerge                      int                `json:"approvals_before_merge"`
-	JobsEnabled                               bool               `json:"jobs_enabled"`
-	WikiEnabled                               bool               `json:"wiki_enabled"`
-	SnippetsEnabled                           bool               `json:"snippets_enabled"`
-	ResolveOutdatedDiffDiscussions            bool               `json:"resolve_outdated_diff_discussions"`
-	ContainerRegistryEnabled                  bool               `json:"container_registry_enabled"`
-	CreatedAt                                 *time.Time         `json:"created_at,omitempty"`
-	LastActivityAt                            *time.Time         `json:"last_activity_at,omitempty"`
-	CreatorID                                 int                `json:"creator_id"`
-	Namespace                                 *ProjectNamespace  `json:"namespace"`
-	ImportStatus                              string             `json:"import_status"`
-	ImportError                               string             `json:"import_error"`
-	Permissions                               *Permissions       `json:"permissions"`
-	MarkedForDeletionAt                       *ISOTime           `json:"marked_for_deletion_at"`
-	Archived                                  bool               `json:"archived"`
-	AvatarURL                                 string             `json:"avatar_url"`
-	SharedRunnersEnabled                      bool               `json:"shared_runners_enabled"`
-	ForksCount                                int                `json:"forks_count"`
-	StarCount                                 int                `json:"star_count"`
-	RunnersToken                              string             `json:"runners_token"`
-	PublicBuilds                              bool               `json:"public_builds"`
-	AllowMergeOnSkippedPipeline               bool               `json:"allow_merge_on_skipped_pipeline"`
-	OnlyAllowMergeIfPipelineSucceeds          bool               `json:"only_allow_merge_if_pipeline_succeeds"`
-	OnlyAllowMergeIfAllDiscussionsAreResolved bool               `json:"only_allow_merge_if_all_discussions_are_resolved"`
-	RemoveSourceBranchAfterMerge              bool               `json:"remove_source_branch_after_merge"`
-	LFSEnabled                                bool               `json:"lfs_enabled"`
-	RequestAccessEnabled                      bool               `json:"request_access_enabled"`
-	MergeMethod                               MergeMethodValue   `json:"merge_method"`
-	ForkedFromProject                         *ForkParent        `json:"forked_from_project"`
-	Mirror                                    bool               `json:"mirror"`
-	MirrorUserID                              int                `json:"mirror_user_id"`
-	MirrorTriggerBuilds                       bool               `json:"mirror_trigger_builds"`
-	OnlyMirrorProtectedBranches               bool               `json:"only_mirror_protected_branches"`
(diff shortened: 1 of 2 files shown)


projects.go
```diff
@@ -464,59 +465,82 @@ func (s *ProjectsService) GetProjectEvents(pid interface{}, opt *GetProjectEvent
...
+// ProjectContainerExpirationPolicy represents the container expiration attributes for a project.
+type ProjectContainerExpirationPolicy struct {
+	Cadence         string    `json:"cadence"`
+	KeepN           uint      `json:"keep_n"`
+	OlderThan       string    `json:"older_than"`
+	NameRegex       string    `json:"name_regex"` // NameRegex will be replaced by NameRegexDelete but now calls without setting this field are not accepted.
+	NameRegexDelete string    `json:"name_regex_delete"`
+	NameRegexKeep   string    `json:"name_regex_keep"`
+	Enabled         bool      `json:"enabled"`
+	NextRunAt       time.Time `json:"next_run_at"`
```
svanharmelen, line 477:
```suggestion
	NextRunAt       *time.Time `json:"next_run_at"`
```

projects.go
```diff
@@ -464,59 +465,82 @@ func (s *ProjectsService) GetProjectEvents(pid interface{}, opt *GetProjectEvent
 	return p, resp, err
 }
 
+// ProjectContainerExpirationPolicy represents the container expiration attributes for a project.
+type ProjectContainerExpirationPolicy struct {
+	Cadence         string    `json:"cadence"`
+	KeepN           uint      `json:"keep_n"`
+	OlderThan       string    `json:"older_than"`
+	NameRegex       string    `json:"name_regex"` // NameRegex will be replaced by NameRegexDelete but now calls without setting this field are not accepted.
```
svanharmelen, line 473:
```suggestion
```

This struct is used to marshal the response, so it should not matter that this field doesn't exist.

projects.go
```diff
@@ -464,59 +465,82 @@ func (s *ProjectsService) GetProjectEvents(pid interface{}, opt *GetProjectEvent
 	return p, resp, err
 }
 
+// ProjectContainerExpirationPolicy represents the container expiration attributes for a project.
```
svanharmelen, line 468:
```suggestion
// ProjectContainerExpirationPolicy represents the container expiration attributes for a project.
//
// GitLab API docs: https://docs.gitlab.com/ce/api/projects.html
```

projects.go
```diff
@@ -464,59 +465,82 @@ func (s *ProjectsService) GetProjectEvents(pid interface{}, opt *GetProjectEvent
...
+	KeepN           uint      `json:"keep_n"`
+	OlderThan       string    `json:"older_than"`
+	NameRegex       string    `json:"name_regex"` // NameRegex will be replaced by NameRegexDelete but now calls without setting this field are not accepted.
+	NameRegexDelete string    `json:"name_regex_delete"`
+	NameRegexKeep   string    `json:"name_regex_keep"`
+	Enabled         bool      `json:"enabled"`
+	NextRunAt       time.Time `json:"next_run_at"`
+}
+
+// ProjectContainerExpirationPolicyAttributes represents the container expiration attributes for a project.
```
svanharmelen, line 480:
```suggestion
// ProjectContainerExpirationPolicyAttributes represents the container expiration attributes for a project.
//
// GitLab API docs: https://docs.gitlab.com/ce/api/projects.html
```

projects.go
```diff
@@ -464,59 +465,82 @@ func (s *ProjectsService) GetProjectEvents(pid interface{}, opt *GetProjectEvent
...
+type ProjectContainerExpirationPolicyAttributes struct {
```
svanharmelen, line 481: I would rename this one too:
```suggestion
type ContainerExpirationPolicyOptions struct {
```
Changing the `Attributes` part to `Options` is for keeping things consistent.

projects.go
```diff
@@ -464,59 +465,82 @@ func (s *ProjectsService) GetProjectEvents(pid interface{}, opt *GetProjectEvent
...
+	Cadence         string `json:"cadence"`
+	KeepN           uint   `json:"keep_n"`
+	OlderThan       string `json:"older_than"`
+	NameRegex       string `json:"name_regex"` // NameRegex will be replaced by NameRegexDelete but now calls without setting this field are not accepted.
+	NameRegexDelete string `json:"name_regex_delete"`
+	NameRegexKeep   string `json:"name_regex_keep"`
+	Enabled         bool   `json:"enabled"`
+}
```
svanharmelen, line 489: I think these fields all need to be pointers and have `omitempty` added to the tags. Additionally please also add the `url` tags to make sure this new code is consistent.

svanharmelen (changes requested):
Thanks @cakoolen! Do have a few requests we need to look at before merging this one...

projects.go
```diff
@@ -464,59 +465,82 @@ func (s *ProjectsService) GetProjectEvents(pid interface{}, opt *GetProjectEvent
 	return p, resp, err
 }
 
+// ProjectContainerExpirationPolicy represents the container expiration attributes for a project.
+type ProjectContainerExpirationPolicy struct {
```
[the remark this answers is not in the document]
cakoolen: I can see where that is coming from. Renamed as suggested.

projects.go
```diff
@@ -464,59 +465,82 @@ func (s *ProjectsService) GetProjectEvents(pid interface{}, opt *GetProjectEvent
...
+	Cadence         string    `json:"cadence"`
+	KeepN           uint      `json:"keep_n"`
```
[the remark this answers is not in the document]
cakoolen: It's weird that this number is defined as an int but at least it makes it consistent with the documentation from Gitlab itself.

projects.go
cakoolen: I've removed the field. Though when sending it using the Attributes struct, the fields needs to be present for the request to be accepted by Gitlab.

projects.go
cakoolen: Changed as requested

projects.go
cakoolen: Changed as requested

cakoolen:
@svanharmelen thanks for the quick response and the excellent review comments. I've tried to address them all. Please let me know if I missed something or there is still something I can do to improve the PR.

projects.go
```diff
@@ -464,59 +465,85 @@ func (s *ProjectsService) GetProjectEvents(pid interface{}, opt *GetProjectEvent
...
+	NameRegexDelete string     `json:"name_regex_delete"`
+	NameRegexKeep   string     `json:"name_regex_keep"`
+	Enabled         bool       `json:"enabled"`
+	NextRunAt       *time.Time `json:"next_run_at"`
+}
+
+// ContainerExpirationPolicyAttributes represents the container expiration attributes for a project.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/projects.html
+type ContainerExpirationPolicyAttributes struct {
```
svanharmelen, line 484: I think you missed this one:
```suggestion
type ContainerExpirationPolicyOptions struct {
```

projects.go
```diff
@@ -464,59 +465,85 @@ func (s *ProjectsService) GetProjectEvents(pid interface{}, opt *GetProjectEvent
...
+	Cadence         *string `url:"cadence,omitempty" json:"cadence,omitempty"`
+	KeepN           *int    `url:"keep_n,omitempty" json:"keep_n,omitempty"`
+	OlderThan       *string `url:"older_than,omitempty" json:"older_than,omitempty"`
+	NameRegex       *string `url:"name_regex,omitempty" json:"name_regex,omitempty"` // NameRegex will be replaced by NameRegexDelete but now calls without setting this field are not accepted.
```
svanharmelen, line 488: What value needs to be set here? The same as used with `NameRegexDelete`? AS if that is the case it will be very unclear for users of this API to know that they have to specify the same value twice.

Maybe we should set the value ourselves to allow users to ignore this field. What do you think? Could be something like adding this to the function bodies:

```
if opt.ContainerExpirationPolicyOptions != nil {
  opt.ContainerExpirationPolicyOptions.NameRegex = opt.ContainerExpirationPolicyOptions.NameRegexDelete
}
```

cakoolen: I would test if `opt.ContainerExpirationPolicyOptions.NameRegex == nil` before assigning `opt.ContainerExpirationPolicyOptions.NameRegexDelete` to it but it sounds like a good idea.

cakoolen: I added a method to do this since it's used in three different places. Please let me know what you think of the solution.

svanharmelen (approved):
I updated it just a little bit, but looks fine now! Thanks @cakoolen!

cakoolen:
Thanks @svanharmelen to help me out finishing this PR!

Issue #1029, opened 2021-01-12
add support for group/project invitations api


calebwashburn:
Looking for ability to leverage invitations api https://docs.gitlab.com/ee/api/invitations.html#invite-by-email-to-group-or-project

svanharmelen:
Does seem to be supported by this package yet, but open to merge a PR for it if your up for that...

framled:
I'm working for a PR for this.

Bouwdie:
@svanharmelen this has a closed mr - I guess we can close this?

svanharmelen:
Yes, seems so. Thanks for the ping 👍🏻

PR #1035, opened 2021-01-15
Support for groups/projects invites

framled:
Support for groups/projects invites #1029


svanharmelen (approved):
I updated and tweaked the solution a bit. Should be good now 👍🏻 

Thanks @framled!

PR #1040, opened 2021-01-21
update GetMergeRequestChanges() Closes #1038

### LINKED ISSUE #1038: Get single MR changes: Add support for `access_raw_diffs` and `overflow`
> Introduced in GitLab 13.6, diffs associated with the set of changes have the same size limitations applied as other diffs returned by the API or viewed via the UI. When these limits impact the results, the `overflow` field contains a value of true. Diff data without these limits applied can be retrieved by adding the `access_raw_diffs` parameter, but it is slower and more resource-intensive.

Please add `overflow` as a json return parameter and `access_raw_diffs` as a parameter attribute.

These changes were added here: https://docs.gitlab.com/ee/api/merge_requests.html#get-single-mr-changes

This will really help bot developers who need the full set of changes. Right now, there is no indication that the whole set of changes is not included via the API.

Thanks!

rhd:
Allow sending the `access_raw_diffs` parameter to the GitLab API
Parse the `overflow` field from the returned JSON

This is currently untested. I'm going to try and test later today/tomorrow. I just wanted to throw this out there.

Closes #1038

### DIFF
--- a/merge_requests.go
+++ b/merge_requests.go
@@ -104,6 +104,7 @@ type MergeRequest struct {
 	} `json:"task_completion_status"`
 	HasConflicts                bool `json:"has_conflicts"`
 	BlockingDiscussionsResolved bool `json:"blocking_discussions_resolved"`
+	Overflow                    bool `json:"overflow"`
 }
 
 func (m MergeRequest) String() string {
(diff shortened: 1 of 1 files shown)


svanharmelen (approved):
Thanks @rhd! LGTM 👍🏻

Issue #1041, opened 2021-01-21
Add support for group and project audit event APIs


cmmarslender:
API Docs:
https://docs.gitlab.com/ee/api/audit_events.html#group-audit-events

I'll work on a PR for this

PR #1042, opened 2021-01-21
Add support for group and project audit events API

cmmarslender:
Closes https://github.com/xanzy/go-gitlab/issues/1041

Should be good to go - I've tested this with an internal metrics/logging platform


audit_events.go
```diff
@@ -0,0 +1,198 @@
...
+	With          string      `json:"with"`
+	Add           string      `json:"add"`
+	As            string      `json:"as"`
+	Change        string      `json:"change"`
+	From          string      `json:"from"`
+	To            string      `json:"to"`
+	Remove        string      `json:"remove"`
+	CustomMessage string      `json:"custom_message"`
+	AuthorName    string      `json:"author_name"`
+	TargetID      TargetID    `json:"target_id"` // Sometimes this is an int and sometimes is a string with project path
```
svanharmelen, line 38: I prefer to use this field as an `interface{}` is this case. That seems to be more inline with the other locations were an ID that can be both a string and an integer.

audit_events.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	return aes, resp, err
+}
+
+// GetProjectAuditEvent gets a specific project audit event
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/audit_events.html
+func (s *AuditEventsService) GetProjectAuditEvent(pid interface{}, aid interface{}, options ...RequestOptionFunc) (*AuditEvent, *Response, error) {
```
svanharmelen, line 116:
```suggestion
func (s *AuditEventsService) GetProjectAuditEvent(pid interface{}, event int, options ...RequestOptionFunc) (*AuditEvent, *Response, error) {
```

audit_events.go
```diff
@@ -0,0 +1,198 @@
...
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	auditEventID, err := parseID(aid)
+	if err != nil {
+		return nil, nil, err
+	}
+
```
svanharmelen, line 126:
```suggestion
```

audit_events.go
```diff
@@ -0,0 +1,198 @@
...
+	u := fmt.Sprintf("projects/%s/audit_events/%s", pathEscape(project), pathEscape(auditEventID))
```
svanharmelen, line 127:
```suggestion
	u := fmt.Sprintf("projects/%s/audit_events/%d", pathEscape(project), event)
```

audit_events.go
```diff
@@ -0,0 +1,198 @@
...
+
+	req, err := s.client.NewRequest("GET", u, nil, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	a := new(AuditEvent)
```
svanharmelen, line 134:
```suggestion
	ae := new(AuditEvent)
```

audit_events.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	return aes, resp, err
+}
+
+// GetGroupAuditEvent gets a specific group audit event
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/audit_events.html
+func (s *AuditEventsService) GetGroupAuditEvent(gid interface{}, aid interface{}, options ...RequestOptionFunc) (*AuditEvent, *Response, error) {
```
svanharmelen, line 173: Same changes that I suggested to `GetProjectAuditEvent` apply here...

svanharmelen (changes requested):
Thanks @cmmarslender! Have a few suggestions/tweaks, but looks good. Also it seems the instance audit events are omitted. Maybe we should add those as well to make the new service complete? But I can do that as well, so feel free to only fix the current PR!

cmmarslender:
@svanharmelen Applied the suggestions. I didn't do the instance audit events since I was testing this against gitlab.com and don't have admin there.. I might be able to tackle that against my self hosted instance in a later PR if you don't beat me to it :)

svanharmelen (approved):
Tweaked stuff a little, but looks now! Thanks @cmmarslender!

Issue #1043, opened 2021-01-22
Support for packages 


kordianbruck:
Would love to see support for Packages (similar to containers/registry): https://docs.gitlab.com/ee/api/packages.html

svanharmelen:
Fixed by #1044

PR #1044, opened 2021-01-22
Add support for packages APIs.

kordianbruck:
This should be the basic supported for package apis requested in #1043 

* Listing packages in a project
* Listing package files 
* Deleting packages

Missing, but probably will do a separate PR for the following:

* Listing packages in a group
* Get a single project package (by ID only) - not sure this is that useful

### DIFF
--- a/packages.go
+++ b/packages.go
@@ -134,7 +134,8 @@ func (s *PackagesService) ListPackageFiles(pid interface{}, packageID int, opt *
 	if err != nil {
 		return nil, nil, err
 	}
-	u := fmt.Sprintf("projects/%s/packages/%d/package_files",
+	u := fmt.Sprintf(
+		"projects/%s/packages/%d/package_files",
 		pathEscape(project),
 		packageID,
 	)
--- a/packages.go
+++ b/packages.go
@@ -36,7 +36,7 @@ type Package struct {
 	ID        int        `json:"id"`
 	Name      string     `json:"name"`
 	Version   string     `json:"version"`
-	Type      string     `json:"package_type"`
+	PackageType      string     `json:"package_type"`
 	CreatedAt *time.Time `json:"created_at"`
 }
 
--- a/packages.go
+++ b/packages.go
@@ -79,11 +79,11 @@ func (s PackageFile) String() string {
 // https://docs.gitlab.com/ee/api/packages.html#within-a-project
 type ListProjectPackagesOptions struct {
 	ListOptions
-	OrderBy            string `url:"order_by,omitempty" json:"order_by,omitempty"`
-	Sort               string `url:"sort,omitempty" json:"sort,omitempty"`
-	PackageType        string `url:"package_type,omitempty" json:"package_type,omitempty"`
-	PackageName        string `url:"package_name,omitempty" json:"package_name,omitempty"`
-	IncludeVersionless bool   `url:"include_versionless,omitempty" json:"include_versionless,omitempty"`
+	OrderBy            *string `url:"order_by,omitempty" json:"order_by,omitempty"`
+	Sort               *string `url:"sort,omitempty" json:"sort,omitempty"`
+	PackageType        *string `url:"package_type,omitempty" json:"package_type,omitempty"`
+	PackageName        *string `url:"package_name,omitempty" json:"package_name,omitempty"`
+	IncludeVersionless *bool   `url:"include_versionless,omitempty" json:"include_versionless,omitempty"`
 }
 
 // ListProjectPackages gets a list of packages in a project.
--- a/packages.go
+++ b/packages.go
@@ -48,7 +48,7 @@ func (s Package) String() string {
 // PackageLinks holds links for itself and deleting.
 type PackageLinks struct {
 	WebPath       string `json:"web_path"`
-	DeleteApiPath string `json:"delete_api_path"`
+	DeleteAPIPath string `json:"delete_api_path"`
 }
 
 func (s PackageLinks) String() string {
--- a/packages.go
+++ b/packages.go
@@ -103,12 +103,12 @@ func (s *PackagesService) ListProjectPackages(pid interface{}, opt *ListProjectP
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/packages.html#delete-a-project-package
-func (s *PackagesService) DeleteProjectPackage(pid interface{}, packageId int, options ...RequestOptionFunc) (*Response, error) {
+func (s *PackagesService) DeleteProjectPackage(pid interface{}, packageID int, options ...RequestOptionFunc) (*Response, error) {
 	project, err := parseID(pid)
 	if err != nil {
 		return nil, err
 	}
-	u := fmt.Sprintf("projects/%s/packages/%d", pathEscape(project), packageId)
+	u := fmt.Sprintf("projects/%s/packages/%d", pathEscape(project), packageID)
 
 	req, err := s.client.NewRequest("DELETE", u, nil, options)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


packages.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// Package represents a GitLab single package
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/packages.html
+type Package struct {
+	ID        int        `json:"id"`
+	Name      string     `json:"name"`
+	Version   string     `json:"version"`
+	Type      string     `json:"package_type"`
```
svanharmelen, line 39: I see the repetition, but lets use the actual names.

```suggestion
	PackageType      string     `json:"package_type"`
```

packages.go
```diff
@@ -0,0 +1,154 @@
...
+	CreatedAt *time.Time `json:"created_at"`
+}
```
svanharmelen, line 41: Are `_links` and `tags` missing?

packages.go
```diff
@@ -0,0 +1,154 @@
...
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/packages.html#within-a-project
+type ListProjectPackagesOptions struct {
+	ListOptions
+	OrderBy            string `url:"order_by,omitempty" json:"order_by,omitempty"`
+	Sort               string `url:"sort,omitempty" json:"sort,omitempty"`
+	Type               string `url:"package_type,omitempty" json:"package_type,omitempty"`
+	Name               string `url:"package_name,omitempty" json:"package_name,omitempty"`
+	IncludeVersionless bool   `url:"include_versionless,omitempty" json:"include_versionless,omitempty"`
+}
```
svanharmelen, line 75: These should all be pointers...

packages.go
```diff
@@ -0,0 +1,154 @@
...
+
+// ListProjectPackagesOptions are the parameters available in a ListProjectPackages() Operation
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/packages.html#within-a-project
+type ListProjectPackagesOptions struct {
+	ListOptions
+	OrderBy            string `url:"order_by,omitempty" json:"order_by,omitempty"`
+	Sort               string `url:"sort,omitempty" json:"sort,omitempty"`
+	Type               string `url:"package_type,omitempty" json:"package_type,omitempty"`
```
svanharmelen, line 72:
```suggestion
	PackageType               string `url:"package_type,omitempty" json:"package_type,omitempty"`
```

packages.go
```diff
@@ -0,0 +1,154 @@
...
+	Name               string `url:"package_name,omitempty" json:"package_name,omitempty"`
```
svanharmelen, line 73:
```suggestion
	PackageName               string `url:"package_name,omitempty" json:"package_name,omitempty"`
```

packages.go
```diff
@@ -0,0 +1,154 @@
...
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/packages", pathEscape(project))
+
+	req, err := s.client.NewRequest("GET", u, opt, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var packages []*Package
```
svanharmelen, line 93:
```suggestion
	var ps []*Package
```

packages.go
```diff
@@ -0,0 +1,154 @@
...
+	resp, err := s.client.Do(req, &packages)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return packages, resp, err
+}
+
+// DeleteRegistryRepository deletes a repository in a registry.
```
svanharmelen, line 102: Copy/paste error?

packages.go
```diff
@@ -0,0 +1,154 @@
...
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/packages.html#delete-a-project-package
+func (s *PackagesService) DeleteProjectPackage(pid interface{}, packageID int, options ...RequestOptionFunc) (*Response, error) {
```
svanharmelen, line 106:
```suggestion
func (s *PackagesService) DeleteProjectPackage(pid interface{}, package int, options ...RequestOptionFunc) (*Response, error) {
```

packages.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	var tags []*PackageFile
+	resp, err := s.client.Do(req, &tags)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return tags, resp, err
+}
```
svanharmelen, line 154: Please follow the order of the docs. So this method should be above the `DeleteProjectPackage` method.

packages.go
```diff
@@ -0,0 +1,154 @@
...
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/packages.html
+type PackageFile struct {
+	ID        int        `json:"id"`
+	PackageID int        `json:"package_id"`
+	CreatedAt *time.Time `json:"created_at"`
+	FileName  string     `json:"file_name"`
+	Size      int        `json:"size"`
+	MD5       string     `json:"file_md5"`
+	SHA1      string     `json:"file_sha1"`
```
svanharmelen, line 57:
```suggestion
	FileMD5       string     `json:"file_md5"`
	FileSHA1      string     `json:"file_sha1"`
```

packages.go
```diff
@@ -0,0 +1,154 @@
...
+}
```
svanharmelen, line 58: It seems 'pipelines' is missing?

packages.go
```diff
@@ -0,0 +1,154 @@
...
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/packages.html#list-package-files
+type ListPackageFilesOptions ListOptions
+
+// ListPackageFiles gets a list of files that are within a package
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/packages.html#list-package-files
+func (s *PackagesService) ListPackageFiles(pid interface{}, packageID int, opt *ListPackageFilesOptions, options ...RequestOptionFunc) ([]*PackageFile, *Response, error) {
```
svanharmelen, line 132:
```suggestion
func (s *PackagesService) ListPackageFiles(pid interface{}, package int, opt *ListPackageFilesOptions, options ...RequestOptionFunc) ([]*PackageFile, *Response, error) {
```

packages.go
```diff
@@ -0,0 +1,154 @@
...
+	project, err := parseID(pid)
+	if err != nil {
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/packages/%d/package_files",
+		pathEscape(project),
+		packageID,
+	)
```
svanharmelen, line 140: For consistency:

```suggestion
	u := fmt.Sprintf(
		"projects/%s/packages/%d/package_files",
		pathEscape(project),
		packageID,
	)
```

packages.go
```diff
@@ -0,0 +1,154 @@
...
+
+	req, err := s.client.NewRequest("GET", u, opt, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var tags []*PackageFile
```
svanharmelen, line 147: I guess `tags` should be `pfs`?

```suggestion
	var pfs []*PackageFile
```

svanharmelen (changes requested):
Thanks @kordianbruck! Al lot of small suggestions, but should be easy to update/fix...

packages.go
kordianbruck: `package` is a resevered word in go, so that won't work. pid is also already taken.

svanharmelen: Ah... Good point. Maybe we should call it `pkg` then? While I understand it's an ID, in the rest of this package these variables are names after the entity without adding `ID` at the end. That's why I would like this to do the same...

kordianbruck: SGTM

packages.go
kordianbruck: indeed

packages.go
kordianbruck: Indeed, unfortunately I have no idea how the "tags" would look like (strings or more nested structure?). I've added the links, as that seems to be similar as in other calls.

packages.go
kordianbruck: Added

svanharmelen (approved):
LGTM! Thanks @kordianbruck!

loganknecht:
Hey all!

This is fantastic work! Thank you for providing the ability to get packages.

Do you know if there is anyone currently in charge of the effort for the missing pieces?

- Listing packages in a group
- Get a single project package (by ID only)

Right now my organization is trying to find a solution that can retrieve packages and I was anticipating support for that here. But it looks like we have more work to do 😄

PR #1045, opened 2021-01-23
Add allow_collaboration to MergeRequest struct

profclems:
The MergeRequest struct was missing the allow_collaboration parameter which is part of the "GET Single Merge Request" response payload

https://docs.gitlab.com/ce/api/merge_requests.html#get-single-mr

### DIFF
--- a/merge_requests.go
+++ b/merge_requests.go
@@ -105,6 +105,7 @@ type MergeRequest struct {
 	HasConflicts                bool `json:"has_conflicts"`
 	BlockingDiscussionsResolved bool `json:"blocking_discussions_resolved"`
 	Overflow                    bool `json:"overflow"`
+	AllowCollaboration          bool `json:"allow_collaboration"`
 }
 
 func (m MergeRequest) String() string {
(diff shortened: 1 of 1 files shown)


merge_requests.go
```diff
@@ -105,6 +105,7 @@ type MergeRequest struct {
 	HasConflicts                bool `json:"has_conflicts"`
 	BlockingDiscussionsResolved bool `json:"blocking_discussions_resolved"`
 	Overflow                    bool `json:"overflow"`
+	AllowCollaboration          bool `json:"allow_collaboration"`
```
svanharmelen, line 108: Maybe add `allow_maintainer_to_push` as well? And please order them following the docs (so these should be above the `WebURL` field.

svanharmelen (changes requested):
Thanks @profclems! Please have a look at the inline suggestions...

merge_requests.go
profclems: > Maybe add allow_maintainer_to_push as well?

`allow_maintainer_to_push` is deprecated in favour of `allow_collaboration` field so I don't think it should be added

profclems: > And please order them following the docs (so these should be above the WebURL field.

Fixed that

profclems:
@svanharmelen `allow_maintainer_to_push` is deprecated in favour of `allow_collaboration` field so I don't think it should be added.

merge_requests.go
svanharmelen: Ah, good point. Didn't know that 👍🏻

svanharmelen (approved):

svanharmelen:
Thanks @profclems 👍🏻

PR #1046, opened 2021-01-24
docs: add webhook example

Yengas:
## Description
The `ParseWebhook` function has a pseudo-code example above it. I wanted to add a runnable example of how to use `go-gitlab` with webhooks.

### DIFF
--- a/examples/webhook.go
+++ b/examples/webhook.go
@@ -13,18 +13,16 @@ import (
 	"github.com/xanzy/go-gitlab"
 )
 
-// webhook is a HTTP Handler for Gitlab Webhook events
+// webhook is a HTTP Handler for Gitlab Webhook events.
 type webhook struct {
 	Secret         string
 	EventsToAccept []gitlab.EventType
 }
 
-// webhookExample shows how to create a Webhook server to parse Gitlab events
-// Listens /webhook endpoint on host 0.0.0.0 port 8080
+// webhookExample shows how to create a Webhook server to parse Gitlab events.
 func webhookExample() {
 	wh := webhook{
-		Secret: "your-gitlab-secret",
-		// Define which types of events we want to subscribe to in our case, Merge Request and Pipeline
+		Secret:         "your-gitlab-secret",
 		EventsToAccept: []gitlab.EventType{gitlab.EventTypeMergeRequest, gitlab.EventTypePipeline},
 	}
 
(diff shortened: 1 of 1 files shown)


examples/webhook.go
```diff
@@ -0,0 +1,109 @@
...
+func webhookExample() {
+	wh := webhook{
+		Secret: "your-gitlab-secret",
+		// only listen for MergeRequest and Pipeline events
+		EventsToAccept: []gitlab.EventType{gitlab.EventTypeMergeRequest, gitlab.EventTypePipeline},
+	}
+
+	mux := http.NewServeMux()
+	mux.Handle("/webhook", wh)
+	_ = http.ListenAndServe("0.0.0.0:8080", mux)
```
svanharmelen, line 33: Please handle the errors (goes for the whole file).

examples/webhook.go
```diff
@@ -0,0 +1,109 @@
...
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"github.com/xanzy/go-gitlab"
+	"io"
+	"io/ioutil"
+	"net/http"
+)
```
svanharmelen, line 11: Please place external deps in their own block...

```suggestion
import (
	"encoding/json"
	"errors"
	"fmt"
	"io"
	"io/ioutil"
	"net/http"
	
	"github.com/xanzy/go-gitlab"
)
```

examples/webhook.go
```diff
@@ -0,0 +1,109 @@
...
+
+// parse errors
```
svanharmelen, line 13: In Go, comments are suppose to be sentences. So they should start with a capital letter ann end with a dot.

examples/webhook.go
```diff
@@ -0,0 +1,109 @@
...
+	return gitlab.ParseWebhook(gitLabEvent, payload)
+}
+
+func isEventSubscribed(event gitlab.EventType, events []gitlab.EventType) bool {
+	for _, e := range events {
+		if event == e {
+			return true
+		}
+	}
+
```
svanharmelen, line 107:
```suggestion
```

examples/webhook.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+// parse errors
+var (
+	errInvalidHTTPMethod             = errors.New("invalid HTTP Method")
+	errMissingGitLabEventHeader      = errors.New("missing X-Gitlab-Event Header")
+	errGitLabTokenVerificationFailed = errors.New("X-Gitlab-Token validation failed")
+	errEventNotFound                 = errors.New("event not defined to be parsed")
+	errParsingPayload                = errors.New("error parsing payload")
+)
```
svanharmelen, line 20: While this is perfectly fine, I prefer to not declare these up front in this example. In my opinion it takes the focus away from the actual example and they are used only once anyway...

examples/webhook.go
```diff
@@ -0,0 +1,109 @@
...
+	mux := http.NewServeMux()
+	mux.Handle("/webhook", wh)
+	_ = http.ListenAndServe("0.0.0.0:8080", mux)
+}
+
+// webhook is a HTTP Handler for Gitlab Webhook events
+type webhook struct {
+	Secret         string
+	EventsToAccept []gitlab.EventType
+}
```
svanharmelen, line 40: Please place structs in the beginning of the file (for consistency). So just below the import section in this case...

svanharmelen (changes requested):
Thanks for adding this @Yengas! But to make it fit in with the rest of the code, it needs some tweaks first...

examples/webhook.go
Yengas: hey, I am adding a fatal log in case the HTTP server start fails. 

how would you like me to handle all the other errors like writing a response to the HTTP request, closing the request body, etc?

svanharmelen: I would say that in an example like this, using `log.Fatalf` or just `log.Fatal` should be fine 👍🏻

examples/webhook.go
svanharmelen: Please end your comments with a dot...

svanharmelen:
Almost there @Yengas!

Yengas:
Hey @svanharmelen, thanks for all the feedback. I've added the missing logging to error cases. LMK if you want anything else changed.

svanharmelen (approved):
Tweaked it just a little more, looks good now! Thanks!

PR #1048, opened 2021-01-25
Add operations_access_level flag to projects

cmmarslender:
@svanharmelen I've got another PR for you :)

This flag was added in the most recent release of GitLab


svanharmelen:
Looks good, but I think we should then also add the field to the create and edit structs. Right?

cmmarslender:
> Looks good, but I think we should then also add the field to the create and edit structs. Right?

Yes, good call. Added that as well

svanharmelen (approved):
Thanks @cmmarslender! LGTM!

PR #1051, opened 2021-01-29
GetGroupIssuesStatisticsOptions.Milestone should be a string

ax1036:
https://docs.gitlab.com/ee/api/issues_statistics.html#get-group-issues-statistics

image


svanharmelen (approved):
Good find @ax1036 👍🏻

PR #1055, opened 2021-02-05
Added list billable member func

Starttoaster:
Under the same API as the regular group member functions, GitLab has a list "billable" group members function that was released fairly recently. See: https://docs.gitlab.com/ee/api/members.html#list-all-billable-members-of-a-group

The structure of a billable group member is fairly close to that of a regular group member, but the main difference is the removal of the `expires_at`, `access_level`, and `group_saml_identity` fields, and the addition of a `last_activity_on` field.

### DIFF
--- a/group_members.go
+++ b/group_members.go
@@ -18,6 +18,7 @@ package gitlab
 
 import (
 	"fmt"
+	"net/http"
 )
 
 // GroupMembersService handles communication with the group members
@@ -161,6 +162,7 @@ type BillableGroupMember struct {
 	State          string  `json:"state"`
 	AvatarURL      string  `json:"avatar_url"`
 	WebURL         string  `json:"web_url"`
+	Email          string  `json:"email"`
 	LastActivityOn ISOTime `json:"last_activity_on"`
 }
 
@@ -186,7 +188,7 @@ func (s *GroupsService) ListBillableGroupMembers(gid interface{}, opt *ListBilla
 	}
 	u := fmt.Sprintf("groups/%s/billable_members", pathEscape(group))
 
-	req, err := s.client.NewRequest("GET", u, opt, options)
+	req, err := s.client.NewRequest(http.MethodGet, u, opt, options)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/group_members.go
+++ b/group_members.go
@@ -179,7 +179,7 @@ type ListBillableGroupMembersOptions struct {
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/members.html#list-all-billable-members-of-a-group
-func (s *GroupsService) ListBillableGroupMembers(gid interface{}, opt *ListGroupMembersOptions, options ...RequestOptionFunc) ([]*BillableGroupMember, *Response, error) {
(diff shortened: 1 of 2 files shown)


Starttoaster:
"Done" playing with this now 😄 Had some conversations internal to my group regarding this API function addition that caused some post-PR opening changes. 

I know the convention you tend to prefer (and it makes sense to me) is to have a new service for each html file in GitLab's API docs, and this is underneath `members.html`. I looked for a `members.go`, but it looks like everyone else has been putting these functions under `group_members.go` or `project_members.go` so far. If you require it to be a new service, let me know and I can do that.

group_members.go
```diff
@@ -151,6 +151,55 @@ func (s *GroupMembersService) GetGroupMember(gid interface{}, user int, options
...
+// GitLab API docs: https://docs.gitlab.com/ee/api/members.html#list-all-billable-members-of-a-group
+type BillableGroupMember struct {
+	ID             int     `json:"id"`
+	Username       string  `json:"username"`
+	Name           string  `json:"name"`
+	State          string  `json:"state"`
+	AvatarURL      string  `json:"avatar_url"`
+	WebURL         string  `json:"web_url"`
+	LastActivityOn ISOTime `json:"last_activity_on"`
+}
```
svanharmelen, line 165: It seems you missed the `email` field (in between the `WebURL` and the `LastActivityOn` fields).

Additionally I wonder if we shouldn't just add the missing fields to the `GroupMember` struct and use that one instead. What do you think?

group_members.go
```diff
@@ -151,6 +151,55 @@ func (s *GroupMembersService) GetGroupMember(gid interface{}, user int, options
...
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/members.html#list-all-billable-members-of-a-group
+func (s *GroupsService) ListBillableGroupMembers(gid interface{}, opt *ListBillableGroupMembersOptions, options ...RequestOptionFunc) ([]*BillableGroupMember, *Response, error) {
+	group, err := parseID(gid)
+	if err != nil {
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("groups/%s/billable_members", pathEscape(group))
+
+	req, err := s.client.NewRequest("GET", u, opt, options)
```
svanharmelen, line 189: I just merged a PR to use the constants everywhere, so let use that here as well:
```suggestion
	req, err := s.client.NewRequest(http.MethodGet, u, opt, options)
```

svanharmelen (changes requested):
Looks good, but have 2 comments...

group_members.go
svanharmelen: See you already addressed this in your comments 😊 I'm good with using this new struct if you add the missing `email` field...

Starttoaster: > It seems you missed the email field (in between the WebURL and the LastActivityOn fields).

🤦 The email field wasn't in the first example billable group member object in the list so I missed it. Good catch!

Starttoaster: Fixed. Thanks @svanharmelen !

group_members.go
Starttoaster: Fixed locally :)

svanharmelen (approved):
Thanks @Starttoaster! LGTM!

PR #1062, opened 2021-02-09
Add ResourceStateEvent service

5nord:
This PR introduces the resource state event API provided since GitLab 13.2

Resource state events keep track of what happens to GitLab issues and merge requests.
You use them to track which state was set, who did it, and when it happened.


svanharmelen (approved):
Made a few small tweaks, but looks good to me! Thanks @5nord 👍🏻

PR #1067, opened 2021-02-11
renamed FileAction to FileActionValue + added helper function

### LINKED ISSUE #1063: Provide a factory function for FileAction
Hi !

While using commits api, type `FileAction` is required to provide `Action` to `CommitActionOptions`.
Action ask for `*FileAction` while vars FileCreate/FileDelete/FileMove/FileUpdate returns `FileAction`.
In practice, it means we have to declare a new variable, then use this variable pointer as parameter like below :

```go
fa := gitlab.FileCreate
ca := gitlab.CommitActionOptions{Action: &fa}
```

I propose to add a factory function like `String(v string) *string` `Bool(v string) *string` to get better developper experience.
I was thinking of `NewFileAction(v FileAction) *FileAction` but it would break the naming convention.
`FileAction(v FileAction) *FileAction` is not possible cause of conflict with type name.

Any idea how to handle this properly ?

ahuret:
renamed `FileAction` to `FileActionValue` + added helper function
```go
func FileAction(v FileActionValue) *FileActionValue
```
that helps instantiate `*FileActionValue` type required in Commits API (commits.go).
breaks BC
close #1063

### DIFF
--- a/commits.go
+++ b/commits.go
@@ -197,13 +197,13 @@ type CreateCommitOptions struct {
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/commits.html#create-a-commit-with-multiple-files-and-actions
 type CommitActionOptions struct {
-	Action          *FileAction `url:"action,omitempty" json:"action,omitempty"`
-	FilePath        *string     `url:"file_path,omitempty" json:"file_path,omitempty"`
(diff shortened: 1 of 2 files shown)


svanharmelen (approved):
Looks good to me and since it's only used in one location, I'm also not too worried about the possible impact. Should be a super quick and easy fix when updating the package... So I'll go ahead and merge this one 👍🏻

PR #1069, opened 2021-02-13
Add chmod action option

harsimranmaan:
Works for setting exectuable files
See https://docs.gitlab.com/ce/api/commits.html\#create-a-commit-with-multiple-files-and-actions


svanharmelen (approved):
Thanks @harsimranmaan! LGTM!

PR #1074, opened 2021-02-19
FIX(repository_files): Fix Filename Escape

Just-maple:
here should use  `pathEscape` to parse `.` to `%2E`

not so some filename(most of normally) with `.` will return file not found 

[gitlab doc](https://docs.gitlab.com/ce/api/repository_files.html#get-raw-file-from-repository)

### DIFF
--- a/repository_files.go
+++ b/repository_files.go
@@ -112,7 +112,7 @@ func (s *RepositoryFilesService) GetFileMetaData(pid interface{}, fileName strin
 	u := fmt.Sprintf(
 		"projects/%s/repository/files/%s",
 		pathEscape(project),
-		pathEscape(url.PathEscape(fileName)),
+		pathEscape(fileName),
 	)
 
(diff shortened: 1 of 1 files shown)


repository_files.go
```diff
@@ -112,7 +112,7 @@ func (s *RepositoryFilesService) GetFileMetaData(pid interface{}, fileName strin
 	u := fmt.Sprintf(
 		"projects/%s/repository/files/%s",
 		pathEscape(project),
-		url.PathEscape(fileName),
+		pathEscape(url.PathEscape(fileName)),
```
svanharmelen, line 115: No need to call `url.PathEscape` twice:

```suggestion
		pathEscape(fileName),
```

svanharmelen (changes requested):

svanharmelen (approved):
Thanks @Just-maple! LGTM 👍🏻

PR #1076, opened 2021-02-20
fix(repository_files): fix filename escape

Just-maple:
it seems that all filename param in this api scope have the same issue so they need add the `.` escape


svanharmelen (approved):
LGTM! Thanks @Just-maple 👍🏻

PR #1078, opened 2021-02-21
Add field "search" to ListLabelsOptions

etherealy:


etherealy:
Do you want a test associated with it? 
As they are mocked, it doesn't feel like it would add much value ... 

I did test it against a GitLab instance just to be sure

svanharmelen (approved):
LGTM! Thanks @etherealy 👍🏻

PR #1081, opened 2021-02-25
Fix struct field types for MergeCommentEvent.ObjectAttributes.ResolvedByID and ResolvedByPush

venyii:
Taken from a GitLab.com instance webhook some hours ago:

```
"object_attributes": {
  ...
  "resolved_by_id": 4872349,
  "resolved_by_push": false,
  ...
}
```

The GitLab docs don't mention the fields unfortunately in their example.

I can add a test but I wasn't sure if I should just duplicate `note_merge_request` - let me know how to handle the different variations here if needed :)


svanharmelen (approved):
Thanks @venyii! LGTM 👍🏻

PR #1082, opened 2021-02-25
Add support for the Release webhook

vntw:
See docs at https://docs.gitlab.com/ce/user/project/integrations/webhooks.html#release-events

The `release.json` is an actual webhook payload from a few hours ago, the only thing I manually added was the `Assets.Links` field from the docs so I could test it

I also added the `releases_events` to various hook options. The only part I left as it is, is the [Services API](https://docs.gitlab.com/ee/api/services.html) since I couldn't find out whether it supported the flag, so I left it for now. All other changed structs are documented


svanharmelen (approved):
Thanks @vntw! I just reordered some things, but LGTM!

PR #1095, opened 2021-03-08
Fix incorrect Slack service property webhook

### LINKED ISSUE #1094: Release 0.45.0 broke slack service
The latest release 0.45.0 mistakenly renamed the `webhook` property to `web_hook`, in the Slack service.

https://github.com/xanzy/go-gitlab/blob/7988b8a7ad8809789196b385a9629a3050bbf066/services.go#L900

armsnyder:
Fixes #1094 

API ref: https://docs.gitlab.com/ee/api/services.html#slack-notifications

### DIFF
--- a/services.go
+++ b/services.go
@@ -897,7 +897,7 @@ type SlackService struct {
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/services.html#slack
 type SlackServiceProperties struct {
-	WebHook                   string    `json:"web_hook"`
+	WebHook                   string    `json:"webhook"`
 	Username                  string    `json:"username"`
 	Channel                   string    `json:"channel"`
 	NotifyOnlyBrokenPipelines BoolValue `json:"notify_only_broken_pipelines"`


services.go
```diff
@@ -897,7 +897,7 @@ type SlackService struct {
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/services.html#slack
 type SlackServiceProperties struct {
-	WebHook                   string    `json:"web_hook"`
+	WebHook                   string    `json:"webhook"`
```
svanharmelen, line 900:
```suggestion
	Webhook                   string    `json:"webhook"`
```

svanharmelen (commented):
Sorry for that! If you can update the field name, I'll merge it right away 👍🏻

services.go
armsnyder: This is a breaking change

svanharmelen: Interesting... I now see they are all "wrong" in this file. I guess that is also why this slipped in. I used the add tags logic from `gopls` which converts CamelCase to snake_case to create the tag names. But in this case (while it's clearly wrong) I'm fine with leaving it like this...

svanharmelen (approved):
While in a way still wrong, I'm good with this one! Thanks @armsnyder!

svanharmelen:
Released `v0.46.0`

services.go
armsnyder: Yeah I'd obviously prefer not renaming the struct field so we don't need to deal with the breaking change. 😄 

Though if you have code generation working I can see why you'd rename it.

armsnyder:
Thanks!

PR #1099, opened 2021-03-09
Add suggestion_commit_message to project options

madpipeline:
Documented in https://docs.gitlab.com/ce/api/projects.html#edit-project and https://docs.gitlab.com/ce/api/projects.html#create-project-for-user

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -535,6 +535,7 @@ type CreateProjectOptions struct {
 	PackagesEnabled                           *bool                                `url:"packages_enabled,omitempty" json:"packages_enabled,omitempty"`
 	ServiceDeskEnabled                        *bool                                `url:"service_desk_enabled,omitempty" json:"service_desk_enabled,omitempty"`
 	AutocloseReferencedIssues                 *bool                                `url:"autoclose_referenced_issues,omitempty" json:"autoclose_referenced_issues,omitempty"`
+	SuggestionCommitMessage			  *string			       `url:"suggestion_commit_message,omitempty" json:"suggestion_commit_message,omitempty"`
 
 	// Deprecated members
 	IssuesEnabled        *bool `url:"issues_enabled,omitempty" json:"issues_enabled,omitempty"`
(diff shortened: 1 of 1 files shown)


projects.go
```diff
@@ -672,6 +673,7 @@ type EditProjectOptions struct {
 	PackagesEnabled                           *bool                                `url:"packages_enabled,omitempty" json:"packages_enabled,omitempty"`
 	ServiceDeskEnabled                        *bool                                `url:"service_desk_enabled,omitempty" json:"service_desk_enabled,omitempty"`
 	AutocloseReferencedIssues                 *bool                                `url:"autoclose_referenced_issues,omitempty" json:"autoclose_referenced_issues,omitempty"`
+	SuggestionCommitMessage			  *string			       `url:"suggestion_commit_message,omitempty" json:"suggestion_commit_message,omitempty"`
```
svanharmelen, line 676: I guess we should also add the field to the `Project` struct right? Just below the `AutocloseReferencedIssues` field to keep the ordering inline with the docs...

madpipeline: Added. Thank you.

projects.go
```diff
@@ -106,6 +106,7 @@ type Project struct {
 	PagesAccessLevel                          AccessControlValue         `json:"pages_access_level"`
 	OperationsAccessLevel                     AccessControlValue         `json:"operations_access_level"`
 	AutocloseReferencedIssues                 bool                       `json:"autoclose_referenced_issues"`
+	SuggestionCommitMessage                   string                     `url:"suggestion_commit_message,omitempty" json:"suggestion_commit_message,omitempty"`
```
svanharmelen, line 109: The url tag and `omitempty` is not needed in this struct:

```suggestion
	SuggestionCommitMessage                   string                     `json:"suggestion_commit_message"`
```

svanharmelen (approved):
Thanks @madpipeline 👍🏻

DavidMosoarca:
I am waiting to use the functionality proposed into these Pull Requests and I will wait for a new release which will contain this improvements. Should I understand that it will appear any time soon?

madpipeline:
@svanharmelen what is the release schedule for this project? Is it something established or whenever there are enough or big enough changes?

PR #1103, opened 2021-03-11
Add `Warnings` field to `LintResult` struct

profclems:
The `ci/lint` endpoint does not only return `status` and `errors` but returns `warnings` as well in the response payload and shouldn't be ignored


svanharmelen (approved):
Thanks @profclems 👍🏻

PR #1106, opened 2021-03-16
Add support for the filter field in GroupLDAPLink

### LINKED ISSUE #1105: Support the filter field on GroupLDAPLink
It's documented here: https://docs.gitlab.com/ee/api/groups.html#add-ldap-group-link-with-cn-or-filter.

sirlatrom:
Closes #1105.

### DIFF
--- a/groups.go
+++ b/groups.go
@@ -81,6 +81,7 @@ type Group struct {
 // GitLab API docs: https://docs.gitlab.com/ce/api/groups.html#ldap-group-links
 type LDAPGroupLink struct {
 	CN          string           `json:"cn"`
+	Filter      string           `json:"filter"`
 	GroupAccess AccessLevelValue `json:"group_access"`
 	Provider    string           `json:"provider"`
 }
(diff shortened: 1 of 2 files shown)


groups.go
```diff
@@ -458,6 +459,18 @@ func (s *GroupsService) ListGroupLDAPLinks(gid interface{}, options ...RequestOp
 // https://docs.gitlab.com/ee/api/groups.html#add-ldap-group-link-starter
 type AddGroupLDAPLinkOptions struct {
 	CN          *string `url:"cn,omitempty" json:"cn,omitempty"`
+	Filter      *string `url:"filter,omitempty" json:"filter,omitempty"`
+	GroupAccess *int    `url:"group_access,omitempty" json:"group_access,omitempty"`
```
svanharmelen, line 463:
```suggestion
	GroupAccess *AccessLevelValue    `url:"group_access,omitempty" json:"group_access,omitempty"`
```

groups.go
```diff
@@ -508,6 +521,26 @@ func (s *GroupsService) DeleteGroupLDAPLink(gid interface{}, cn string, options
 	return s.client.Do(req, nil)
 }
 
+// DeleteGroupLDAPLinkWithCNOrFilter deletes a group LDAP link. Available only for users who
+// can edit groups.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/groups.html#delete-ldap-group-link-with-cn-or-filter
+func (s *GroupsService) DeleteGroupLDAPLinkWithCNOrFilter(gid interface{}, opts *DeleteGroupLDAPLinkWithCNOrFilterOptions, options ...RequestOptionFunc) (*Response, error) {
```
svanharmelen, line 529: I think calling this just `DeleteGroupLDAPLink` would probably make more sense, right?

groups.go
```diff
@@ -458,6 +459,18 @@ func (s *GroupsService) ListGroupLDAPLinks(gid interface{}, options ...RequestOp
...
+}
+
+// DeleteGroupLDAPLinkWithCNOrFilterOptions represents the available DeleteGroupLDAPLinkWithCNOrFilter() options.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/groups.html#delete-ldap-group-link-with-cn-or-filter
+type DeleteGroupLDAPLinkWithCNOrFilterOptions struct {
+	CN          *string `url:"cn,omitempty" json:"cn,omitempty"`
+	Filter      *string `url:"filter,omitempty" json:"filter,omitempty"`
 	GroupAccess *int    `url:"group_access,omitempty" json:"group_access,omitempty"`
```
svanharmelen, line 474: This field is not supported here.

svanharmelen (changes requested):

groups.go
sirlatrom: I tend to agree, but there already is such a function with a single string parameter for the `cn` field, so that would be a breaking change.

svanharmelen: Check... Then I'll merge this as is 👍🏻

svanharmelen (approved):
Thanks @sirlatrom 👍🏻

PR #1119, opened 2021-03-31
projects: add ProjectLicense to projects response type

MaicoTimmerman:
The `GetProjectOptions` has the license flag to include license in the response, however the response value is missing the fields.

Reference: https://docs.gitlab.com/ee/api/projects.html#get-single-project

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -75,6 +75,8 @@ type Project struct {
 	EmptyRepo                                 bool                       `json:"empty_repo"`
 	Archived                                  bool                       `json:"archived"`
 	AvatarURL                                 string                     `json:"avatar_url"`
+	LicenseURL                                string                     `json:"license_url"`
+	License                                   *ProjectLicense            `json:"license"`
 	SharedRunnersEnabled                      bool                       `json:"shared_runners_enabled"`
 	ForksCount                                int                        `json:"forks_count"`
 	StarCount                                 int                        `json:"star_count"`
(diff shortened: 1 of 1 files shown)


projects.go
```diff
@@ -75,6 +75,8 @@ type Project struct {
 	EmptyRepo                                 bool                       `json:"empty_repo"`
 	Archived                                  bool                       `json:"archived"`
 	AvatarURL                                 string                     `json:"avatar_url"`
+	License                                   *ProjectLicense            `json:"license"`
+	LicenseURL                                string                     `json:"license_url"`
```
svanharmelen, line 79: Please switch these two fields so they follow the order in the docs...

projects.go
```diff
@@ -122,6 +124,14 @@ type Project struct {
 	BuildCoverageRegex   string             `json:"build_coverage_regex"`
 }
 
+type ProjectLicense struct {
+	Key       string `json:"key"`
+	Name      string `json:"name"`
+	Nickname  string `json:"nickname"`
+	HTMLURL   string `json:"html_url"`
+	SourceURL string `json:"source_url"`
+}
```
svanharmelen, line 133: Please place this struct below the `ProjectNamespace` struct. Thanks!

svanharmelen (changes requested):
Thanks! Two minor requests through...

projects.go
MaicoTimmerman: I've switched the fields as you requested, however note that there are some existing fields that are not in order compared to the docs. See for example `MarkedForDeletionAt` and `EmptyRepo`.

svanharmelen (approved):
Thanks @MaicoTimmerman! LGTM 👍🏻

projects.go
svanharmelen: Check. Thanks for the heads up... Probably missed that in an earlier PR...

PR #1121, opened 2021-04-01
client_options: let user pass custom Logger and LeveledLogger

bmeneguele:
go-retryablehttp has the option to add a custom Logger and LeveledLogger,
allowing users to pass their own implementation of them and possibly
silencing HTTP debug messages in case of retry or timeouts.

This patch adds two HTTP client options for setting these Loggers.

Signed-off-by: Bruno Meneguele <bmeneg@redhat.com>

### DIFF
--- a/client_options.go
+++ b/client_options.go
@@ -40,12 +40,28 @@ func WithCustomBackoff(backoff retryablehttp.Backoff) ClientOptionFunc {
 	}
 }
 
+// WithCustomLogger can be used to configure a custom retryablehttp leveled logger
+func WithCustomLeveledLogger(leveledLogger retryablehttp.LeveledLogger) ClientOptionFunc {
+	return func(c *Client) error {
+		c.client.Logger = leveledLogger
+		return nil
+	}
+}
+
 // WithCustomLimiter injects a custom rate limiter to the client.
 func WithCustomLimiter(limiter RateLimiter) ClientOptionFunc {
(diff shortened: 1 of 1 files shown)


client_options.go
```diff
@@ -50,6 +50,22 @@ func WithCustomLimiter(limiter RateLimiter) ClientOptionFunc {
...
+	}
+}
+
+// WithCustomLogger can be used to configure a custom retryablehttp leveled logger
+func WithCustomLeveledLogger(leveledLogger retryablehttp.LeveledLogger) ClientOptionFunc {
+	return func(c *Client) error {
+		c.client.Logger = leveledLogger
+		return nil
+	}
+}
```
svanharmelen, line 67: Can you switch them to keep the order alphabetical? Thanks!

bmeneguele: Yes sure. Done!
However, WithCustomLeveledLogger < WithCustomLimiter, should we bring it before the Limiter too?
Or considering CustomLeveledLogger and CustomLogger are closely related it's fine just switching them?

svanharmelen: No, good point... Please make them all alphabetically ordered. Thanks, I missed that one 👍🏻

bmeneguele: Done!

svanharmelen (approved):
Thanks @bmeneguele! LGTM 🎉

PR #1122, opened 2021-04-03
add Group import export

previ:
Implements Group Import and Export API as defined in [GitLab docs](https://docs.gitlab.com/ee/api/group_import_export.html).

Methods are added to the [Group](https://github.com/previ/go-gitlab/blob/master/groups.go) type:

- [GroupExportRequest](https://github.com/previ/go-gitlab/blob/master/groups.go#L788)
- [GroupExportDownload](https://github.com/previ/go-gitlab/blob/master/groups.go#L812)
- [GroupImport](https://github.com/previ/go-gitlab/blob/master/groups.go#L846)


previ:
Hello @svanharmelen I proposed this PR which has been idle for a few weeks, I was asking if it's missing something in terms of quality or other. I also have a few other contributions (support for avatar upload for projects and groups). If you think it's worth, I may create a PR for that too. Thanks!

svanharmelen (approved):
Sorry this took so long @previ! It needed quite a few changes so I needed to make some time for this...

In the end I thought to just make the changes directly so it can be merged and you can see all the things that I would like to have changed before merging this one.

Thank you for the PR and adding support for this API 👍🏻

PR #1124, opened 2021-04-05
Add GetUser with_custom_attributes option

ax1036:
Add missing `with_custom_attributes` option for `GetUser` (need admin account)

Gitlab docs:

image

### DIFF
--- a/users.go
+++ b/users.go
@@ -147,8 +147,6 @@ func (s *UsersService) ListUsers(opt *ListUsersOptions, options ...RequestOption
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/users.html#single-user
 type GetUsersOptions struct {
-
-	// The options below are only available for admins.
 	WithCustomAttributes *bool `url:"with_custom_attributes,omitempty" json:"with_custom_attributes,omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


users.go
```diff
@@ -143,13 +143,22 @@ func (s *UsersService) ListUsers(opt *ListUsersOptions, options ...RequestOption
 	return usr, resp, err
 }
 
+// GetUsersOptions represents the available GetUser() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/users.html#single-user
+type GetUsersOptions struct {
+
+	// The options below are only available for admins.
```
svanharmelen, line 151:
```suggestion
```

svanharmelen (changes requested):

svanharmelen (approved):
Thanks @ax1036. While this is a breaking change, I'm afraid we have to change this to keep the package inline with the API. So LGTM 👍🏻

ax1036:
> Thanks @ax1036. While this is a breaking change, I'm afraid we have to change this to keep the package inline with the API. So LGTM 👍🏻

Yes, I hope this will not broke too many people code :ok_hand:

PR #1144, opened 2021-05-06
Add missing ReleasedAt field to Release struct

etiennetremel:
From [the docs](https://docs.gitlab.com/ee/api/releases/#list-releases), the ReleasedAt field was missing in the Releases struct.


svanharmelen (approved):
Thanks @etiennetremel 👍🏻

PR #1155, opened 2021-05-26
Avatar support

pkostohrys:
Adding gitlab avatar support https://docs.gitlab.com/ee/api/avatar.html

### DIFF
--- a/avatar.go
+++ b/avatar.go
@@ -20,15 +20,6 @@ import (
 	"net/http"
 )
 
-// AccessRequest represents a access request for a user avatars.
-//
-// GitLab API docs:
-// https://docs.gitlab.com/ee/api/avatar.html
-type AvatarOptions struct {
-	Email string `json:"email"`
-	Size  int    `json:"size"`
-}
-
 // AccessRequestsService handles communication with the avatar
 // avatar related methods of the GitLab API.
 //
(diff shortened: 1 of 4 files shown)


avatar.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (s *AvatarRequestsService) GetAvatar(opts *AvatarOptions, options ...RequestOptionFunc) (error, *string) {
+	req, err := s.client.NewRequest(http.MethodGet, "avatar", opts, options)
+	if err != nil {
+		return err, nil
+	}
+
+	avatar := struct {
+		Url string `json:"avatar_url"`
+	}{}
```
svanharmelen, line 48: I prefer to create and return a struct instead:

```
type Avatar struct {
    AvatarURL string `json:"avatar_url"`
}
```

avatar.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+// AccessRequest represents a access request for a user avatars.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/avatar.html
+type AvatarOptions struct {
+	Email string `json:"email"`
+	Size  int    `json:"size"`
+}
```
svanharmelen, line 30: These should be pointers, the struct should be places above the `GetAvatar` method and this the struct should be called `GetAvatarOptions`...

avatar.go
```diff
@@ -0,0 +1,56 @@
...
+	avatar := struct {
+		Url string `json:"avatar_url"`
+	}{}
+
+	_, err = s.client.Do(req, &avatar)
+	if err != nil {
+		return err, nil
+	}
+
+	return nil, &avatar.Url
```
svanharmelen, line 55: We always return the response as well.

svanharmelen (changes requested):
Thanks @pkostohrys, but it needs some work before we can merge it. Please checkout the rest of the codebase and try to make sure your PR is consistent with it 👍🏻

pkostohrys:
Thank @svanharmelen for you comments. I have pushed an update. Could you please check it now ?

svanharmelen (approved):
Thanks @pkostohrys! I've updated the code a little to fix the last inconsistencies. Please check the latest commit to see what I've changed 👍🏻

PR #1157, opened 2021-05-27
Escape tagName path in release related URLs

### LINKED ISSUE #1156: Support semver metadata in releases made from tags
Releases can have characters such as `+` in their name, e.g. `1.0.0-beta+exp.sha.5114f85`. When `ReleasesService` and `ReleaseLinksService` are used with such tags, the requests fail due to `+` being a special character in HTTP URLs. If it is escaped, it should work as expected.

sirlatrom:
Fixes #1156.

### DIFF
--- a/fixtures_test.go
+++ b/fixtures_test.go
@@ -244,6 +244,54 @@ const (
 		}
 	}`
 
+	// exampleReleaseResponse provides fixture for Releases tests.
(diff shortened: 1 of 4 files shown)


svanharmelen (approved):
Thanks @sirlatrom, LGTM 👍🏻

PR #1169, opened 2021-06-20
add note Type

devfd:
Add missing Note type, can be `DiscussionNote` or `DiffNote`.

### DIFF
--- a/notes.go
+++ b/notes.go
@@ -30,16 +30,30 @@ type NotesService struct {
 	client *Client
 }
 
+// NoteTypeValue represents the type of a Note
+type NoteTypeValue string
+
+// List of note type values.
+//
+// Though not present in the GitLab API docs the possible values can be found here
+// https://gitlab.com/gitlab-org/gitlab/-/blob/master/app/models/note.rb#L132
+const (
+	LegacyDiffNoteType NoteTypeValue = "LegacyDiffNote"
+	DiffNoteType       NoteTypeValue = "DiffNote"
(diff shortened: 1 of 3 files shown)


notes.go
```diff
@@ -35,6 +35,7 @@ type NotesService struct {
 // GitLab API docs: https://docs.gitlab.com/ce/api/notes.html
 type Note struct {
 	ID         int    `json:"id"`
+	Type       string `json:"type"`
```
devfd, line 38: the Gitlab documentation does not specify all possible values for this field, so keep a generic `string` type

svanharmelen: Maybe check the GitLab source code to find out and make a proper type for it? Also I don't see this field in the docs, can you point me to something that shows this field exists?

devfd: for exemple https://gitlab.com/api/v4/projects/278964/merge_requests/61843/notes, corresponding to this MR https://gitlab.com/gitlab-org/gitlab/-/merge_requests/61843

svanharmelen: Ok, I did some searching through the GitLab code and think we need to make a type (like this: https://github.com/xanzy/go-gitlab/blob/master/types.go#L27-L41) for these note types:

```
LegacyDiffNote
DiffNote
DiscussionNote
Note
```

devfd: ok thanks for the pointers. PR updated.

svanharmelen (approved):
Thanks @devfd! LGTM 👍🏻

PR #1173, opened 2021-06-28
Changed deployment api options to support Gitlab 14

### LINKED ISSUE #1171: GitLab v14.0 'List Project Deployments' API Breaking Changes
In version GitLab 14.0, there are some breaking changes that are happening to the `list project deployments` API. 

* `updated_after` is now `finished_after`.
* `updated_before` is now `finished_before`.
* Queries specifying the `updated_at` filter must set `order_by` to `updated_at`.

The affected code appears to be here: https://github.com/xanzy/go-gitlab/blob/master/deployments.go
Note the `ListProjectDeploymentsOptions` struct 

sources:
* https://about.gitlab.com/blog/2021/06/04/gitlab-moving-to-14-breaking-changes/
* https://docs.gitlab.com/ee/api/deployments.html#list-project-deployments

zk-kb4:
This should fix #1171, but it will be a breaking change to this library as any versions less than 14 will not work with these changes.

### DIFF
--- a/deployments.go
+++ b/deployments.go
@@ -72,12 +72,18 @@ type Deployment struct {
 // https://docs.gitlab.com/ce/api/deployments.html#list-project-deployments
 type ListProjectDeploymentsOptions struct {
 	ListOptions
-	OrderBy       *string    `url:"order_by,omitempty" json:"order_by,omitempty"`
-	Sort          *string    `url:"sort,omitempty" json:"sort,omitempty"`
+	OrderBy     *string `url:"order_by,omitempty" json:"order_by,omitempty"`
+	Sort        *string `url:"sort,omitempty" json:"sort,omitempty"`
+	Environment *string `url:"environment,omitempty" json:"environment,omitempty"`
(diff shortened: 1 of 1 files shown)


zk-kb4:
@svanharmelen If you need to create a `gitlab-14` branch or something else to merge this in to so that it doesn't mess with master quite yet, please feel free to change where this merges.

svanharmelen:
I guess one solution (for the time being) could be to add the new fields and leave the existing fields in place. And then order them as two subgroups each with a comment saying these are for versions <14 and these are for >=14...

zk-kb4:
@svanharmelen I've updated the PR with your suggestion.

svanharmelen (approved):
Looks good for now... Thanks!

zk-kb4:
@svanharmelen awesome, any idea on when you're planning on tagging a new release?

svanharmelen:
Just tagged master with `v0.50.1` 👍🏻

zk-kb4:
@svanharmelen Awesome, thank you!

PR #1101, opened 2021-03-09
Add GenericPackagesService

sirlatrom:
Fixes #1100.

TODO:
  - [x] Tests
  - [ ] Docs

Issue #1100, opened 2021-03-09
Support publishing and downloading generic packages


sirlatrom:
There are no separate API docs, but the API endpoints are documented here: https://docs.gitlab.com/ee/user/packages/generic_packages/

sirlatrom:
@svanharmelen What is your opinion on adding a non-API function to GenericPackagesService (added in this PR in the first place) which gives you the API download URL for a generic package? Or should other users of the library figure that out on their own after publishing a generic package? I'm specifically trying to add that for goreleaser/goreleaser#2006.

svanharmelen:
Go for it 👍🏻 I'm open to merging that PR when you finish it 😃

PR #1180, opened 2021-07-22
Add GenericPackagesServices (#1100)

neomantra:
This PR continues the work in #1101, rebasing it to master, updating its interface (returning download URL), and some real-world testing (fixed the URL endpoint and request type).

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ to add new and/or missing endpoints. Currently the following services are suppor
 - [x] Events
 - [x] Feature Flags
 - [ ] Geo Nodes
+- [x] Generic Packages
 - [x] GitLab CI Config Templates
 - [x] Gitignores Templates
 - [x] Group Access Requests
@@ -57,6 +58,7 @@ to add new and/or missing endpoints. Currently the following services are suppor
 - [x] Notes (comments)
 - [x] Notification Settings
 - [x] Open Source License Templates
+- [x] Packages
 - [x] Pages
 - [x] Pages Domains
 - [x] Personal Access Tokens
--- a/generic_packages.go
+++ b/generic_packages.go
@@ -97,10 +97,11 @@ func (s *GenericPackagesService) PublishPackageFile(pid interface{}, packageName
 		pathEscape(fileName),
 	)
 
-	// This is currently the only way to use a PUT request to upload a non-JSON file
+	// This is currently the only way to use a PUT request to upload a non-JSON file.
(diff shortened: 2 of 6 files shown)


matejvelikonja:
@svanharmelen  any chance for merging this?

dewey:
Thanks for working on it @neomantra! Hope this gets merged soon to unblock the goreleaser fix.

neomantra:
I just rebased this branch onto latest master and resolved the minor conflicts.

svanharmelen:
I'll have a look at this one tomorrow!

svanharmelen (commented):
I pushed a commit to tweak the PR and make it more inline with the rest of this package. I think the only thing that is still missing (which would make sense to add, but is not required), is adding support for the `select` option when publishing a package.

Please let me know what you think about the changes and if this will still work for you.

neomantra:
Thanks for the review and improvements.  I'm adding `select` and am testing this end-to-end  with `goreleaser` patches.

svanharmelen:
I again added a commit to complete the PR and I removed your last commit as I don't like to add and export custom logic, but only support calls that the Gitlab API supports.

Please confirm if this solution works for you so I can go ahead and merge it. Thanks!

neomantra:
Thanks for the review and update.   It's pretty convenient to have that `GetGenericPackageURL` function as the end-user needs to know the URL structure, which `gitlab-go` already knows as it needs it too; they must also include a function like `pathEscape`.   I also note that the `select` result of `GenericPackagesFile.File.URL` is not the same URL as `GetGenericPackageURL` but some AWS S3 URL (at least for gitlab.com).  

But I also appreciate a purist approach to the library.  I tested the latest against an updated `goreleaser` and it works.  I also tested the `select` function and it works.  Thanks again for the attention to this issue.

svanharmelen:
Hmm... Well OK then add your latest commit back so we can include that one after all...

neomantra:
OK, I added it back.  I changed the verb from `Get` to `Format`... `FormatPackageURL` .. to prevent confusion with the common `Get*` Gitlab API operations and documented that it doesn't make an API call.  I'm fine with another verb, but changed it given your feedback.

svanharmelen:
Nice, I can appreciate that one! Now if you can only swap the `PublishPackageFileOptions` and the `FormatPackageURL` method so that `PublishPackageFileOptions` is just above the method that uses it, I think we are good to go!

svanharmelen (approved):
Thanks for your patience with this one @neomantra 👍🏻

PR #1183, opened 2021-07-28
Added UserUsername (sic) to tag push events.

apollo13:
While the documentation does not list the field user_username in
https://docs.gitlab.com/ce/user/project/integrations/webhooks.html#tag-events
it was added 4 years ago in https://gitlab.com/gitlab-org/gitlab-foss/-/merge_requests/11272
and can still be seen in: https://gitlab.com/gitlab-org/gitlab-foss/-/blob/9e16447c7c4cec7373bb2df0d94b4dc8b8300b08/lib/gitlab/data_builder/push.rb#L123

I have also verified locally that the webhook does indeed receive the
field in question for a tag_push.

### DIFF
--- a/event_parsing_webhook_test.go
+++ b/event_parsing_webhook_test.go
@@ -381,7 +381,11 @@ func TestParseTagHook(t *testing.T) {
 	}
 
 	if event.UserName != exampleEventUserName {
-		t.Errorf("Username is %s, want %s", event.UserName, exampleEventUserName)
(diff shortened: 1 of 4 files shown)


apollo13:
This commit will be required for https://github.com/mattermost/mattermost-plugin-gitlab/issues/254

svanharmelen (approved):
Thanks @apollo13 LGTM 👍🏻

PR #1188, opened 2021-08-11
[minor] add user availability to UserStatus

costela:
This minor PR adds support for the `availability` field in `UserStatus`, added in 13.6.0.

### DIFF
--- a/users.go
+++ b/users.go
@@ -925,20 +925,11 @@ func (s *UsersService) GetUserActivities(opt *GetUserActivitiesOptions, options 
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/users.html#user-status
 type UserStatus struct {
-	Emoji        string `json:"emoji"`
-	Availability string `json:"availability"`
-	Message      string `json:"message"`
-	MessageHTML  string `json:"message_html"`
+	Emoji        string            `json:"emoji"`
+	Availability AvailabilityValue `json:"availability"`
+	Message      string            `json:"message"`
+	MessageHTML  string            `json:"message_html"`
 }
-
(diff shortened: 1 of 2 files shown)


users.go
```diff
@@ -925,9 +925,10 @@ func (s *UsersService) GetUserActivities(opt *GetUserActivitiesOptions, options
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/users.html#user-status
 type UserStatus struct {
-	Emoji       string `json:"emoji"`
-	Message     string `json:"message"`
-	MessageHTML string `json:"message_html"`
+	Emoji        string `json:"emoji"`
+	Availability string `json:"availability"`
```
svanharmelen, line 929: Is there a fixed list of availability values? If so we should probably make a `const` for it instead of using a `string`

costela: good point. I tried to find one (for about 5 minutes :see_no_evil:) and came up empty.
I'll see what gitlab's code has to say about it :+1:

costela: ok, it seems gitlab [currently has only two possible options](https://gitlab.com/gitlab-org/gitlab-foss/-/blob/3cfe2ab491cdaf46d5a8ff6359fc6a89a7557eea/app/models/user_status.rb#L22): `not_set` and `busy`.

added as consts :heavy_check_mark:

costela:
FYI: the API docs for _setting_ the availability is [currently do not mention the `availability` parameter](https://docs.gitlab.com/ce/api/users.html#set-user-status), but it works anyway, so I included it in the PR.

svanharmelen (approved):
Thanks @costela! Made a few last tweaks, but looks good now 👍🏻

PR #1189, opened 2021-08-16
Add GetGroupOptions

Starttoaster:
The `/groups/:id` endpoint actually has a limited set of options available which aren't reflected in the library currently. See: https://docs.gitlab.com/ee/api/groups.html#details-of-a-group

Also noting that this is a backwards incompatible change since the GetGroup function didn't previously require an options struct sent with it. Shouldn't affect users that pin to their specific versions with go modules, but to those upgrading it might be a tiny bit of surprise cleanup.

### DIFF
--- a/groups.go
+++ b/groups.go
@@ -127,8 +127,8 @@ func (s *GroupsService) ListGroups(opt *ListGroupsOptions, options ...RequestOpt
 // GitLab API docs: https://docs.gitlab.com/ce/api/groups.html#details-of-a-group
 type GetGroupOptions struct {
 	ListOptions
-	WithProjects         *string `url:"with_projects,omitempty" json:"with_projects,omitempty"`
-	WithCustomAttributes *bool   `url:"with_custom_attributes,omitempty" json:"with_custom_attributes,omitempty"`
+	WithProjects         *bool `url:"with_projects,omitempty" json:"with_projects,omitempty"`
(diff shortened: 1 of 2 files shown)


groups.go
```diff
@@ -122,17 +122,26 @@ func (s *GroupsService) ListGroups(opt *ListGroupsOptions, options ...RequestOpt
 	return g, resp, err
 }
 
+// GetGroupOptions represents the available GetGroup() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ce/api/groups.html#details-of-a-group
+type GetGroupOptions struct {
+	ListOptions
+	WithProjects         *string `url:"with_projects,omitempty" json:"with_projects,omitempty"`
```
svanharmelen, line 130: This should be a `*bool`

groups.go
```diff
@@ -122,17 +122,26 @@ func (s *GroupsService) ListGroups(opt *ListGroupsOptions, options ...RequestOpt
...
+	WithCustomAttributes *bool   `url:"with_custom_attributes,omitempty" json:"with_custom_attributes,omitempty"`
```
svanharmelen, line 131: Please order these using the ordering in the docs...

groups.go
Starttoaster: Totally agree. Changed that :)

groups.go
Starttoaster: My apologies, ordered them per the docs.

svanharmelen (approved):
Thanks, LGTM 👍🏻

PR #1191, opened 2021-08-17
GroupMember ExpiresAt wrong type

Starttoaster:
The API return for the ExpiresAt date is actually ISOTime. See the below example query and response:

```
% curl -s -H "Private-Token: [redacted]" "https://gitlab.com/api/v4/groups/[redacted]/members" | jq
[
  {
    "id": [redacted],
    "name": "Brandon Butler",
    "username": "[redacted]",
    "state": "active",
    "avatar_url": "https://gitlab.com/uploads/-/system/user/avatar/[redacted]/avatar.png",
    "web_url": "https://gitlab.com/[redacted]",
    "access_level": 50,
    "created_at": "2020-09-04T18:21:04.197Z",
    "expires_at": "2021-08-18"
  }
]
```

The docs actually seem to be wrong about the returned time format but I can confirm it's either ISOTime or null. Wouldn't make any sense for it to be a timestamp because there's no option to specify a specific expiration time anyway, only the date. And gitlab fulfills the expiration at midnight (I believe UTC.)

### DIFF
--- a/group_members.go
+++ b/group_members.go
@@ -52,7 +52,7 @@ type GroupMember struct {
 	AvatarURL         string                   `json:"avatar_url"`
 	WebURL            string                   `json:"web_url"`
 	CreatedAt         *time.Time               `json:"created_at"`
-	ExpiresAt         *time.Time               `json:"expires_at"`
+	ExpiresAt         *ISOTime                 `json:"expires_at"`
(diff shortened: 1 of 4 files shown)


Starttoaster:
Deep sigh

PR #1193, opened 2021-08-20
Add Topics to Project

lwalen:
`tag_list` was deprecated in GitLab 14. https://docs.gitlab.com/ee/api/projects.html

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -48,6 +48,7 @@ type Project struct {
 	HTTPURLToRepo                             string                     `json:"http_url_to_repo"`
 	WebURL                                    string                     `json:"web_url"`
 	ReadmeURL                                 string                     `json:"readme_url"`
+	Topics                                    []string                   `json:"topics"`
 	TagList                                   []string                   `json:"tag_list"`
 	Owner                                     *User                      `json:"owner"`
 	Name                                      string                     `json:"name"`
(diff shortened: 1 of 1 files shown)


projects.go
```diff
@@ -48,6 +48,7 @@ type Project struct {
 	HTTPURLToRepo                             string                     `json:"http_url_to_repo"`
 	WebURL                                    string                     `json:"web_url"`
 	ReadmeURL                                 string                     `json:"readme_url"`
+	Topics                                    []string                   `json:"topics"`
```
svanharmelen, line 51: These fields are ordered using to ordering in the docs. For this field that means it should be placed between just below the `TagList` field.

svanharmelen (changes requested):
One super minor change request, but looks fine otherwise 👍🏻

projects.go
lwalen: Done, thanks!

svanharmelen (approved):
Thanks @lwalen 👍🏻

Deichindianer:
Is there any indication when this might be released? Currently waiting for this feature to write some systems around topics.

If there is something left to do, I'm also happy to try and help out :)

svanharmelen:
Just released version `v0.50.4`

PR #1194, opened 2021-08-23
Fix create project struct to align with API

jguttman94:
The existing struct `CreateProjectOptions` has fields that are not supported within the API. These values have been removed. This PR also contains some additions to keep up with the support options.

https://docs.gitlab.com/ee/api/projects.html#create-project


svanharmelen (approved):
Thanks @jguttman94! LGTM 👍🏻

PR #1202, opened 2021-09-03
Add JOB-TOKEN support

denisvmedia:
Container Registry API can be used within a Job using $CI_JOB_TOKEN. When using $CI_JOB_TOKEN, one should set `JOB-TOKEN` header instead of `PRIVATE-TOKEN` .

See: https://docs.gitlab.com/ee/api/container_registry.html

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -58,9 +58,9 @@ type authType int
 // GitLab API docs: https://docs.gitlab.com/ce/api/
 const (
 	basicAuth authType = iota
+	jobToken
 	oAuthToken
 	privateToken
-	jobToken
 )
 
 // A Client manages communication with the GitLab API.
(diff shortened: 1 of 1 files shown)


gitlab.go
```diff
@@ -60,6 +60,7 @@ const (
 	basicAuth authType = iota
 	oAuthToken
 	privateToken
+	jobToken
```
svanharmelen, line 63: Please place this in the correct alphabetical order to keep things consistent.

gitlab.go
```diff
@@ -214,6 +215,18 @@ func NewClient(token string, options ...ClientOptionFunc) (*Client, error) {
...
+// authentication, provide a valid job token.
+func NewJobClient(token string, options ...ClientOptionFunc) (*Client, error) {
+	client, err := newClient(options...)
+	if err != nil {
+		return nil, err
+	}
+	client.authType = jobToken
+	client.token = token
+	return client, nil
+}
```
svanharmelen, line 228: Same about the ordering. Please place this below the `NewBasicAuthClient` function.

gitlab.go
```diff
@@ -649,6 +662,8 @@ func (c *Client) Do(req *retryablehttp.Request, v interface{}) (*Response, error
 		req.Header.Set("Authorization", "Bearer "+c.token)
 	case privateToken:
 		req.Header.Set("PRIVATE-TOKEN", c.token)
+	case jobToken:
+		req.Header.Set("JOB-TOKEN", c.token)
```
svanharmelen, line 666: Again ordering 😏

svanharmelen (changes requested):
Assuming you have tested this and it works as expected, I'll take this once you've updated the ordering to keep things consistent. Thanks!

gitlab.go
denisvmedia: I put it in this order in order to not introduce potential backwards incompatibility. Although, it's an unexported field, I thought, it could be dangerous to put it before. But as you think it's fine, I will change it.

svanharmelen (approved):
Thanks @denisvmedia, LGTM!

PR #1203, opened 2021-09-03
Fix value assignment

jguttman94:
Fix silly issue mapping off -> on and on -> off


svanharmelen (approved):
Nice catch! Thanks @jguttman94 👍🏻

jguttman94:
Thanks! Sorry for making the mistake the first time around.

PR #1206, opened 2021-09-04
Added Tests for Namespaces

Urento:
According to Docs: https://docs.gitlab.com/13.12/ee/api/namespaces.html and https://docs.gitlab.com/ee/api/namespaces.html


svanharmelen (approved):
Again tweaked it a little, but looks fine like this 👍🏻

PR #1208, opened 2021-09-08
Support deploy keys in protected branches

mottetm:
On top of allowing specific users and groups to push to protected branches, Gitlab also gives the option to allow deploy keys using the `deploy_key_id` field.

image

### DIFF
--- a/protected_branches.go
+++ b/protected_branches.go
@@ -138,6 +138,7 @@ type ProtectRepositoryBranchesOptions struct {
 type BranchPermissionOptions struct {
 	UserID      *int              `url:"user_id,omitempty" json:"user_id,omitempty"`
 	GroupID     *int              `url:"group_id,omitempty" json:"group_id,omitempty"`
+	DeployKeyID *int              `url:"deploy_key_id,omitempty" json:"deploy_key_id,omitempty"`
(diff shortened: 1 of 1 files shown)


protected_branches.go
```diff
@@ -138,6 +138,7 @@ type ProtectRepositoryBranchesOptions struct {
 type BranchPermissionOptions struct {
 	UserID      *int              `url:"user_id,omitempty" json:"user_id,omitempty"`
 	GroupID     *int              `url:"group_id,omitempty" json:"group_id,omitempty"`
+	DeployKeyId *int              `url:"deploy_key_id,omitempty" json:"deploy_key_id,omitempty"`
```
svanharmelen, line 141: Can you show me where in the docs or the code this field is mentioned? And for it to be a proper Go field name, it should be `DeployKeyID`

mottetm: Hi @svanharmelen,
I renamed the field.
The feature is documented [here](https://docs.gitlab.com/ee/user/project/protected_branches.html#allow-deploy-keys-to-push-to-a-protected-branch). They however have failed to update the API documentation when adding the feature. [Here](https://gitlab.com/gitlab-org/gitlab/-/issues/30769) is a link to the issue that introduced the change with all the related work.

svanharmelen (approved):
Cool, thanks @mottetm 👍🏻

PR #1211, opened 2021-09-10
#1210 add WithToken for change token

### LINKED ISSUE #1210: Request with user token
If we need to make requests from different users, then in the current situation we will have to recreate the client.

eaglemoor:
Fix #1210 

For use user token we need change token


svanharmelen:
I see what you are trying to do (and what you are trying to solve), but I'm not sure if I like this approach. I need to think a little bit about this one...

korotin:
@svanharmelen there could be an alternative solution: we may add to `gitlab.Client` new method `WithToken()` which will return a new instance of client with updated token.

Either way it's a nice feature to have: switching between tokens is a little bit painful now.

eaglemoor:
@svanharmelen any update?

svanharmelen:
Sorry for the delay... But I think I prefer to use a new option like you suggested (`WithToken`) instead of the current solution.

svanharmelen:
Please check the commit I just added, this is more what I was thinking about. This would still solve your use case, right?

eaglemoor:
@svanharmelen we have problem with your code.

```
const (
	basicAuth authType = iota
	jobToken
	oAuthToken
	privateToken
)
```

I can't use private type and const for this method. And code `gitlab.WithToken(2, token),` seems not good.

eaglemoor:
And this solution not well as use context. For example I have one func who call some GitLab API.
```
func GetProjectInfo(ctx context.Context, token, projectID string) ... {
  author := GetProejctAuthor(ctx, token, projectID)
  tags := GetProjectTags(ctx, token, projectID)
  
  // or
  
    reqOpts := []gitlab.RequestOptionFunc{
	gitlab.WithToken(OAuthToken, token),
	gitlab.WithContext(ctx),
    }
    
    author := GetProjectAuthor(ctx, reqOpts, projectID string)
    tags := GetProjectTags(ctx, reqOpts, projectID string)
}

func GetProjectAuthor(ctx context.Context, token string, projectID string) ... {
    reqOpts := []gitlab.RequestOptionFunc{
	gitlab.WithToken(OAuthToken, token),
	gitlab.WithContext(ctx),
    }
}

func GetProjectTags(ctx context.Context, token string, projectID string) ... {
    reqOpts := []gitlab.RequestOptionFunc{
	gitlab.WithToken(OAuthToken, token),
	gitlab.WithContext(ctx),
    }
}
```

In all func, when I use call api I need send context + reqOpts or token.

svanharmelen:
Good point about the fact that `authType` is private. That's a mistake, it should of course be public to make this work. I will update the code for that.

But I don't understand your other comment. What is the problem there?

svanharmelen:
Updated the code to export the `AuthType`. And reading you second comment again, I think I now understand your point.

Your comment is about the fact that you now have to send both the context and the token or request options, instead of only the context right? I understand, but I don't want to rely on the context as there are also projects that do not use a context when making API calls.

In your case you can still use your own context and put the token in there to send it around. And then in the function where you make the API call you can extract it from your context and put it into the request options.

eaglemoor:
> but I don't want to rely on the context as there are also projects that do not use a context when making API calls.

Can we stile both variant?

```
func WithToken(authType AuthType, token string) RequestOptionFunc{}
func WithContext(ctx context.Context) RequestOptionFunc{}

func ContextWithAuth(ctx context.Context, authType AuthType, token string) context.Context {}
```

but need to prioritize this methods.

svanharmelen:
I don't prefer to add both variants, sorry. I personally wouldn't even have added the `WithToken` option as I see the token as part of the client config. So if you need multiple connections with multiple tokens I would say just create multiple clients.

Yet I don't want to make things harder as they need to be, so adding `WithToken` seems like the right compromise.

eaglemoor:
Ok. It's good for me, thx. Waiting for new release )

Issue #636, opened 2019-06-13
Error when importing project


spencergilbert:
`Post https://gitlab.com/api/v4/projects/import: http: ContentLength=141 with Body length 0`

I'm receiving the above error when trying to import a project to gitlab.com

```
git := gitlab.NewClient(nil, "<token>")

	file := "/path/to/exported/project.tar.gz"
	importPath := "imported-project"
	namespace := "my-namepsace"

	opt := gitlab.ImportFileOptions{
		Namespace: &namespace,
		File:      &file,
		Path:      &importPath,
	}

	status, resp, err := (*git.ProjectImportExport).ImportProject(&opt)
	if err != nil {
		fmt.Println(status)
		fmt.Println(resp)
		log.Fatal(err)
	}
```

This returns `<nil>`, `<nil>`, and the above error. Any pointers?

svanharmelen:
When having a closer look at the API it looks like uploading a file from your local filesystem is indeed not supported by the current implementation.

I think PR #621 was not been fully tested before it was created as the import file API requires the file to be uploaded using a multipart message which is currently not something that is supported by this package.

spencergilbert:
Awesome, wasn't sure if I had been misusing the API there or not. I ended up just using the standard library there for a workaround

davidcvan:
Hey @spencergilbert would you mind posting the workaround code.
I tried to work around it but it doesnt seem to work.
Thanks

nvanheuverzwijn:
@svanharmelen If I open a PR to fix this issue, will you do the review ? I have a semi working solution for me but I'd prefer to see that in the upstream lib instead of my hack.

svanharmelen:
Of course! PR's are always welcome 👍🏻

dheitman42:
Just wanted to check if there has been any progress on this, and to give my +1.  I spent a day trying to figure out why my imports were failing.  Maybe there should be a comment or something in the docs to indicate that this doesn't work in the meantime, since anyone reading the docs would assume it works as advertised.

dheitman42:
Put up a PR here: https://github.com/xanzy/go-gitlab/pull/1215

Did some quick and dirty testing, but I'll test it more thoroughly after the weekend.

svanharmelen:
This issue should be solved by PR #1318 which fixes a couple of upload issues.

PR #1215, opened 2021-09-11
Fix ImportFile() to send a multipart form request

dheitman42:
Update ImportFile() to send a multipart form request, add missing "Name" parameter to ImportFileOptions.

### DIFF
--- a/project_import_export.go
+++ b/project_import_export.go
@@ -18,8 +18,13 @@ package gitlab
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
+	"io"
+	"mime/multipart"
 	"net/http"
+	"os"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


dheitman42:
I'll test this more thoroughly after the weekend.

dheitman42:
Note this is intended to resolve https://github.com/xanzy/go-gitlab/issues/636

project_import_export.go
```diff
@@ -173,10 +179,41 @@ type ImportFileOptions struct {
...
+		if strings.ToLower(key) == "file" {
+			part, err := writer.CreateFormFile("file", file.Name())
+			if err != nil {
+				return nil, nil, err
+			}
+			_, err = io.Copy(part, file)
+		} else {
+			_ = writer.WriteField(strings.ToLower(key), fmt.Sprintf("%v", value))
+		}
+	}
```
svanharmelen, line 206: No clue what you are trying to do here, but this doesn't look right to me. Can you please elaborate on this approach?

dheitman42: Sure, we're creating a multipart form, creating the file component with CreateFormFile(), copying the file contents to that part, then adding each of the fields from the ImportFileOptions struct as additional field data in the multipart form.  That's the intent, anyway.  I'll admit to being n00bish with golang, so maybe this isn't the proper way to do the struct-to-string-map conversion, but that's the basic idea.

dheitman42: Dunno if that answered your question, happy to clarify further if need be.

svanharmelen:
This should be fixed by PR #1318 which also adds some additional file upload improvements. Please open an issue if you have any issues with the new code. Thanks for the PR and the discussions!

PR #1218, opened 2021-09-16
add tests for strings

venky333:
* Added tests for strings


strings_test.go
```diff
@@ -0,0 +1,50 @@
...
+type Company struct {
+	Name string
+	Address Address
+	Country string
+}
+
+var (
+	response       = "gitlab.Person{Name:\"name\", Age:16, NickNames:[\"n\" \"a\" \"m\" \"e\"], Address:gitlab.Address{Street:\"\", City:\"\", Province:\"\", Country:\"\"}}"
+	emptySliceResp = "gitlab.Person{Name:\"name\", Age:16, Address:gitlab.Address{Street:\"\", City:\"\", Province:\"\", Country:\"\"}}"
+)
```
svanharmelen, line 32: Can these be formatted for readability and maybe use backtics instead of escaping the quotes? Or does that break the tests?

svanharmelen: And since all of these are used only once, place declare them inside the test instead of making these globals. Thanks!

svanharmelen (changes requested):

strings_test.go
venky333: * backticks will break the tests. So not changing anything there.
* declared them inside test.

svanharmelen (approved):
Thanks 👍🏻

PR #1219, opened 2021-09-16
Add allow_failure field to Pipeline webhook event

vntw:
As documented in https://docs.gitlab.com/ee/user/project/integrations/webhooks.html#pipeline-events, it's part of `builds` in pipeline events:

```json
{
  "builds": [
	{
	  "id": 380,
	  "stage": "deploy",
	  "name": "production",
	  "status": "skipped",
	  "created_at": "2016-08-12 15:23:28 UTC",
	  "started_at": null,
	  "finished_at": null,
	  "when": "manual",
	  "manual": true,
 >	  "allow_failure": false,
	  "user": {
		"id": 1,
		"name": "Administrator",
		"username": "root",
		"avatar_url": "http://www.gravatar.com/avatar/e32bd13e2add097461cb96824b7a829c?s=80\u0026d=identicon",
		"email": "admin@example.com"
	  },
	  "runner": null,
	  "artifacts_file": {
		"filename": null,
		"size": null
	  },
	  "environment": {
		"name": "production",
		"action": "start",
		"deployment_tier": "production"
	  }
	}
  ],
}
```

Thank you for the support and the quick merges 🙂

### DIFF
--- a/event_webhook_types.go
+++ b/event_webhook_types.go
@@ -678,17 +678,18 @@ type PipelineEvent struct {
 		} `json:"author"`
 	} `json:"commit"`
 	Builds []struct {
-		ID         int        `json:"id"`
-		Stage      string     `json:"stage"`
-		Name       string     `json:"name"`
-		Status     string     `json:"status"`
-		CreatedAt  string     `json:"created_at"`
-		StartedAt  string     `json:"started_at"`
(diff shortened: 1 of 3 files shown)


svanharmelen (approved):
Thanks @vntw!

PR #1225, opened 2021-09-18
Escaping namespace path

levkoburburas:
When it tries to get nested namespace, the path has to be escaped otherwise an error is occured.


svanharmelen (approved):
Thanks 👍🏻

PR #1228, opened 2021-09-20
updated commit id with X-Gitlab-Commit-Id instead of X-Gitlab-Last-Co…

venky333:
updated commit id with X-Gitlab-Commit-Id instead of X-Gitlab-Last-Commit-Id in GetFileMetaData function


svanharmelen:
Do you have a link to the docs about this one? To make sure it's correct...

venky333:
Check `In addition to the GET method, you can also use HEAD to get just file metadata.` section of the following link: https://docs.gitlab.com/ee/api/repository_files.html#get-file-from-repository

svanharmelen (approved):
Check, thanks @venky333! I missed that we used the same header twice 👍🏻

PR #1233, opened 2021-09-23
Fix SquashOption type

mkjmdski:
Hi I just noticed that I used incorrect type in my previous pull request, it doesn't allow to convert const into string in dependent repo. Sorry for that. Could you release patch after merge? Thanks


svanharmelen (approved):
Hmm... Missed that 😞

PR #1235, opened 2021-09-24
Changed User struct - Add JobTitle field

OrlovDiga:
Hello, you make good lib for GitLab API, but I need also one field to User struct - JobTitle. Can you approve the pull request?

### DIFF
--- a/users.go
+++ b/users.go
@@ -70,6 +70,7 @@ type User struct {
 	Twitter                        string             `json:"twitter"`
 	WebsiteURL                     string             `json:"website_url"`
 	Organization                   string             `json:"organization"`
+	JobTitle                       string             `json:"job_title"`
 	ExternUID                      string             `json:"extern_uid"`
 	Provider                       string             `json:"provider"`
 	ThemeID                        int                `json:"theme_id"`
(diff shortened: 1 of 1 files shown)


svanharmelen (changes requested):
I see you also have renamed `xanzy` to `OrlovDiga`, please revert that... And if you only did that for local testing, you should probably check out this one: https://golang.org/ref/mod#go-mod-file-replace

In short, there is no need to rename all the import paths if you want to test with a forked version of the module. And one more tip, always make PR's from a dedicated branch instead of from your `master` branch. That way you can continue to make changes to your `master` without that having impact on the PR...

users.go
```diff
@@ -185,6 +186,7 @@ type CreateUserOptions struct {
 	Twitter             *string `url:"twitter,omitempty" json:"twitter,omitempty"`
 	WebsiteURL          *string `url:"website_url,omitempty" json:"website_url,omitempty"`
 	Organization        *string `url:"organization,omitempty" json:"organization,omitempty"`
+	JobTitle            *string `json:"job_title,omitempty"`
```
svanharmelen, line 189: You missed the `url` tag here...

OrlovDiga: Change it!

svanharmelen (approved):
Thanks @OrlovDiga , LGTM 👍🏻

PR #1249, opened 2021-09-29
Add Title field to PushEvent and TagEvent structs

graillus:
This PR adds the `Title` field to the `PushEvent.Commits` and `TagEvent.Commits` webhook structs, populated from the corresponding Gitlab event payloads :

- https://docs.gitlab.com/ee/user/project/integrations/webhooks.html#push-events
- https://docs.gitlab.com/ee/user/project/integrations/webhooks.html#tag-events

I'm currently working on a project and I need to get the commit titles from the webhook event.
Instead of using the `Message` field (which can contain several lines), I'd like to rely on the `Title` field, the one which is used by Gitlab itself in the commit history.

### DIFF
--- a/event_webhook_types.go
+++ b/event_webhook_types.go
@@ -742,6 +742,7 @@ type PushEvent struct {
 	Commits    []*struct {
 		ID        string     `json:"id"`
 		Message   string     `json:"message"`
+		Title     string     `json:"title"`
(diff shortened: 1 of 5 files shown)


svanharmelen (approved):
Thanks @graillus, LGTM 👍🏻

PR #1252, opened 2021-09-30
renamed groups_badges_test to group_badges_test

venky333:
`group_badges_test` was named as `groups_badges_test`. So just to maintain consistency and maintain order renamed it accordingly


svanharmelen (approved):
Nice catch 👍🏻

PR #1254, opened 2021-10-01
add tests for group_boards and correct the reference of array in Upda…

venky333:
* add tests for group_boards and correct the reference of array in UpdateIssueBoardList method


svanharmelen (approved):
Nice catch! Thanks!

PR #1257, opened 2021-10-01
adding variables to pipeline event webhook object attributes

### LINKED ISSUE #1120: Declare 'Variables' field as subfield of ObjectAttributes in PipelineEvent struct
Per https://docs.gitlab.com/ce/user/project/integrations/webhooks.html#pipeline-events, webhook pipeline events are have a `variables` field in `object_attributes`, so the PipelineEvent struct seems to need a similar field, e.g.:

type PipelineEvent struct {
    …
    ObjectAttributes struct {
	…
	Variables []map[string]string   `json:”variables"`
	…

M0roSan:
close #1120 

Note: not tested against actual webhook. Based on the payload, this works. I'm not sure how it behaves when a variable is masked. let me know if you need me to test on that behavior

### DIFF
--- a/event_webhook_types.go
+++ b/event_webhook_types.go
@@ -640,6 +640,10 @@ type PipelineEvent struct {
 		FinishedAt     string   `json:"finished_at"`
 		Duration       int      `json:"duration"`
 		QueuedDuration int      `json:"queued_duration"`
+		Variables      []struct {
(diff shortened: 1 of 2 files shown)


svanharmelen (approved):
Thanks @M0roSan! LGTM 👍🏻

Issue #1261, opened 2021-10-02
External Status Checks


M0roSan:
external status checks is added in 14.0 behind FF and FF is removed on 14.1. 
This issue is to implement the client for https://docs.gitlab.com/ee/api/status_checks.html

PR #1262, opened 2021-10-02
service and list status check for a MR

M0roSan:
#1261 

- list status checks for a MR

### DIFF
--- a/external_status_checks.go
+++ b/external_status_checks.go
@@ -1,0 +1,46 @@
+package gitlab
+
+import (
+	"fmt"
+	"net/http"
+)
+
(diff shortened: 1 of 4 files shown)


M0roSan:
@svanharmelen can you take a look at your convenience?

For now, I'm adding list APIs as I don't have time to implement update/delete. they are coming in subsequent MR(s).

I'm not huge fun of `StatusCheckProtectedBranch` struct.
- I initially have it as unnamed struct and embed it in `ProjectStatusCheck` but problem while testing that I need to re-write the same struct
- there is no other (protected) branch struct that has the same attributes defined for external status check. I would've just reused if there were any.

PR #1264, opened 2021-10-03
add tests for boards

venky333:
* add tests for CreateIssueBoard, UpdateIssueBoard, DeleteIssueBoard, ListIssueBoards, GetIssueBoard in boards


svanharmelen (changes requested):
All the fields that use the [zero value](https://golangbyexample.com/go-default-zero-value-all-types/) of the type, don't have to be written out explicitly. Especially with the tests in this file that would save a whole lot of space, so could you update that?

svanharmelen (approved):
Thanks 👍🏻

PR #1267, opened 2021-10-04
add tests for audit_events

venky333:
* add tests for audit_events


svanharmelen (approved):
Same here about writing out the zero values explicitly... But in this file it's not too bad (in terms of the required space).

PR #1268, opened 2021-10-04
removed explicitly defining default values as per review comment

venky333:
* removed explicitly defining default values as per review comment


svanharmelen (approved):
Nice 👍🏻

PR #1274, opened 2021-10-12
Add repository storage to project struct and ListProjects

catouc:
Came across this during a large scale storage migration you have to extend the structs manually if you want to get the fields into a project struct to work with it.

Tests pass locally for me.

Anything else I need to do? :)


Deichindianer:
Just added those two :) Didn't quite see those thanks for the hint!

Issue #1168, opened 2021-06-17
Rename WIP status to Draft


bmeneg:
GitLab is planning to fully remove the `WIP` word from their `Draft` status [1] in their 14.0 version, which is about to be released.
It still not clear what's going to happen in the API side, but it would be nice to rename the `WIP` struct member from `ListMergeRequestsOptions` to `Draft` to properly represent the terminology.

```
Adding WIP: to the start of the merge request’s title still marks a merge request as a draft.
This feature is scheduled for removal in GitLab 14.0. Use Draft: instead.
```
Source: [2]

[1] https://gitlab.com/gitlab-org/gitlab/-/issues/228685
[2] https://docs.gitlab.com/ee/user/project/merge_requests/drafts.html

bvicinus:
documentation on how WIP is being deprecated in favor of Draft, along with a slew of other changes. 

https://about.gitlab.com/blog/2021/06/04/gitlab-moving-to-14-breaking-changes/

Deichindianer:
Did create the field in my PR.
Tests run but I think there is more discussion to be had around what this really means, the docs are a bit confusing as well as the behaviour of GitLab itself on this one :thinking:

svanharmelen:
"fixed" by #1283

PR #1283, opened 2021-11-01
Add the `Draft` parameter for merge request structs

Deichindianer:
Currently, in GitLab 14.4.0 the `draft` url parameter is not implemented -
and I'm still unsure whether it should be? You still filter for non-draft
merge requests with `wip=yes`. If GitLab ever removes the `wip` url param
and replaces it with `draft` the typing of this field becomes a difficult
question I think, might be wrong here.

This should be backwards compatible due to the `omitempty` being in both the
`Draft` and `WIP` fields of the merge request structs.

Also adjusted the tests to use `Draft:` instead of `WIP:`. It might be worth
considering having tests for both but that would raise questions on which
behaviour to expect - the removed `WIP:` not working or until it is still in
supported versions to maintain backwards compatibility.

(This tries to work towards #1168)


bmeneg:
The API doc is indeed quite confusion regarding the use of `draft`: there are no mention of `?draft=` instead of `?wip=` parameter and  there are nothing about the `draft:` field in the response body other than its presence in the snippets. However, I agree and like the approach you took of adding the `Draft` struct member instead of renaming/substituting `WIP` or `WorkInProgress`. It seems good to me.

svanharmelen (approved):
I'll take this one... Thanks @Deichindianer

PR #1291, opened 2021-11-23
fix: update pipeline webhook event type payload

brewkode:
This change brings in `Environment` struct as part of the Pipeline Event Payload. Recent versions of the gitlab api's pipeline event payload contain the environment details which would be useful to expose via this sdk.


svanharmelen (approved):
LGTM 👍🏻

Issue #1292, opened 2021-11-24
UpdateProjectLevelRuleOptions should not omit empty GroupIDs/UserIDs


QuingKhaos:
### SDK Version

```
// go.mod
github.com/xanzy/go-gitlab v0.51.1
```

```
// go.sum
github.com/xanzy/go-gitlab v0.51.1 h1:wWKLalwx4omxFoHh3PLs9zDgAD4GXDP/uoxwMRCSiWM=
github.com/xanzy/go-gitlab v0.51.1/go.mod h1:Q+hQhV508bDPoBijv7YjK/Lvlb4PhVhJdKqXVQrUoAE=
```

### Expected Behavior

When calling `client.Projects.UpdateProjectApprovalRule` with empty `GroupIDs` or empty `UserIDs` it should send an empty attribute to the API and thus removing all groups/users from the project-level merge request approval rule.

### Actual Behavior

Due to the `omitempty` attribute in the `UpdateProjectLevelRuleOptions` struct, an empty group/user list is not send in the PUT request to the API and I'm not able to remove all groups or users while updating the project-level merge request approval rule.

### References

- #700 
- #717 
- gitlabhq/terraform-provider-gitlab#755 - the issue in the GitLab Terraform Provider raising this issue

svanharmelen:
PR #1294 is a nasty breaking change, but it does fix this issue...

QuingKhaos:
Nasty 😅 

I don't know Go well enough, but with a pointer `omitempty` then omits it only, when the pointer is empty and not the list?

With a BC like that, that will take the GitLab Terraform Provider a while to catch up ^^

svanharmelen:
Yeah correct about the behavior of `omitempty` with a pointer... And it's really simple and straight forward to fix, so I would be surprised if it would take a while for the TF provider to update to the latest version.

QuingKhaos:
Straightforward, but still needs time and a new release ^^

svanharmelen:
This should be fixed by PR #1294 which is just merged...

PR #1294, opened 2021-11-25
Make all option struct fields pointers

### LINKED ISSUE #427: Cannot use UpdateIssue to remove all issue assignees
Since `UpdateIssueOptions` defines `AssigneeIDs` as a `[]int`, and marks it as `omitempty`, it is omitted from the JSON request if an empty array is passed into it. The Gitlab API docs describe the field as follows:

> The ID of the user(s) to assign the issue to. Set to 0 or provide an empty value to unassign all assignees.

Based on the current definition of `UpdateIssueOptions`, it is impossible to unassign all assignees. You cannot remove the `omitempty` item, or every call to `UpdateIssue` will remove all issue assignees. if you want to support this, I think there needs to be a separate method.

svanharmelen:
Fixes #427 #1292


mitar:
If we are doing this breaking change, could we also make sure to get rid of all `[]*Struct` and convert them into `[]Struct`? For example:

```go
type ReleaseAssets struct {
	Links []*ReleaseAssetLink `url:"links" json:"links"`
}
```

There is no point in having `*` there. Does API even survive sending a `null` inside the JSON array?

svanharmelen:
`There is no point in having * there` that depends of how and where your use the items from the slice I guess. But regardless, I guess we can have a look to also update that as over time I also came to the same conclusion.

`Does API even survive sending a null inside the JSON array?` This struct is not mean to be send to the API, but I see it actually is, so that's another issue on itself 😏

mitar:
When do you plan to merge this?

svanharmelen:
Not sure yet... But certainly before the end of 2021, so in the coming few weeks.

mitar:
Any reason why to delay? I am using this and it works well. I think it will be just hard to maintain this branch (it already has conflicts).

PR #1298, opened 2021-11-26
Fix field type of PipelineEvent.[]Builds.Runner.RunnerType

### LINKED ISSUE #1297: Pipeline Webhook Event RunnerType contains wrong field type with 0.52.1
Hi,

I started getting an error with 0.52.1 `json: cannot unmarshal string into Go struct field .builds.runner.runner_type of type bool`. #1291 added the `RunnerType` field which seems to be wrong, see https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html#pipeline-events - this should be a `string`, not a `bool`.

While it's documented, here's also a real example that I'm getting:

```json
"runner": {
  "id": 1,
  "description": "shared-runners-manager-5.gitlab.com",
  "runner_type": "instance_type",
  "active": true,
  "is_shared": true,
  "tags": [
	"docker",
	"gce"
  ]
}
```

I can send a PR (before or after the incoming breaking change in #1294) if there are no objections/reasons for this 🙂

vntw:
As documented in ⬇️

Fixes #1297

### DIFF
--- a/event_parsing_webhook_test.go
+++ b/event_parsing_webhook_test.go
@@ -272,6 +272,10 @@ func TestParsePipelineHook(t *testing.T) {
 	if event.Builds[0].ID != 380 {
 		t.Errorf("Builds[0] ID is %v, want %v", event.Builds[0].ID, 380)
 	}
+
+	if event.Builds[0].Runner.RunnerType != "instance_type" {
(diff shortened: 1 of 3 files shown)


svanharmelen (approved):
Thanks @vntw 👍🏻

PR #1301, opened 2021-12-01
Add support for User `current_sign_in_ip` and `last_sign_in_ip`

sichak:
The available fields when getting/listing a GitLab User vary when the request is performed as a `regular` user vs an `admin`:

`Regular (User)`
* https://docs.gitlab.com/ee/api/users.html#for-user

`Admin`
* https://docs.gitlab.com/ee/api/users.html#for-admin

One such variation is the `current_sign_in_ip` and `last_sign_in_ip` fields.

Adding them to the struct appears to correctly return an IP address when the request is made with an `admin` token and `<nil>` when requested by a regular user token.


users.go
```diff
@@ -85,7 +86,9 @@ type User struct {
 	CanCreateProject               bool               `json:"can_create_project"`
 	ProjectsLimit                  int                `json:"projects_limit"`
 	CurrentSignInAt                *time.Time         `json:"current_sign_in_at"`
+	CurrentSignInIP                *net.IP            `json:"current_sign_in_ip"`
 	LastSignInAt                   *time.Time         `json:"last_sign_in_at"`
+	LastSignInIP                   *net.IP            `json:"last_sign_in_ip"`
```
svanharmelen, line 91: Can you add a test to make sure the marshaling is working correctly?

svanharmelen (changes requested):

users.go
sichak: Sure thing! - just added them now

svanharmelen (approved):
Thanks!

sichak:
Thank you @svanharmelen!

PR #1305, opened 2021-12-02
Added missing FilePath and LinkType to ReleaseAssetLink

mitar:
See: https://docs.gitlab.com/ee/api/releases/#create-a-release

### DIFF
--- a/releases.go
+++ b/releases.go
@@ -116,37 +116,38 @@ func (s *ReleasesService) GetRelease(pid interface{}, tagName string, options ..
 	return r, resp, err
 }
 
-// ReleaseAssets represents release assets in CreateRelease() options
-//
-// GitLab API docs:
-// https://docs.gitlab.com/ce/api/releases/index.html#create-a-release
-type ReleaseAssets struct {
-	Links []*ReleaseAssetLink `url:"links" json:"links"`
-}
-
-// ReleaseAssetLink represents release asset link in CreateRelease() options
-//
-// GitLab API docs:
-// https://docs.gitlab.com/ce/api/releases/index.html#create-a-release
-type ReleaseAssetLink struct {
-	Name     string         `url:"name" json:"name"`
-	URL      string         `url:"url" json:"url"`
-	FilePath *string        `url:"filepath,omitempty" json:"filepath,omitempty"`
-	LinkType *LinkTypeValue `url:"link_type,omitempty" json:"link_type,omitempty"`
-}
-
 // CreateReleaseOptions represents CreateRelease() options.
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/releases/index.html#create-a-release
 type CreateReleaseOptions struct {
-	Name        *string        `url:"name" json:"name"`
-	TagName     *string        `url:"tag_name" json:"tag_name"`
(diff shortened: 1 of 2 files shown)


releases.go
```diff
@@ -129,8 +129,10 @@ type ReleaseAssets struct {
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/releases/index.html#create-a-release
 type ReleaseAssetLink struct {
-	Name string `url:"name" json:"name"`
-	URL  string `url:"url" json:"url"`
+	Name     string         `url:"name" json:"name"`
+	URL      string         `url:"url" json:"url"`
+	FilePath *string        `url:"filepath,omitempty" json:"filepath,omitempty"`
+	LinkType *LinkTypeValue `url:"link_type,omitempty" json:"link_type,omitempty"`
```
svanharmelen, line 135: I guess we need to add a `ReleaseAssetLinkOptions` struct instead and use that when making calls to keep things consistent.

svanharmelen (changes requested):

releases.go
mitar: Sorry, I do not get it. Why a different struct? Or are you saying that we should merge `CreateReleaseLinkOptions`, `UpdateReleaseLinkOptions`, and `ReleaseAssetLink` into one struct?

svanharmelen: All option types have `url` tags and define all fields as pointers. All structs that are used to parse the response don't. So I thing we should add a `ReleaseAssetLinkOptions` struct that is used in other options structs.

mitar: Sorry, I do not see `ReleaseAssetLinkOptions` in existing codebase.

I do not get why exactly some other change would be needed. I mean, maybe there has to be some refactor, if you think. But this PR simply adds missing fields to otherwise working API call, no?

svanharmelen (approved):
I updated the PR a little to show you what I meant. The reason why I wanted to update it like this is not because of your changes, but because your changes made me realize this wasn't following the same logic as is used for the rest of package. I guess it slipped through in another PR some time ago...

mitar:
Thanks for updating it and merging. I didn't completely get, what you were asking. I am not familiar enough with the existing codebase. But I see now.

releases.go
```diff
@@ -116,35 +116,38 @@ func (s *ReleasesService) GetRelease(pid interface{}, tagName string, options ..
...
-// ReleaseAssetLink represents release asset link in CreateRelease() options
+// ReleaseAssetsOptions represents release assets in CreateRelease() options.
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/releases/index.html#create-a-release
-type ReleaseAssetLink struct {
-	Name string `url:"name" json:"name"`
-	URL  string `url:"url" json:"url"`
+type ReleaseAssetsOptions struct {
+	Links []*ReleaseAssetLinkOptions `url:"links,omitempty" json:"links,omitempty"`
```
mitar, line 138: I still do not get why `[]*ReleaseAssetLinkOptions` and not simply `[]ReleaseAssetLinkOptions`. :-)

svanharmelen: For now for consistency. If we change/update that, I want to update it everywhere in the package so using any API has the same consistent behavior.

Issue #1288, opened 2021-11-17
Support GitLab Topics API 


timofurrer:
GitLab 14.5 introduced a new GitLab Topics API: https://gitlab.com/gitlab-org/gitlab/-/blob/master/doc/api/topics.md

It would be great if this library could support this. There are also requests in the [GitLab terraform provider to support it](https://github.com/gitlabhq/terraform-provider-gitlab/issues/748) which depends on this library :)

svanharmelen:
Fixed by #1309

PR #1309, opened 2021-12-09
Add topics API support

eliobischof:
I outscoped the ability to upload an avatar as I couldn't find a reference implementation for uploading files. Maybe that should be added for other resources too and with an extra PR?


eliobischof:
Closes: #1288

svanharmelen (approved):
Tweaked things just a little to clean it up, but LGTM! Thanks!

PR #1313, opened 2021-12-13
Expose PathEscape

mitar:
I found it useful when doing GitLab interaction using this package when there are cases that I have to construct URLs myself (like when package does not yet support something).


svanharmelen (approved):
I'll take this, thanks 👍🏻

PR #1314, opened 2021-12-13
Add none documented response property

Micaso:
Hi! 

First of all thank you for the great package. 
As I used it, I discovered a missing property which is not documented in the official GitLab API docs. 
If you defined a [CODEOWNERS](https://docs.gitlab.com/ee/user/project/code_owners.html) file you get as "rule_type" "code_owner" and also the property "section". I only added the property to the response.

### DIFF
--- a/merge_request_approvals_test.go
+++ b/merge_request_approvals_test.go
@@ -85,6 +85,7 @@ func TestGetApprovalState(t *testing.T) {
 					}
 				],
 				"contains_hidden_groups": false,
+				"section": "test",
 				"approved_by": [
 					{
 						"id": 5,
@@ -157,6 +158,7 @@ func TestGetApprovalState(t *testing.T) {
 						FullPath:             "group1",
 					},
 				},
+				Section: "test",
 				ApprovedBy: []*BasicUser{
 					{
 						ID:        5,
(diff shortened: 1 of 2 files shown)


svanharmelen:
I'm usually not a bit fan of adding undocumented things, so a little bit in doubt if we should add this. Do you have a specific need to the field? And if so, can you point me to something that shows this field does indeed exist? Thanks!

Micaso:
Yes I can fully understand you. 
I created a [test repo](https://gitlab.com/micaso/test-codeowners) on GitLab.

The response of following GET response will show the property:
https://gitlab.com/api/v4/projects/32139950/merge_requests/1/approval_rules 

And yes I do have a need for this field as we want to distinguish between reviewer and assignee. Its a workflow thing for us. Based on the section we could set persons as assignee/reviewer. 

Thank you for your time

merge_request_approvals_test.go
```diff
@@ -188,6 +188,7 @@ func TestGetApprovalRules(t *testing.T) {
 				"id": 1,
 				"name": "security",
 				"rule_type": "regular",
+				"section": "test",
```
svanharmelen, line 191: Please place it in the "correct" location, looking at the output that should be just below `contains_hidden_groups`. If you can do that for all instances that would be great.

svanharmelen (commented):
OK, I'm fine to merge it in that case. Yet please update to ordering to it matches the docs (or in this case the actual output). Thanks.

merge_request_approvals_test.go
Micaso: I restructured it

svanharmelen (approved):
Thanks, LGTM!

PR #1318, opened 2021-12-15
Added UploadAvatar

### LINKED ISSUE #229: Support for avatar option in create/edit project
https://docs.gitlab.com/ce/api/projects.html#edit-project -> Currently the CreateProjectOptions does not contain an entry to send along an `avatar`, even thought it is supported in the API. 
Would be great if we could upload an avatar :smile_cat: 

Maybe we can also directly support the other new params `ci_config_path` and `tag_list`?

mitar:
Fixes #229.

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -17,13 +17,9 @@
 package gitlab
 
 import (
-	"bytes"
 	"fmt"
 	"io"
-	"mime/multipart"
 	"net/http"
-	"os"
-	"path/filepath"
 	"time"
 )
 
@@ -1278,7 +1274,7 @@ func (s *ProjectsService) DeleteProjectForkRelation(pid interface{}, options ...
 	return s.client.Do(req, nil)
 }
 
-// ProjectFile represents an uploaded project file
+// ProjectFile represents an uploaded project file.
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/projects.html#upload-a-file
 type ProjectFile struct {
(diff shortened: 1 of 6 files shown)


mitar:
Any hope of having this merged in?

svanharmelen:
Yes, but I'm kind a mega busy with my day job lately... Will get back on this next week of so.

svanharmelen:
@mitar I thought it was maybe better to refactor it all in one go, so I took a stab at it and added a commit to your PR. I don't have good access to a GitLab instance, so it would be really cool if you could help me verify all 3 related methods (`UploadFile`, `UploadAvatar` and `ImportFromFile`).

If they all work as expected I'll merge this PR together with the other backwards incompatible PR. Thanks!

svanharmelen:
Hmm... I just pushed another commit with another solution for the same logic. This one used a request option function. The advantage could be that this option can also be used when creating or editing a project, more inline with what @Bouwdie was trying to work towards earlier.

I'm not sure which solution I like most, maybe the option function isn't needed and makes things more confusing for a user? What are your thoughts @mitar and @Bouwdie?

request_options.go
```diff
@@ -17,25 +17,73 @@
...
 func WithContext(ctx context.Context) RequestOptionFunc {
-	return func(req *retryablehttp.Request) error {
+	return func(req *retryablehttp.Request, opt interface{}) error {
 		*req = *req.WithContext(ctx)
 		return nil
 	}
 }
 
+// WithFile transforms the request to a multipart request with a file.
+func WithFile(content io.Reader, filename string, uploadType UploadType) RequestOptionFunc {
```
mitar, line 42: I think it would be more general if `uploadType` type would be just a string?

mitar:
It looks good. I like the new approach.

mitar:
I can confirm uploading avatar works.

svanharmelen:
Why are you pusing the old version again?

svanharmelen:
I still have the updated version locally, so I will push that and create a new PR from that one. Still not sure which approach the take. Using the `WithFile` option func or the internal `UploadRequest` method....

svanharmelen:
Here is the branch with my changes and the two different solutions (see the individual commits) https://github.com/xanzy/go-gitlab/tree/support-avatar

svanharmelen:
As for `I think it would be more general if uploadType type would be just a string?`, it's not a free format field IMO. We only want to support the types that the GitLab API expects/supports, so I prefer this over a `string`.

mitar:
Oh, my mistake. I tried to rebase against the latest master. I think I made a mistake. Let me fix that.

mitar:
> As for I think it would be more general if uploadType type would be just a string?, it's not a free format field IMO. We only want to support the types that the GitLab API expects/supports, so I prefer this over a string.

Yes, I realized after that so I removed the comment.

mitar:
I restored the brnach.

projects.go
```diff
@@ -1287,57 +1283,55 @@ type ProjectFile struct {
...
+func (s *ProjectsService) UploadAvatar(pid interface{}, avatar io.Reader, options ...RequestOptionFunc) (*Project, *Response, error) {
+	project, err := parseID(pid)
 	if err != nil {
 		return nil, nil, err
 	}
+	u := fmt.Sprintf("projects/%s", PathEscape(project))
 
-	// Set the buffer as the request body.
-	if err = req.SetBody(b); err != nil {
+	options = append(options, WithFile(avatar, "avatar.png", UploadAvatar))
```
mitar, line 1322: I do not think this works. You assume avatars are just `.png` files here. You should allow `filename` argument.

mitar:
> I can confirm uploading avatar works.

Disregard this. I was testing against my original version. Lol.

mitar:
> Still not sure which approach the take. 

Both look fine to me.

Bouwdie:
I like the approach with the option. For the three methods that are now implemented it makes sense, but for the other functions that include a file as input it makes less sense. I do think that ideally we would like the *io.Reader to be part of the create or edit project options instead of a nullable parameter to the signature.  If we can somehow provide a hint in the opt that it might have to deal with files and if it does switch over to the multipart body rendering. Because in the end that is what we are doing right, body in json we prefer and if there is a file upload involved switch over to the multipart. If we are ok with a little reflection, that checks for available reader instances in the opt struct and if they are not nil, we can which the body render impl. In the end a user checks out the api docs from gitlab and expects a property to be available in the opt stuct I think..

Bouwdie:
Oh btw small note, we also need to update the go-querystring dependency to v1.1.0, pointers to slices were not rendered correctly, for example tags in the project options would not be rendered properly. :100:

mitar:
> Because in the end that is what we are doing right, body in json we prefer and if there is a file upload involved switch over to the multipart.

Which endpoint requires both file upload and additional non-file parameters?

Bouwdie:
> > Because in the end that is what we are doing right, body in json we prefer and if there is a file upload involved switch over to the multipart.
> 
> Which endpoint requires both file upload and additional non-file parameters?

For example create project https://docs.gitlab.com/ee/api/projects.html#create-project - where the avatar is just one of the request options.

mitar:
> I do think that ideally we would like the *io.Reader to be part of the create or edit project options

I think the issue with that is that Reader is not enough, you need also a filename, so that GitLab can figure out the MIME type (they seem to do that based on the file extension in the filename).

svanharmelen:
Listening to both your feedback, I actually don't think the option func is very helpful as it probably makes it less clear how and when it can be used and it doesn't mimic a pattern that the GitLab API currently supports. So I think I prefer to use the "internal" `UploadRequest` method instead.

Let me refactor and update the solution just a bit to see if we can also support the idea of switching based on a value in the `opt` struct.

Bouwdie:
@mitar agreed, so probably there will be a new struct type like uploadfile which encapsulates both the reader and the filename

Bouwdie:
@svanharmelen I think if we have a tule for fileupload with a reference in the opt structs we can have with minimal reflection and a null check an indication whether we should switch to form based body rendering. Thx for both your time guys 😊

svanharmelen:
OK, updated the PR again and think it should be good like this.

@Bouwdie I don't like to use reflection based on a specific type to detect if/when to switch the request format. Instead I prefer to check for that explicitly in the methods where it makes sense.

Please let me know any final thoughts or concerns with this approach as I'm ready to merge this one.

Bouwdie:
@svanharmelen perfectly fine not to use refection, it would only result in different behavior for a couple of methods anyway, so most of it would be useless overhead.

projects.go
```diff
@@ -1287,57 +1336,69 @@ type ProjectFile struct {
...
 	}
+	u := fmt.Sprintf("projects/%s", PathEscape(project))
 
-	// Set the buffer as the request body.
-	if err = req.SetBody(b); err != nil {
+	req, err := s.client.UploadRequest(
+		http.MethodPut,
+		u,
+		avatar,
+		"avatar.png",
```
mitar, line 1386: I think you want to pass `filename` through here. :-)

mitar:
Just a minor comment, otherwise it looks great!

projects.go
svanharmelen: Yes, thanks 👍🏻

svanharmelen: Fixed this one...

svanharmelen:
Thanks @mitar and @Bouwdie for all the help input and feedback. I'm going to merge this right after I add one last tag before merging both backwards incompatible PR's.

Bouwdie:
@svanharmelen Thanks for the commits and the fix. I unfortunately am not used to the new review interface of github and didn't push my remarks. One this I would like to stress is to update the go-querystring dependency, refences to slices will otherwise not be encoded properly aka project.Tags[] for example. Thanks again :100:

mitar:
I updated the dependency in my [gitlab-config](https://gitlab.com/tozd/gitlab/config) project and I can confirm that all recent changes work.

svanharmelen:
Nice 👍🏻

PR #1319, opened 2021-12-15
Fixed fork-relation pid argument

mitar:


svanharmelen:
Not sure what is fixed here? Why change this?

mitar:
One could not pass project path string as pid argument like in all other API functions. So this unified these two functions to be the same as the rest of API functions in this package.

svanharmelen (approved):
I completely read this the wrong way 😅 LGTM! Thanks

mitar:
:-)

Thanks for quick reviews!

PR #1326, opened 2021-12-17
Update the NewRequest docs a little

### LINKED ISSUE #1315: NewRequest does not accept absolute path
Its documentation states:

```
// NewRequest creates an API request. A relative URL path can be provided in
// path, in which case it is resolved relative to the base URL of the Client.
// Relative URL paths should always be specified without a preceding slash. If
// specified, the value pointed to by body is JSON encoded and included as the
// request body.
```

Reading this looks like you can provide relative or absolute URL path. But with absolute URL path it just does the wrong thing. So maybe documentation should be improved to say only relative paths are supported or implementation fixed?

svanharmelen:
Fixes #1315

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -520,11 +520,11 @@ func (c *Client) setBaseURL(urlStr string) error {
 	return nil
 }
 
-// NewRequest creates an API request. A relative URL path can be provided in
-// path, in which case it is resolved relative to the base URL of the Client.
-// Relative URL paths should always be specified without a preceding slash. If
-// specified, the value pointed to by body is JSON encoded and included as the
-// request body.
+// NewRequest creates an API request. An optional relative URL path can be
+// provided in path, in which case it is resolved relative to the base URL
+// of the Client.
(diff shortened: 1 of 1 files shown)


gitlab.go
```diff
@@ -520,11 +520,11 @@ func (c *Client) setBaseURL(urlStr string) error {
 	return nil
 }
 
-// NewRequest creates an API request. A relative URL path can be provided in
-// path, in which case it is resolved relative to the base URL of the Client.
-// Relative URL paths should always be specified without a preceding slash. If
-// specified, the value pointed to by body is JSON encoded and included as the
-// request body.
+// NewRequest creates an API request. An optional relative URL path can be
```
mitar, line 523: Is URL path really optional? What happens if you provide an empty string?

svanharmelen: Then it constructs a request that targets the root of the API (see base URL). Not sure if that is a useful endpoint that returns anything useful, but it's perfectly valid to construct a request like that.

PR #1330, opened 2021-12-28
users: Support GPG key operations

BenoitKnecht:
Add functions to list, get, add and remove user GPG keys, as documented in
https://docs.gitlab.com/ee/api/users.html#list-all-gpg-keys.


svanharmelen (approved):
Thanks @BenoitKnecht! LGTM 👍🏻

PR #1335, opened 2022-01-04
Add None and Any options for AssigneeID for all merge request list fu…

### LINKED ISSUE #751: AssigneeID should not be *int
AssigneeID in https://github.com/xanzy/go-gitlab/blob/master/merge_requests.go#L146 which is used in ListMergeRequests() could be 
- int(user id) 
- OR "None" which will return a list of unassigned merge request 
- OR "Any" which will return a list of assigned merge request.

Reference: https://docs.gitlab.com/ce/api/merge_requests.html#list-merge-requests

Suggestion: We can consider changing this from `*int` to `interface{}` since there is no union typing in golang

Bouwdie:
Add None and Any options for AssigneeID for all merge request list functions.
Fixes #751


svanharmelen (approved):
I made it a little more generic and also added `ReviewerID`. But otherwise looks good 👍🏻

PR #1336, opened 2022-01-05
Add project variable filter environment scope option for Get,Update and Remove functions.

Bouwdie:
The github API documentation says it is not required, so i added it as option.


svanharmelen (approved):
Thanks @Bouwdie 👍🏻

PR #1337, opened 2022-01-07
use put for EditProject

hensur:
as described in https://docs.gitlab.com/ee/api/projects.html#edit-project


svanharmelen (approved):
Nice catch, thanks!

Issue #1340, opened 2022-01-11
Change Configuration of Project Level Merge Request Approvals Not Implemented


dbalencar:
https://docs.gitlab.com/ee/api/merge_request_approvals.html#change-configuration

svanharmelen:
Closing this one as it's tied to PR #1341 which I also just closed. Like with the PR, feel free to reopen if needed.

PR #1341, opened 2022-01-11
Change Configuration

dbalencar:
Project Level Merge Request Approvals
https://github.com/xanzy/go-gitlab/issues/1340
https://docs.gitlab.com/ee/api/merge_request_approvals.html#change-configuration

### DIFF
--- a/merge_request_approvals.go
+++ b/merge_request_approvals.go
@@ -35,31 +35,36 @@ type MergeRequestApprovalsService struct {
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/merge_request_approvals.html#merge-request-level-mr-approvals
+// https://docs.gitlab.com/ee/api/merge_request_approvals.html#change-configuration
 type MergeRequestApprovals struct {
-	ID                             int                          `json:"id"`
-	IID                            int                          `json:"iid"`
-	ProjectID                      int                          `json:"project_id"`
-	Title                          string                       `json:"title"`
-	Description                    string                       `json:"description"`
-	State                          string                       `json:"state"`
-	CreatedAt                      *time.Time                   `json:"created_at"`
-	UpdatedAt                      *time.Time                   `json:"updated_at"`
-	MergeStatus                    string                       `json:"merge_status"`
-	Approved                       bool                         `json:"approved"`
-	ApprovalsBeforeMerge           int                          `json:"approvals_before_merge"`
-	ApprovalsRequired              int                          `json:"approvals_required"`
-	ApprovalsLeft                  int                          `json:"approvals_left"`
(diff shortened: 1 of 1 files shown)


merge_request_approvals.go
```diff
@@ -35,31 +35,36 @@ type MergeRequestApprovalsService struct {
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/merge_request_approvals.html#merge-request-level-mr-approvals
+// https://docs.gitlab.com/ee/api/merge_request_approvals.html#change-configuration
```
svanharmelen, line 38: Please remove this line, don't think we need more then one link for one struct.

merge_request_approvals.go
```diff
@@ -417,3 +422,28 @@ func (s *MergeRequestApprovalsService) DeleteApprovalRule(pid interface{}, merge
 
 	return s.client.Do(req, nil)
 }
+
+// ChangeProjectMergeRequestApprovalConfiguration updates the merge request approval configuration of a project.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/merge_request_approvals.html#change-configuration
+func (s *MergeRequestApprovalsService) ChangeProjectApprovalConfiguration(pid interface{}, opt *MergeRequestApprovals, options ...RequestOptionFunc) (*MergeRequestApprovals, *Response, error) {
```
svanharmelen, line 430: This is not correct. `MergeRequestApprovals` is not an option struct so should only be used as return value. Reading the docs it seems this only allows you to change the number of approvals required and so it should probably get its own option struct.

svanharmelen (changes requested):

merge_request_approvals.go
dbalencar: I agree, I can definitely make the required changes... Would you like to suggest a name for this new option struct, please?

merge_request_approvals.go
dbalencar: No problem, I'll do it when I create the new option struct.

merge_request_approvals.go
svanharmelen: Please check the other code for examples, option struct always use the name of the method with the postfix `Options`. So in this case that should be `ChangeProjectApprovalConfigurationOptions`.

svanharmelen:
As far as I can see this is already implemented here: https://github.com/xanzy/go-gitlab/blob/master/projects.go#L1688-L1725

Is that correct or am I missing something? If not, I suggest to close this PR.

svanharmelen:
Closing due to no response. So for now I'm going to assume this is already fixed and no longer needed. If that is incorrect, feel free to reopen this one.

PR #1346, opened 2022-01-20
add auto_cancel_pending_pipelines to Project

itscaro:
In the documentation there is no example with this field in GET method, but it's present in response of PUT example. I checked on 13.12 & 14.4 & 14.7-pre, and also in gitlab source code :-)


svanharmelen (approved):
👍🏻

PR #1355, opened 2022-02-01
Fix: Label and Group Label name escaping

tioxy:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

Allows escaping special characters while operating *Labels* and *Group Labels*

#### Which issue(s) this PR fixes:

NONE

#### Special notes for your reviewer:

I just changed the test scenarios to cover this special usage. 

I've been working with a pattern that prefixes a normal label with a "scope" but does not operate like a scoped label. This allows multiple labels within the same scope to be defined.

Example of MR labels:
- `domain/infra`
- `domain/backend`
- `environment::production`

This MR should fix and allow labels with slashes to work properly.

### DIFF
--- a/group_labels.go
+++ b/group_labels.go
@@ -81,7 +81,7 @@ func (s *GroupLabelsService) GetGroupLabel(gid interface{}, labelID interface{},
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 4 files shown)


svanharmelen (approved):
Looks good to me... Thanks @tioxy!

Issue #1356, opened 2022-02-01
Add draft to MergeRequest


justinrush:
It looks like https://github.com/xanzy/go-gitlab/pull/1283 added support for the new `Draft` field in request options, but it should also be added as a field on the `MergeRequest` object itself.

PR #1357, opened 2022-02-02
Add Draft field to MergeRequest

justinrush:
closes https://github.com/xanzy/go-gitlab/issues/1356

### DIFF
--- a/merge_requests.go
+++ b/merge_requests.go
@@ -106,6 +106,7 @@ type MergeRequest struct {
 	HasConflicts                bool                   `json:"has_conflicts"`
 	BlockingDiscussionsResolved bool                   `json:"blocking_discussions_resolved"`
 	Overflow                    bool                   `json:"overflow"`
+	Draft                       bool                   `json:"draft"`
 }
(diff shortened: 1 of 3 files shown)


merge_requests.go
```diff
@@ -106,6 +106,7 @@ type MergeRequest struct {
 	HasConflicts                bool                   `json:"has_conflicts"`
 	BlockingDiscussionsResolved bool                   `json:"blocking_discussions_resolved"`
 	Overflow                    bool                   `json:"overflow"`
+	Draft                       bool                   `json:"draft"`
```
svanharmelen, line 109: Please place this between `Description` and `WorkInProgress` so it's inline with the ordering in the docs (as all fields are ordered following the docs).

svanharmelen (changes requested):

merge_requests.go
justinrush: done!

svanharmelen (approved):
LGTM! Thanks @justinrush!

PR #1364, opened 2022-02-08
Fetch ClosedAt field for epics

zwerch:
Hey!

We needed the `ClosedAt` field on the epic for a small internal tool, so I added it.

Thanks for your work, this library is really useful!

### DIFF
--- a/epics.go
+++ b/epics.go
@@ -63,6 +63,7 @@ type Epic struct {
 	DueDateFromMilestones   *ISOTime    `json:"due_date_from_milestones"`
 	CreatedAt               *time.Time  `json:"created_at"`
 	UpdatedAt               *time.Time  `json:"updated_at"`
+	ClosedAt                *time.Time  `json:"closed_at"`
 	Labels                  []string    `json:"labels"`
 	Upvotes                 int         `json:"upvotes"`
 	Downvotes               int         `json:"downvotes"`


marceloboeira (approved):

epics.go
```diff
@@ -63,6 +63,7 @@ type Epic struct {
 	DueDateFromMilestones   *ISOTime    `json:"due_date_from_milestones"`
 	CreatedAt               *time.Time  `json:"created_at"`
 	UpdatedAt               *time.Time  `json:"updated_at"`
+	ClosedAt                *time.Time  `json:"closed_at"`
```
svanharmelen, line 66: Looking at the docs I wonder if these 3 `...At` fields should now use the `*ISOTime` type instead?

zwerch: Hey, thanks for your comment, I actually wondered the same.
But looking at https://github.com/xanzy/go-gitlab/blob/a90abbe96def9e224e0d9e7aa0fde90b71916eb0/types.go#L385 I realised that this type is made for date-only types.
For all `...At` fields the API returns full RFC3339 timestamps.

svanharmelen: Ah, good point! Missed that somehow... @pschwartz I think we need to revert that change in the PR you just changed (on my request) as well. Sorry about that!

zwerch:
thanks a lot for merging :)

epics.go
pschwartz: I will open a PR to roll forward. It won’t cause an issue till that PR is in as ISOTime is type defined as time.Time.

pschwartz: There is a bunch of miss uses through the code base that we should clean up if I remember correctly.

svanharmelen: `ISOTime` is indeed defined as `time.Time` but with a specific format to parse values as dates only. So it doesn't break, but is does result in wrong data.

Not sure if there are more places where this should be updated. At least not as far as I know...

PR #1367, opened 2022-02-08
Add missing list option in ExploreSnippets request

### LINKED ISSUE #1365: ExploreSnippets missing list options in request
`ExploreSnippets` currently does not support pagination (`PerPage` and `Page`):

```go
func main() {
	git, err := gitlab.NewClient("glpat-XXXX")
	if err != nil {
		log.Fatalf("Failed to create client: %v", err)
	}

	snippets, _, err := git.Snippets.ExploreSnippets(&gitlab.ExploreSnippetsOptions{Page: 1, PerPage: 5})
	if err != nil {
		fmt.Println(err)
	}

	fmt.Println(len(snippets))
}
```

```
> 20
```

The `ExploreSnippetsOptions` type already exists and is an argument to the function, however the actual request is passing in `nil` as opposed to the option list:

```go
req, err := s.client.NewRequest("GET", "snippets/public", nil, options)
```

sichak:
Fixes #1365


svanharmelen (approved):
Oeps! Nice catch 👍🏻

PR #1368, opened 2022-02-08
Update SSHKey to include expires_at as defined by API doc

pschwartz:

### DIFF
--- a/users.go
+++ b/users.go
@@ -396,6 +396,7 @@ type SSHKey struct {
 	Title     string     `json:"title"`
 	Key       string     `json:"key"`
 	CreatedAt *time.Time `json:"created_at"`
+	ExpiresAt *time.Time `json:"expires_at"`
 }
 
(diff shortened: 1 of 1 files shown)


pschwartz:
@svanharmelen Another fix if your able to review.

marceloboeira (approved):

users.go
```diff
@@ -396,6 +396,7 @@ type SSHKey struct {
 	Title     string     `json:"title"`
 	Key       string     `json:"key"`
 	CreatedAt *time.Time `json:"created_at"`
+	ExpiresAt *time.Time `json:"expires_at"`
```
svanharmelen, line 399: Looking at the docs, should both of these be of type `*ISOTime`?

pschwartz: Would be effectively the same. Will change to match other response structs.

svanharmelen (approved):
Thanks, LGTM 👍🏻

PR #1371, opened 2022-02-09
Add build timeout field to project response

sschustar:
Though not mentioned in the [GitLab Projects API docs](https://docs.gitlab.com/ee/api/projects.html#get-single-project), the `build_timeout` field is included in the GitLab API response (tested with GitLab Enterprise Edition 14.5.2). Would be great to have it supported here.


svanharmelen (approved):
Thanks @sschustar, LGTM!

PR #1374, opened 2022-02-10
Add RequestOptionFunc to remaining methods

hanzei:
Add RequestOptionFunc to methods that don't accept it yet


svanharmelen (approved):
Thanks 👍🏻

Issue #1372, opened 2022-02-09
Add support for multiple files when creating snippets


zemzale:
The snippets API now has deprecated the currently used keys `file_name` and `content` to use a `files` array.

https://docs.gitlab.com/14.7/ee/api/project_snippets.html#create-new-snippet

I am opening the ticket to just track this change. Gonna try to open PR as soon as possible

svanharmelen:
This one is solved by #1375

PR #1375, opened 2022-02-10
feat(snippets): Add `files` array to snippet creation

zemzale:
This adds the new `files` array for creating snippets that replaces the
now deprecated `filename` and `contents` keys.

This is needed for creating more then one file in snippets.

Issue #1372


snippets.go
```diff
@@ -111,6 +124,7 @@ type CreateSnippetOptions struct {
 	Description *string          `url:"description,omitempty" json:"description,omitempty"`
 	Content     *string          `url:"content,omitempty" json:"content,omitempty"`
 	Visibility  *VisibilityValue `url:"visibility,omitempty" json:"visibility,omitempty"`
+	Files       []*SnippetFile   `url:"files,omitempty" json:"files,omitempty"`
```
svanharmelen, line 127: I know this seems weird, but this needs to be `*[]*SnippetFile` to keep it consistent with the rest of the package.

project_snippets.go
```diff
@@ -94,6 +94,7 @@ type CreateProjectSnippetOptions struct {
 	Description *string          `url:"description,omitempty" json:"description,omitempty"`
 	Content     *string          `url:"content,omitempty" json:"content,omitempty"`
 	Visibility  *VisibilityValue `url:"visibility,omitempty" json:"visibility,omitempty"`
+	Files       []*SnippetFile   `url:"files,omitempty" json:"files,omitempty"`
```
svanharmelen, line 97: I know this seems weird, but this needs to be `*[]*SnippetFile` to keep it consistent with the rest of the package.

svanharmelen (changes requested):

svanharmelen (approved):
Thanks 👍🏻

PR #1378, opened 2022-02-12
add topic query parameter to ListGroupProjectsOptions

### LINKED ISSUE #1351: add topics to ListGroupProjectsOptions
you can query group projects by a topic

```
curl  -H "PRIVATE-TOKEN: <TOKEN>" https://gitlab.com/api/v4/groups/<GROUP_ID>/projects?topic=MyTopic
```

Please add that field to the ListGroupProjectsOptions struct.

currently you can add a query parameter by adding a RequestOptionFunc

```
func requestWithTopic(topic string) gitlab.RequestOptionFunc {
	return func(req *retryablehttp.Request) error {
		if topic != "" {
			queryString := req.Request.URL.Query()
			queryString.Add("topic", topic)
			req.Request.URL.RawQuery = queryString.Encode()
		}
		return nil
	}
}

func listGroupProjects(id int, topic string) {
		_, _, _ := g.Client.Groups.ListGroupProjects(id, &gitlab.ListGroupProjectsOptions{}, requestWithTopic(topic))
}
```

morriswinkler:
closes #1351

### DIFF
--- a/groups.go
+++ b/groups.go
@@ -201,6 +201,7 @@ type ListGroupProjectsOptions struct {
 	Simple                   *bool             `url:"simple,omitempty" json:"simple,omitempty"`
 	Sort                     *string           `url:"sort,omitempty" json:"sort,omitempty"`
 	Starred                  *bool             `url:"starred,omitempty" json:"starred,omitempty"`
+	Topic                    *string           `url:"topic,omitempty" json:"topic,omitempty"`
 	Visibility               *VisibilityValue  `url:"visibility,omitempty" json:"visibility,omitempty"`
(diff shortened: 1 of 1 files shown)


svanharmelen:
Can you point to anything (code maybe) that shows this is valid? As the docs don't mention this option...

morriswinkler:
it is not documented, but try 
```
curl https://gitlab.com/api/v4/groups/11505324/projects?topic=wordpress
```

I think it is in the model code for projects, there is a filter option that accepts topic, so since it works works projects the search endpoint it also works with the projects search below the group section of the API, but I have no idea how that is wired.

The code for the project model is in https://gitlab.com/gitlab-org/gitlab/-/blob/master/app/models/project.rb

svanharmelen (approved):
Thanks @morriswinkler!

PR #1381, opened 2022-02-14
Support `avatar` field in create and update method of the topic service

timofurrer:


timofurrer:
@svanharmelen is something like that implemented already somewhere here in the library?

timofurrer:
@svanharmelen It should be properly implemented now. I've also added an example which "works on my machine ™). Maybe you can verify?

timofurrer:
@svanharmelen would be nice to have this in the next release 🎉

svanharmelen:
@timofurrer I added a commit wit a few changes. I tested what the requests look like in different situations and it all looks good. But probably good if you can confirm/test it against a running instance?

timofurrer:
@svanharmelen at least running the example works as expected 👍

PR #1385, opened 2022-02-20
fix incomplete push, push_tag and repositry_update system hooks

jlamur:
The structs for the system hooks `push`, `tag_push` and
`repository_update` were left empty. This commit adds all the fields
from the documentation here:
https://docs.gitlab.com/ee/system_hooks/system_hooks.html#create-a-system-hook
except for some of them that appear to be redundant (like `repostory`
that looks like it's a subset of `project`).

### DIFF
--- a/event_systemhook_types.go
+++ b/event_systemhook_types.go
@@ -147,8 +147,8 @@ type PushSystemEvent struct {
 		Description       string `json:"description"`
 		WebURL            string `json:"web_url"`
 		AvatarURL         string `json:"avatar_url"`
-		GitHttpURL        string `json:"git_http_url"`
-		GitSshURL         string `json:"git_ssh_url"`
+		GitHTTPURL        string `json:"git_http_url"`
+		GitSSHURL         string `json:"git_ssh_url"`
 		Namespace         string `json:"namespace"`
 		VisibilityLevel   int    `json:"visibility_level"`
 		PathWithNamespace string `json:"path_with_namespace"`
@@ -157,10 +157,10 @@ type PushSystemEvent struct {
 		URL               string `json:"url"`
 	} `json:"project"`
 	Commits []struct {
-		ID        string `json:"id"`
-		Message   string `json:"message"`
+		ID        string    `json:"id"`
+		Message   string    `json:"message"`
 		Timestamp time.Time `json:"timestamp"`
-		URL       string `json:"url"`
+		URL       string    `json:"url"`
 		Author    struct {
 			Name  string `json:"name"`
 			Email string `json:"email"`
@@ -190,8 +190,8 @@ type TagPushSystemEvent struct {
 		Description       string `json:"description"`
 		WebURL            string `json:"web_url"`
 		AvatarURL         string `json:"avatar_url"`
-		GitHttpURL        string `json:"git_http_url"`
-		GitSshURL         string `json:"git_ssh_url"`
+		GitHTTPURL        string `json:"git_http_url"`
+		GitSSHURL         string `json:"git_ssh_url"`
 		Namespace         string `json:"namespace"`
 		VisibilityLevel   int    `json:"visibility_level"`
 		PathWithNamespace string `json:"path_with_namespace"`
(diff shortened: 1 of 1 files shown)


event_systemhook_types.go
```diff
@@ -130,6 +130,41 @@ type UserTeamSystemEvent struct {
 // https://docs.gitlab.com/ee/system_hooks/system_hooks.html
 type PushSystemEvent struct {
 	BaseSystemEvent
+	Before       string `json:"before"`
+	After        string `json:"after"`
+	Ref          string `json:"ref"`
+	CheckoutSha  string `json:"checkout_sha"`
```
svanharmelen, line 136: Please write all the `Sha` parts as `SHA`, thanks!

event_systemhook_types.go
```diff
@@ -130,6 +130,41 @@ type UserTeamSystemEvent struct {
...
+	UserID       int    `json:"user_id"`
+	UserName     string `json:"user_name"`
+	UserUsername string `json:"user_username"`
+	UserEmail    string `json:"user_email"`
+	UserAvatar   string `json:"user_avatar"`
+	ProjectID    int    `json:"project_id"`
+	Project      struct {
```
svanharmelen, line 143: Can we use the `Project` struct instead? Or is the returned data too different?

event_systemhook_types.go
```diff
@@ -130,6 +130,41 @@ type UserTeamSystemEvent struct {
...
+		GitHttpUrl        string `json:"git_http_url"`
+		GitSshUrl         string `json:"git_ssh_url"`
+		Namespace         string `json:"namespace"`
+		VisibilityLevel   int    `json:"visibility_level"`
+		PathWithNamespace string `json:"path_with_namespace"`
+		DefaultBranch     string `json:"default_branch"`
+		Homepage          string `json:"homepage"`
+		Url               string `json:"url"`
+	} `json:"project"`
+	Commits []struct {
```
svanharmelen, line 157: Can we use the `Commit` struct instead? Or is the returned data too different?

event_systemhook_types.go
```diff
@@ -130,6 +130,41 @@ type UserTeamSystemEvent struct {
...
+	UserID       int    `json:"user_id"`
+	UserName     string `json:"user_name"`
+	UserUsername string `json:"user_username"`
+	UserEmail    string `json:"user_email"`
+	UserAvatar   string `json:"user_avatar"`
+	ProjectID    int    `json:"project_id"`
+	Project      struct {
+		Name              string `json:"name"`
+		Description       string `json:"description"`
+		WebUrl            string `json:"web_url"`
```
svanharmelen, line 146: Please write all the `Url` parts as `URL`, thanks!

svanharmelen (changes requested):

event_systemhook_types.go
jlamur: If you are refering to the `Commit` struct from [commits.go:37](https://github.com/xanzy/go-gitlab/blob/3ac36fa074a445904aa841c77fb58973ffdb3648/commits.go#L37), then yes it is too different.

FYI most of the structs in `event_webhook_types.go` are also using anonymous structs as they almost always differ.
Eg. https://github.com/xanzy/go-gitlab/blob/3ac36fa074a445904aa841c77fb58973ffdb3648/event_webhook_types.go#L47-L57
https://github.com/xanzy/go-gitlab/blob/3ac36fa074a445904aa841c77fb58973ffdb3648/event_webhook_types.go#L112-L122
https://github.com/xanzy/go-gitlab/blob/3ac36fa074a445904aa841c77fb58973ffdb3648/event_webhook_types.go#L816-L826

svanharmelen (approved):
Check... Thanks @jlamur!

svanharmelen:
Please fix the linting errors...

svanharmelen (approved):
Fixed your linting errors, so all is good now 👍🏻

jlamur:
Sorry, I did not see that the linting was failing. Thanks for fixing the errors and merging the PR.

PR #1387, opened 2022-02-24
Introduce IssueRelation as return-type for ListIssueRelations(…)

### LINKED ISSUE #1386: Make link_type on issue-relations available
Currently `IssueLinksService.ListIssueRelations(…)` returns an `[]*Issue`. Unfortunately this hides the information about the relation-type (`relates_to`, `blocks`,`is_blocked_by`) that is contained in the server-response in the field `link_type`.

As a workaround I created a new type `IssueRelation` which is basically a copy of `Issue` with the additional fields LinkType, LinkCreatedAt and LinkUpdatedAt:

```go
// IssueRelation gets a relation between two issues.
//
// GitLab API docs:
// https://docs.gitlab.com/ee/api/issue_links.html#list-issue-relations
type IssueRelation struct {
	LinkType      string     `json:"link_type"`
	LinkCreatedAt *time.Time `json:"link_created_at"`
	LinkUpdatedAt *time.Time `json:"link_updated_at"`

	ID                   int                    `json:"id"`
	IID                  int                    `json:"iid"`
	ExternalID           string                 `json:"external_id"`
	State                string                 `json:"state"`
	// …and so forth…
}
```

I am a bit unsure about this however, since it seems to me that the result we get from the server does not contain all fields of `Issue`. It however seems that it provides more data than is listed in the example response in the GitLab-api documentation, e.g. `epic`.

I created a pull request for this here: #1387

dse82:
The old return-type Issue hid the link_type, link_created_at and
link_updated_at available in the API-response.

Fixes #1386

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/xanzy/go-gitlab
+module github.com/dse82/go-gitlab
 
 require (
 	github.com/google/go-querystring v1.1.0
--- a/issue_links.go
+++ b/issue_links.go
@@ -19,6 +19,7 @@ package gitlab
 import (
 	"fmt"
 	"net/http"
+	"time"
 )
 
 // IssueLinksService handles communication with the issue relations related methods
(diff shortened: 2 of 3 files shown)


svanharmelen:
I'm absolutely fine with using a dedicated struct here, but then lets make sure it only contains the fields that are actually returned by the API. Can you maybe find that from the docs or otherwise the GitLab code?

dse82:
I will do that.

svanharmelen:
@dse82 can you either update the struct and revert the `go.mod` change or close this PR? Otherwise it's just sitting here (I cannot merge it with that `go.mod` change).

Thanks!

dse82:
Hi @svanharmelen! Sorry for letting you wait. I didn't get to finalizing the change and won't probably in the next days. I will just close the MR and reopen it as soon as I got to finalizing the change.

svanharmelen:
All good 👍🏻

PR #1390, opened 2022-02-24
Add `PreventForkingOutsideGroup` to `Group` struct

### LINKED ISSUE #1389: Add PreventForkingOutsideGroup to Group struct
`prevent_forking_outside_group` is returned in the GET Group API. `PreventForkingOutsideGroup` already exists in `UpdateGroupOptions` and can also be added to the `Group` struct.

jtymes:
Closes #1389

### DIFF
--- a/groups.go
+++ b/groups.go
@@ -75,6 +75,7 @@ type Group struct {
 	ExtraSharedRunnersMinutesLimit int              `json:"extra_shared_runners_minutes_limit"`
 	MarkedForDeletionOn            *ISOTime         `json:"marked_for_deletion_on"`
 	CreatedAt                      *time.Time       `json:"created_at"`
+	PreventForkingOutsideGroup     bool             `json:"prevent_forking_outside_group"`
 }
 
 // LDAPGroupLink represents a GitLab LDAP group link.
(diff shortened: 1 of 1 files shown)


groups.go
```diff
@@ -75,6 +75,7 @@ type Group struct {
 	ExtraSharedRunnersMinutesLimit int              `json:"extra_shared_runners_minutes_limit"`
 	MarkedForDeletionOn            *ISOTime         `json:"marked_for_deletion_on"`
 	CreatedAt                      *time.Time       `json:"created_at"`
+	PreventForkingOutsideGroup     bool             `json:"prevent_forking_outside_group"`
```
svanharmelen, line 78: Please place this between `ExtraSharedRunnersMinutesLimit` and `MarkedForDeletionOn` (so it follows the order in the GitLab code).

And I think we should also add this field as an option to the `UpdateGroup` method (as it's in the docs)?

svanharmelen (changes requested):

groups.go
jtymes: Moved. 8599d7e 

The field is already in `UpdateGroupOptions` https://github.com/xanzy/go-gitlab/blob/fe1a1132a6b7f8536626048c39dcdc0aa750b16a/groups.go#L368
It might have been a field that the API accepted as a parameter but then wouldn't expose (like `membership_lock`) and now it's visible.

svanharmelen (approved):
Thanks @jtymes 👍🏻

PR #1391, opened 2022-02-25
fix: add ImportError field to ImportStatus

zbindenren:


zbindenren:
Otherwise it is impossible to find out import error.

project_import_export.go
```diff
@@ -46,6 +46,7 @@ type ImportStatus struct {
 	PathWithNamespace string     `json:"path_with_namespace"`
 	CreateAt          *time.Time `json:"create_at"`
 	ImportStatus      string     `json:"import_status"`
+	ImportError       string     `json:"import_error"`
```
svanharmelen, line 49: Should we also add `CorrelationID` and `FailedRelations`? As I notice that are also mentioned in the docs...

svanharmelen (changes requested):

project_import_export.go
zbindenren: Yes that could make sense. I just added the minimum to get the reason when the import fails.

svanharmelen (approved):
Added missing fields, should be good now. Thanks @zbindenren...

PR #1392, opened 2022-02-25
Implement first draft of Acceptance Tests

timofurrer:
I took a stab at implementing some basic form of Acceptance Tests which run against a real GitLab instance.

They are located in the `acc/` directory. They run against a GitLab CE instance, but also a GitLab EE instance can be used given that you own license. I've wrote a brief contributing file to document this.

I'm not really sure if that's the best way yet to structure this, but I think it's already a valuable start.

We are actually using a pretty similar setup in the [GitLab terraform provider](https://github.com/gitlabhq/terraform-provider-gitlab).

@svanharmelen please let me know what you think ;)

### DIFF
--- a/.github/workflows/acceptance-tests-ce.yml
+++ b/.github/workflows/acceptance-tests-ce.yml
@@ -1,0 +1,50 @@
+# This workflow runs acceptance tests on pull requests (CE)
+
+name: acceptance-tests-ce
+
+on:
+  push: {}
+  pull_request:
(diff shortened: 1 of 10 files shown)


timofurrer:
@svanharmelen do you have some time to sync on this next week?

timofurrer:
@timofurrer we've migrated the repository over to [GitLab.com](https://gitlab.com/gitlab-org/api/client-go).

PR #1394, opened 2022-02-27
Add file_template_project_id attribute to Group Struct

### LINKED ISSUE #1393: Add support to the Group struct for file_template_project_id
Currently, UpdateGroup allows for passing in the `file_template_project_id` attribute, but the Group struct doesn't include the attribute, so the GetGroup func doesn't return the value. 

I'll go ahead and submit a PR to add the json attribute to the Group struct so that the value is returned appropriately from the GetGroup func.

PatrickRice-KSC:
This PR adds support for the `file_template_project_id` when reading a Group via the API by adding the value to the `Group` struct. In addition, it adds one test to ensure that when the value is present properly in the JSON, it unmarshals properly.

Closes #1393

### DIFF
--- a/groups.go
+++ b/groups.go
@@ -71,6 +71,7 @@ type Group struct {
 	LDAPCN                         string           `json:"ldap_cn"`
 	LDAPAccess                     AccessLevelValue `json:"ldap_access"`
 	LDAPGroupLinks                 []*LDAPGroupLink `json:"ldap_group_links"`
+	FileTemplateProjectId          int              `json:"file_template_project_id"`
 	SharedRunnersMinutesLimit      int              `json:"shared_runners_minutes_limit"`
 	ExtraSharedRunnersMinutesLimit int              `json:"extra_shared_runners_minutes_limit"`
 	PreventForkingOutsideGroup     bool             `json:"prevent_forking_outside_group"`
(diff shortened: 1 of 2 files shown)


timofurrer (approved):

groups.go
```diff
@@ -71,6 +71,7 @@ type Group struct {
 	LDAPCN                         string           `json:"ldap_cn"`
 	LDAPAccess                     AccessLevelValue `json:"ldap_access"`
 	LDAPGroupLinks                 []*LDAPGroupLink `json:"ldap_group_links"`
+	FileTemplateProjectId          int              `json:"file_template_project_id"`
```
svanharmelen, line 74: Please place this between `FullPath` and `ParentID` following the ordering in the docs. And for the naming:
```suggestion
	FileTemplateProjectID          int              `json:"file_template_project_id"`
```

svanharmelen (changes requested):

groups.go
PatrickRice-KSC: Thanks @svanharmelen ! I fixed this issue along with updating the test as well, feel free to re-review 😄

svanharmelen:
Please fix the lint errors (probably need to run `go fmt`)...

PatrickRice-KSC:
Sorry about that! Always a bit tricky with delayed CI :) 

Should be fixed now.

PR #1397, opened 2022-03-02
feat(commit status): add coverage field to CommitStatus

timonwong:


svanharmelen (approved):
Thanks @timonwong. I like to keep the field order inline with the ordering used in the docs, but I noticed this struct didn't have that already so I'm merge this PR and reorder it myself in another commit.

PR #1398, opened 2022-03-02
fix(commit status): coverage can be null

timonwong:
Fix coverage field introduced in #1397.

An example output for null coverage:

```json
{
  "id": 727832,
  "sha": "b0c5d0c9cb396db02fc8a0e249e4c5d113df851a",
  "ref": "ASM-6543",
  "status": "success",
  "name": "test",
  "target_url": null,
  "description": null,
  "created_at": "2022-03-02T13:23:04.500Z",
  "started_at": null,
  "finished_at": "2022-03-02T13:23:21.005Z",
  "allow_failure": false,
  "coverage": null,
  "author": {
    "id": 86,
    "name": "bot",
    "username": "bot",
    "state": "active",
    "avatar_url": "https://gitlab.example.com/uploads/-/system/user/avatar/86/avatar.png",
    "web_url": "https://gitlab.example.com/alaudabot"
  }
}
```

### DIFF
--- a/commits.go
+++ b/commits.go
@@ -399,7 +399,7 @@ type CommitStatus struct {
 	FinishedAt   *time.Time `json:"finished_at"`
 	Name         string     `json:"name"`
 	AllowFailure bool       `json:"allow_failure"`
-	Coverage     float64    `json:"coverage"`
+	Coverage     *float64   `json:"coverage"`
 	Author       Author     `json:"author"`
 	Description  string     `json:"description"`
 	TargetURL    string     `json:"target_url"`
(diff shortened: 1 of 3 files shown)


commits.go
```diff
@@ -399,7 +399,7 @@ type CommitStatus struct {
 	FinishedAt   *time.Time `json:"finished_at"`
 	Name         string     `json:"name"`
 	AllowFailure bool       `json:"allow_failure"`
-	Coverage     float64    `json:"coverage"`
+	Coverage     *float64   `json:"coverage"`
```
svanharmelen, line 402: Sorry but this is not a route I'm willing to go... A lot of these fields can be `null`, but I don't think this API will be usable if we have to `nil` check each and every field we want to use...

PR #1399, opened 2022-03-02
Creation of match between tags.go and Gitlab API

Moon1706:
At present time tags structure doesn't match Gitlab API. For this reason, I add fields `target` and `protected` ([link to Gitlab API](https://docs.gitlab.com/ee/api/tags.html)).

### DIFF
--- a/tags.go
+++ b/tags.go
@@ -17,9 +17,9 @@
 package gitlab
 
 import (
-	"fmt"
-	"net/http"
-	"net/url"
+    "fmt"
+    "net/http"
+    "net/url"
 )
 
 // TagsService handles communication with the tags related methods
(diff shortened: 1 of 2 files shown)


tags.go
```diff
@@ -34,10 +34,12 @@ type TagsService struct {
...
 // GitLab API docs: https://docs.gitlab.com/ce/api/tags.html
 type Tag struct {
-	Commit  *Commit      `json:"commit"`
-	Release *ReleaseNote `json:"release"`
-	Name    string       `json:"name"`
-	Message string       `json:"message"`
+	Commit    *Commit      `json:"commit"`
+	Release   *ReleaseNote `json:"release"`
+	Name      string       `json:"name"`
+	Message   string       `json:"message"`
```
timofurrer, line 40: Not sure how it's handled in general here, but the `message` is actually optional and maybe `null` (according to docs examples).

timofurrer (changes requested):
I think it makes sense to adapt the tests, too :)

tags.go
Moon1706: Firstly, this case doesn't have tests with this fields. Secondly, I add only protected and target fields. I can realize additional featches but in this moment it's overhead)

timofurrer: The tests comment was a general "review" comment, not for this particular line or `message` field.

Tests are here which you'd be able to extend: https://github.com/xanzy/go-gitlab/blob/b5c61e10d5a83e626c3b33823880b0274b393e2d/tags_test.go#L42

svanharmelen (approved):
I tweaked the test a bit, looks good otherwise! Thanks @Moon1706!

Moon1706:
@svanharmelen could you create the tag to this MR? It's needed for [my feature](https://github.com/gitlabhq/terraform-provider-gitlab/pull/910).

svanharmelen:
Tagged: `v0.57.0`

PR #1400, opened 2022-03-02
Implement user `ban` and `unban` APIs

timofurrer:
See https://docs.gitlab.com/ee/api/users.html#ban-user


svanharmelen (approved):
Thanks 👍🏻

PR #1404, opened 2022-03-07
Tweak the ISOTime marshaller

### LINKED ISSUE #1384: Allow to remove `due_date` from Issue
The `.Issues.UpdateIssue()` endpoint doesn't seem to support removing the `due_date`.

This is because the library stored to `due_date` as `*gitlab.ISOTime`. In case it's `nil` it's not passed as payload and therefore the field is left unmodified. However, the API has more states for that field:

* omitted: the field is unchanged
* null / empty string: the due date is removed
* valid date: the due date is added

svanharmelen:
@timofurrer does this fix #1384

### DIFF
--- a/types.go
+++ b/types.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"reflect"
 	"strconv"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


timofurrer:
Just looking at it I would say "yes" 🎉  ... BUT, can't really be 100% sure, maybe some unit test would help here .. If I find some time, I'll look into it ..

PR #1407, opened 2022-03-09
Make the label priority nullable

timofurrer:
The label priority field can actually be `null`.

See examples here: https://docs.gitlab.com/ee/api/labels.html#list-labels

Btw. in the update method we have the same problem as here:

* https://github.com/xanzy/go-gitlab/issues/1384

... `null` is actually a valid value meaning that the priority should be removed from the label. For the `due_date` the fix in the marshaller (https://github.com/xanzy/go-gitlab/pull/1404) may work fine, because the default value isn't a sane due date. However, the integer value `0` for a label priority is a sane value.

I don't see any other option than introduce a special type to support this use case, ...

At the moment I'm trying to implement custom `RequestOptionFunc` to re-write the body which is kinda hacky ...

Actually quite some other API fields suffer from this issue, too.

### DIFF
--- a/labels.go
+++ b/labels.go
@@ -43,7 +43,7 @@ type Label struct {
 	ClosedIssuesCount      int    `json:"closed_issues_count"`
 	OpenMergeRequestsCount int    `json:"open_merge_requests_count"`
 	Subscribed             bool   `json:"subscribed"`
-	Priority               int    `json:"priority"`
+	Priority               *int   `json:"priority"`
 	IsProjectLabel         bool   `json:"is_project_label"`
 }
 
--- a/labels_test.go
+++ b/labels_test.go
@@ -44,7 +44,7 @@ func TestCreateLabel(t *testing.T) {
 	if err != nil {
 		log.Fatal(err)
 	}
-	want := &Label{ID: 1, Name: "My Label", Color: "#11FF22", Priority: 2}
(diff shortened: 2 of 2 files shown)


svanharmelen:
@timofurrer I am not going to introduce pointers in structs used to marshal responses... I mentioned that before already so please stop making PRs for that.

The problem where a `null` value actually has a specific purpose when making an API call (so in the option structs) is a problem I recognize and acknowledge, yet I do not know how to solve it in a generic and consistent manner.

Go makes our lives hard for this specific use case. Open to talk about it using any kind of video conf tool (Discord, Teams, Zoom) if you want.

timofurrer:
> @timofurrer I am not going to introduce pointers in structs used to marshal responses... I mentioned that before already so please stop making PRs for that.

Wasn't aware that you tried to introduce one before 🤔 sorry for that ✌️

svanharmelen:
No worries... I was actually referring to this one https://github.com/xanzy/go-gitlab/pull/1398 But we'll have a chat tomorrow 👍🏻

timofurrer:
Author's name starts with `timo` .. thought it's not me 😅 

> But we'll have a chat tomorrow 👍🏻

👍

svanharmelen:
Whoops!! Sorry about that one then... My bad!

PR #1409, opened 2022-03-11
Setting AssigneeID type on listing options to AssigneeIDValue

### LINKED ISSUE #1408: AssigneeID should also accept "none" and "any"
Hey, I'd like to list unassigned issues, but to do so I'd have to pass the AssigneeID with the value "none", wich is not currently possible, given the option type definition below.

https://github.com/xanzy/go-gitlab/blob/bd30483e865c054c52179448051bef0c597a2495/issues.go#L218

So I was thinking if we could change that in a way to allow for this "any" and "none" possible values.

I'm likely to create a pull request for that sson, but as such strcuture would be quite werid (blame gitlab for that), I'd realy like to know if you guys have tackled this problem and maybe just decided to live it be.

ygarasab:
This allows users to use the 'none' and 'any' AssigneeIDs, as allowed by the GitLab API docs. Naturally, this is likely to generate error on existing dependent code that passes AssigneeID as an integer.

All local tests were successful.

Closes #1408

### DIFF
--- a/issues.go
+++ b/issues.go
@@ -206,31 +206,31 @@ type LabelDetails struct {
 // GitLab API docs: https://docs.gitlab.com/ce/api/issues.html#list-issues
 type ListIssuesOptions struct {
 	ListOptions
-	State              *string			`url:"state,omitempty" json:"state,omitempty"`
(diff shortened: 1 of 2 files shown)


svanharmelen (approved):
Thanks @ygarasab, LGTM!

PR #1410, opened 2022-03-12
Implement API to get single project-level approval rule

timofurrer:


timofurrer:
@svanharmelen by any chance possible to get this merged and have a release with the lately merged MRs? Would be awesome 🎉

projects.go
```diff
@@ -1749,6 +1749,31 @@ func (s *ProjectsService) GetProjectApprovalRules(pid interface{}, options ...Re
...
+		return nil, nil, err
+	}
+	u := fmt.Sprintf("projects/%s/approval_rules/%d", PathEscape(project), ruleID)
+
+	req, err := s.client.NewRequest(http.MethodGet, u, nil, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var par *ProjectApprovalRule
```
svanharmelen, line 1768: The convention is to write this as `par := new(ProjectApprovalRule)` expect when it's a slice... Why? Because that somehow became the way it was written in most methods, so decided to try and keep it consistent.

svanharmelen:
@timofurrer if you can tweak that one little thingy I'm good merging this one...

projects.go
timofurrer: great, thanks and done :)

PR #1411, opened 2022-03-13
Implement GET API for single Project and Group Deploy Token

timofurrer:
I've recently implemented this API in GitLab and it landed in 14.9 and
already available on gitlab.com


svanharmelen (approved):
Thanks again!

PR #1412, opened 2022-03-13
Implement GET API for SSH Key of given user

timofurrer:
I've recently implemented this API in GitLab and will be available in
GitLab 14.9 and is already deployed on gitlab.com


svanharmelen (approved):
Thanks @timofurrer!

PR #1415, opened 2022-03-18
WIP: complete application settings API

timofurrer:
@JanKoppe I've already started here for :

* https://github.com/gitlabhq/terraform-provider-gitlab/issues/957

... maybe you can continue?

PR #1425, opened 2022-03-22
complete application settings API

JanKoppe:
This is taking over the work that @timofurrer has done so far in 
* #1415

I'm currently finishing up the `UpdateSettingsOptions` as well on this in preparation for 
* https://github.com/gitlabhq/terraform-provider-gitlab/issues/957

### DIFF
--- a/settings.go
+++ b/settings.go
@@ -42,325 +42,323 @@ type SettingsService struct {
 // https://gitlab.com/gitlab-org/gitlab/-/blob/v14.9.3-ee/ee/lib/ee/api/helpers/settings_helpers.rb#L10
 // https://gitlab.com/gitlab-org/gitlab/-/blob/v14.9.3-ee/ee/app/helpers/ee/application_settings_helper.rb#L20
 type Settings struct {
-	ID                                        int               `json:"id"`
-	CreatedAt                                 *time.Time        `json:"created_at"`
-	UpdatedAt                                 *time.Time        `json:"updated_at"`
-	AdminMode                                 bool              `json:"admin_mode"`
+	ID                                                    int               `json:"id"`
+	CreatedAt                                             *time.Time        `json:"created_at"`
+	UpdatedAt                                             *time.Time        `json:"updated_at"`
+	AbuseNotificationEmail                                string            `json:"abuse_notification_email"`
+	AdminMode                                             bool              `json:"admin_mode"`
 	AdminNotificationEmail                                string            `json:"admin_notification_email"` // deprecated
-	AbuseNotificationEmail                            string            `json:"abuse_notification_email"`
-	AfterSignUpText                           string            `json:"after_sign_up_text"`
 	AfterSignOutPath                                      string            `json:"after_sign_out_path"`
-	AkismetEnabled                            bool              `json:"akismet_enabled"`
+	AfterSignUpText                                       string            `json:"after_sign_up_text"`
 	AkismetAPIKey                                         string            `json:"akismet_api_key"`
-	AllowLocalRequestsFromWebHooksAndServices bool              `json:"allow_local_requests_from_web_hooks_and_services"`
+	AkismetEnabled                                        bool              `json:"akismet_enabled"`
+	AllowGroupOwnersToManageLDAP                          bool              `json:"allow_group_owners_to_manage_ldap"`
 	AllowLocalRequestsFromHooksAndServices                bool              `json:"allow_local_requests_from_hooks_and_services"` // deprecated
 	AllowLocalRequestsFromSystemHooks                     bool              `json:"allow_local_requests_from_system_hooks"`
-	ArchiveBuildsInHumanReadable              string            `json:"archive_builds_in_human_readable"`
-	AssetProxyEnabled                         bool              `json:"asset_proxy_enabled"`
-	AssetProxyURL                             string            `json:"asset_proxy_url"`
+	AllowLocalRequestsFromWebHooksAndServices             bool              `json:"allow_local_requests_from_web_hooks_and_services"`
+	ArchiveBuildsInHumanReadable                          string            `json:"archive_builds_in_human_readable"`
+	AssetProxyAllowlist                                   []string          `json:"asset_proxy_allowlist"`
+	AssetProxyEnabled                                     bool              `json:"asset_proxy_enabled"`
+	AssetProxyURL                                         string            `json:"asset_proxy_url"`
 	AssetProxyWhitelist                                   []string          `json:"asset_proxy_whitelist"` // deprecated
-	AssetProxyAllowlist                               []string          `json:"asset_proxy_allowlist"`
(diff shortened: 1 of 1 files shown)


JanKoppe:
In lieu of a proper test case, I've run this code against a local GitLab instance spun up by the tooling from the terraform-gitlab-provider:

```go
package main

import (
	"log"

	"github.com/davecgh/go-spew/spew"
	"github.com/jinzhu/copier"
	"github.com/xanzy/go-gitlab"
)

func main() {
	git, err := gitlab.NewClient("ACCTEST1234567890123", gitlab.WithBaseURL("http://localhost:3456"))
	if err != nil {
		log.Fatalf("fail! %v", err)
	}

	// Retrieve Current Settings
	settings, _, err := git.Settings.GetSettings()
	if err != nil {
		log.Fatal(err)
	}

	spew.Dump(settings)

	new_settings := &gitlab.UpdateSettingsOptions{}

	// use copier to copy retrieved values over
	copier.Copy(&new_settings, &settings)

	// fix these issues by niling the fields:
	// {error: enabled_git_access_protocol does not have a valid value, password_authentication_enabled_for_web, signin_enabled are mutually exclusive}
	new_settings.EnabledGitAccessProtocol = nil
	new_settings.PasswordAuthenticationEnabledForWeb = nil
	// {message: {external_pipeline_validation_service_timeout: [must be greater than 0]}}
	new_settings.ExternalPipelineValidationServiceTimeout = nil

	_, _, err = git.Settings.UpdateSettings(new_settings)
	if err != nil {
		log.Fatal(err)
	}
}
```

Nothing exploded, and the messages being passed around as intercepted with mitmproxy look good.

I've noticed that the GitLab documentation for this API seems to be incomplete, and has some mistakes (e.g. wrongly stating that `signin_enabled` is a string type, when it's a boolean in reality).

I'll do some manual comparisons with real responses from the GitLab API to find any remaining inconsistencies.

settings.go
```diff
@@ -33,172 +33,239 @@ type SettingsService struct {
...
-	UniqueIPsLimitPerUser                     int               `json:"unique_ips_limit_per_user"`
-	UniqueIPsLimitTimeWindow                  int               `json:"unique_ips_limit_time_window"`
-	UsagePingEnabled                          bool              `json:"usage_ping_enabled"`
-	UserDefaultExternal                       bool              `json:"user_default_external"`
-	UserDefaultInternalRegex                  string            `json:"user_default_internal_regex"`
-	UserOauthApplications                     bool              `json:"user_oauth_applications"`
-	UserShowAddSSHKeyMessage                  bool              `json:"user_show_add_ssh_key_message"`
-	VersionCheckEnabled                       bool              `json:"version_check_enabled"`
-	WebIDEClientsidePreviewEnabled            bool              `json:"web_ide_clientside_preview_enabled"`
+	ID                                                int               `json:"id"`
```
timofurrer, line 36: I started to just take over the fields in the order they are in the GitLab doc, however I think that @svanharmelen usually wants them in alphabetical order (which I generally would prefer, too) ... so maybe we need to change that here :)

JanKoppe: That should be an easy fix. I need to make another pass at it today as I noticed a lot of missing fields when looking at a real reply from gitlab. Currently thinking that I might need to dig into the ruby code to get real answers 😅

timofurrer: To save you the Ruby code search:

* [parameters for setting the settings](https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/api/settings.rb#L33)
* [Entity returned when getting the settings](https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/api/entities/application_setting.rb)
* [The visible attributes references in the code above](https://gitlab.com/gitlab-org/gitlab/-/blob/master/app/helpers/application_settings_helper.rb#L192)

svanharmelen: Yeah... Keeping the order consistent is an ongoing battle... Not too long ago in the docs a couple of locations started to use alphabetical ordering, but in the docs it's not really consistent.

Until now I generally try to follow to order used in the docs so that it's easy to lookup something (should be in the same position). But thinking about making a PR that just makes all structs (so both option and response structs) alphabetical to just have a single consistent approach in this package 🤷🏻 

So for now this order looks fine 👍🏻

PatrickRice-KSC: I know I'm not on the thread yet, but I'd personally suggest we don't use un-documented fields for exactly the reason you suggest. GitLab even has some APIs that are explicitly documented as internal, so they're not even internal and stable, they're just not documented 😕 

I'd say if we want to use them, we create an issue for GitLab and look to document them, which would allow their PM/engineers to chime in on whether they're stable.

svanharmelen: I think I agree with that...

svanharmelen:
Let me know when this is ready for a review...

Shocktrooper:
@svanharmelen @PatrickRice-KSC @timofurrer Is there an easy way to keep up with new/changed fields that GitLab adds from release to release. I am mentioning this because I do a custom diff outside of terraform to account for these new/changed fields that Gitlab doesn't necessarily mention in the release notes from month to month. Potentially when we implement acceptance testing against a real GitLab instance for go-gitlab we can error out on unknown/new and unaccounted for fields in the API response? Failure to keep up from month to month might end up rendering the downstream provider resource nigh unusable if GitLab does not maintain backwards compatibility with the API attributes.  We also have to think about documenting what version was tested against for each release of the resource.

JanKoppe:
> Let me know when this is ready for a review...

This effectively depends on what @Shocktrooper mentioned in their comment - There needs to be a decision on which attributes to actually expose, and with that source of truth to work with. Seems like GitLab changes the behaviour of this specific API a lot, including breaking changes, so there needs to be an agreement on how to handle that. One option could of course be "Latest state of documentation is the source of truth, period." - this would result in some maintenance work every month.

JanKoppe:
Updated with all the attributes that I could extract from the 14.9.3 GitLab source. Sorted for easier maintainability. My hands hurt. :)

svanharmelen (approved):
Added a commit with some sorting "fixes". Additionally we generally only want to support documented features/fields, but I open to make an exception for this one...

The title still reads WIP, is it still or can this one be merged?

JanKoppe:
Thanks for the touchup, @svanharmelen. I forgot to remove the WIP prefix, removed now. Can be merged IMO.

svanharmelen:
@Shocktrooper I am looking into options to give more love and attention to this repository as I notice I don't have the time and/or energy to really think about these kinds (e.g. better CI, proper automated testing, release management) of improvements.

Probably the fact that I don't actually use this package myself anymore (for quite a few years now), doesn't help in finding (i.e. making) time. Will keep you (all of you of course) posted with regards to any changes in ownership or governance of this repo, but let's not use this PR for that discussion 😏 

So I'll go ahead and merge this one and when there is more clarity I will update the README or create an issue to communicate next steps. Cheers!

PR #1426, opened 2022-03-22
Add get wiki page options

jtymes:
GitLab 14.9 introduced new options for retrieving Wiki pages
- `render_html` - Return `content` as HTML
- `version` - Retrieve a wiki page at a given commit SHA

The response now also includes `encoding`.

### DIFF
--- a/group_wikis_test.go
+++ b/group_wikis_test.go
@@ -14,7 +14,14 @@ func TestListGroupWikis(t *testing.T) {
 	mux.HandleFunc("/api/v4/groups/1/wikis",
 		func(w http.ResponseWriter, r *http.Request) {
 			testMethod(t, r, http.MethodGet)
-			fmt.Fprint(w, `[{"content":"content here","format":"markdown","slug":"deploy","title":"deploy title"}]`)
+			fmt.Fprint(w, `[
+				{
+					"content": "content here",
+					"format": "markdown",
+					"slug": "deploy",
+					"title": "deploy title"
+				}
+			]`)
 		})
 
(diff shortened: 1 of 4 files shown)


group_wikis_test.go
```diff
@@ -35,15 +35,15 @@ func TestGetGroupWikiPage(t *testing.T) {
...
 		})
 
-	groupwiki, _, err := client.GroupWikis.GetGroupWikiPage(1, "deploy")
+	groupwiki, _, err := client.GroupWikis.GetGroupWikiPage(1, "deploy", &GetGroupWikiPageOptions{})
 	if err != nil {
 		t.Errorf("GroupWikis.GetGroupWikiPage returned error: %v", err)
 	}
 
-	want := &GroupWiki{Content: "content here", Format: WikiFormatASCIIDoc, Slug: "deploy", Title: "deploy title"}
+	want := &GroupWiki{Content: "content here", Encoding: "UTF-8", Format: WikiFormatASCIIDoc, Slug: "deploy", Title: "deploy title"}
```
svanharmelen, line 46: Maybe nice to format these test bodies similar to how they are formatted in `wikis_test.go`?

svanharmelen (changes requested):
Just a small style update request, looks good otherwise!

group_wikis_test.go
jtymes: I went through and reformatted the `want`s and also reformatted the JSON strings in both files a little bit. Let me know if that was what you were looking for!

svanharmelen (approved):
Thanks for the extra work! Much appreciated! 👍🏻

PR #1427, opened 2022-03-23
Align group hooks with GitLab API

jtymes:
`push_events_branch_filter` is not yet documented, but it is there. https://gitlab.com/gitlab-org/gitlab/-/merge_requests/82531

`subgroup_events` has been around since 2021 https://gitlab.com/gitlab-org/gitlab/-/merge_requests/52658

### DIFF
--- a/group_hooks.go
+++ b/group_hooks.go
@@ -42,7 +42,7 @@ type GroupHook struct {
 	WikiPageEvents           bool       `json:"wiki_page_events"`
 	DeploymentEvents         bool       `json:"deployment_events"`
 	ReleasesEvents           bool       `json:"releases_events"`
-	SubgroupEvents           bool       `json:"subgroup_events"`
+	SubGroupEvents           bool       `json:"subgroup_events"`
 	EnableSSLVerification    bool       `json:"enable_ssl_verification"`
 	CreatedAt                *time.Time `json:"created_at"`
 }
@@ -113,7 +113,7 @@ type AddGroupHookOptions struct {
 	WikiPageEvents           *bool   `url:"wiki_page_events,omitempty"  json:"wiki_page_events,omitempty"`
 	DeploymentEvents         *bool   `url:"deployment_events,omitempty" json:"deployment_events,omitempty"`
 	ReleasesEvents           *bool   `url:"releases_events,omitempty" json:"releases_events,omitempty"`
-	SubgroupEvents           *bool   `url:"subgroup_events,omitempty" json:"subgroup_events,omitempty"`
+	SubGroupEvents           *bool   `url:"subgroup_events,omitempty" json:"subgroup_events,omitempty"`
 	EnableSSLVerification    *bool   `url:"enable_ssl_verification,omitempty"  json:"enable_ssl_verification,omitempty"`
 	Token                    *string `url:"token,omitempty" json:"token,omitempty"`
 }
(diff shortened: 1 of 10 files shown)


group_hooks.go
```diff
@@ -41,6 +42,7 @@ type GroupHook struct {
 	WikiPageEvents           bool       `json:"wiki_page_events"`
 	DeploymentEvents         bool       `json:"deployment_events"`
 	ReleasesEvents           bool       `json:"releases_events"`
+	SubGroupEvents           bool       `json:"subgroup_events"`
```
timofurrer, line 45:
```suggestion
	SubgroupEvents           bool       `json:"subgroup_events"`
```

Looking at the pattern of other names I suggest to *camel case* when *snake cased* :D

group_hooks.go
```diff
@@ -110,6 +113,7 @@ type AddGroupHookOptions struct {
 	WikiPageEvents           *bool   `url:"wiki_page_events,omitempty"  json:"wiki_page_events,omitempty"`
 	DeploymentEvents         *bool   `url:"deployment_events,omitempty" json:"deployment_events,omitempty"`
 	ReleasesEvents           *bool   `url:"releases_events,omitempty" json:"releases_events,omitempty"`
+	SubGroupEvents           *bool   `url:"subgroup_events,omitempty" json:"subgroup_events,omitempty"`
```
timofurrer, line 116:
```suggestion
	SubgroupEvents           *bool   `url:"subgroup_events,omitempty" json:"subgroup_events,omitempty"`
```

group_hooks.go
```diff
@@ -156,6 +161,7 @@ type EditGroupHookOptions struct {
 	WikiPageEvents           *bool   `url:"wiki_page_events,omitempty" json:"wiki_page_events,omitempty"`
 	DeploymentEvents         *bool   `url:"deployment_events,omitempty" json:"deployment_events,omitempty"`
 	ReleasesEvents           *bool   `url:"releases_events,omitempty" json:"releases_events,omitempty"`
+	SubGroupEvents           *bool   `url:"subgroup_events,omitempty" json:"subgroup_events,omitempty"`
```
timofurrer, line 164:
```suggestion
	SubgroupEvents           *bool   `url:"subgroup_events,omitempty" json:"subgroup_events,omitempty"`
```

group_hooks_test.go
```diff
@@ -131,6 +138,7 @@ func TestGetGroupHook(t *testing.T) {
 		WikiPageEvents:           true,
 		DeploymentEvents:         true,
 		ReleasesEvents:           true,
+		SubGroupEvents:           true,
```
timofurrer, line 141:
```suggestion
		SubgroupEvents:           true,
```

group_hooks_test.go
```diff
@@ -194,6 +205,7 @@ func TestAddGroupHook(t *testing.T) {
...
```
timofurrer, line 208:
```suggestion
		SubgroupEvents:           true,
```

group_hooks_test.go
```diff
@@ -257,6 +272,7 @@ func TestEditGroupHook(t *testing.T) {
...
```
timofurrer, line 275:
```suggestion
		SubgroupEvents:           true,
```

group_hooks.go
timofurrer: ... and don't get me wrong - I have the same urge to camel case to `SubGroup` ... but would favor consistency in this case 💃

timofurrer (commented):
Thanks for going the extra mile 🎉

group_hooks.go
svanharmelen: Agree with both these statements 😏 So lets keep _camel casing_ whatever the _snake cased_ names are. Even through in this case (no pun intended) I would also tent to write `SubGroup...`

jtymes: I was originally going to use `SubgroupEvents`, but this struct https://github.com/xanzy/go-gitlab/blob/1a0a4f28753dbb04978845336d90d402ab4b36b5/event_webhook_types.go#L894 was `SubGroupEvent`, so I was trying to be consistent with that casing. Should I change both places, or just here?

svanharmelen: Nice find! Please update both places then :) Thanks!

jtymes: Ok, I cleaned up those that were related to events. However, there are a few other places that use `SubGroup*` camel casing. To keep this PR focused, I left them alone for now.

svanharmelen (approved):
Tweaked the `SubGroup` stuff a bit.. Against my previous comment I think that in this very exceptional case I prefer to use `SubGroup` instead of `Subgroup`...

Thanks for your PR @jtymes 👍🏻

PR #1429, opened 2022-03-26
Support `shared_runners_enabled` field in group response

timofurrer:
See [docs](https://docs.gitlab.com/ee/api/groups.html#:~:text=17T07%3A47%3A25.881Z%22%2C-,%22shared_runners_enabled%22%3A%20true%2C,-%22creator_id%22%3A%201%2C%0A%20%20%20%20%20%20%22namespace)


svanharmelen (approved):
Thanks @timofurrer 👍🏻

PR #1430, opened 2022-03-26
Support `namespace_id` for `GET /users/...` as admin

timofurrer:
See the [docs](https://docs.gitlab.com/ee/api/users.html#:~:text=For%20admins-,The%20namespace_id%20field%20in%20the%20response%20was%20introduced%20in%20GitLab%2014.10.,-GET%20/users)

I've introduced this field for GitLab 14.10 and require it in the provider :)


svanharmelen (approved):
Thanks 🎉

PR #1434, opened 2022-03-26
Support GET endpoints to retrieve single GAT and PAT

timofurrer:
I've recently implemented this in GitLab and is available since 14.10:

* https://gitlab.com/gitlab-org/gitlab/-/merge_requests/82714

### DIFF
--- a/group_access_tokens.go
+++ b/group_access_tokens.go
@@ -57,8 +57,7 @@ func (v GroupAccessToken) String() string {
 // https://docs.gitlab.com/ee/api/group_access_tokens.html#list-group-access-tokens
 type ListGroupAccessTokensOptions ListOptions
 
-// ListGroupAccessTokens gets a list of all Group Access Tokens in a
-// group.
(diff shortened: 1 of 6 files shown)


group_access_tokens.go
```diff
@@ -61,6 +61,31 @@ type ListGroupAccessTokensOptions ListOptions
...
+	gat := new(GroupAccessToken)
+	resp, err := s.client.Do(req, &gat)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return gat, resp, err
+}
+
+// GetGroupAccessToken gets a single Group Access Tokens in a group.
```
jtymes, line 86: I think the docs got switched around. This should be `ListGroupAccessTokens`.

jtymes (changes requested):

group_access_tokens.go
timofurrer: good catch, thanks. Still puzzled how that could have happened 🤔

jtymes (approved):

PR #1438, opened 2022-03-30
Adding tokens LastUsedAt field

notofir:
* The field isn't documented in the official documentation, but found in both project and group access tokens responses.
* `last_used_at: null` is returned in case the token was never used.
* Personal access token doesn't have this field.


svanharmelen (approved):
Rebased and fixed the merge conflicts, looks good now. Thanks @notofir!

PR #1441, opened 2022-03-30
Fixed CreatePipelineOptions variable type

### LINKED ISSUE #1440: Incorrect type for Variables in CreatePipelineOptions struct
The type  requested is *[]*PipelineVariable, but that's not really possible or needed as far as I can tell.

cbbm142:
Changed variable type to something that will work, fixes #1440


svanharmelen:
I'm going to close this as its an unneeded change. Let's discuss further in #1440 if you have any concerns that we need to take into account. Thanks!

PR #1445, opened 2022-04-04
Update runner struct and APIs based on deprecated values

### LINKED ISSUE #1444: Runner API Options no longer match GitLab API
It looks like GitLab has added some options for fields when registering a new runner, namely:

- Paused (deprecates Active)
- Access Level
- Maintenance Note

Oddly, the Maintenance Note is not supported in the Update API (or it may also be undocumented, but as expressed in a different issue, it seems like we should avoid undocumented fields for now).

We should update the Runner resources to support this. I'll submit a PR shortly to do so.

PatrickRice-KSC:
This PR adds the "paused" value to the runner and runner details structs. It also adds several new fields to the RegisterNewRunnerOptions struc. I added a deprecation note to the details structs around the "Active" value, but since the deprecation notice for those fields is over 1 year out, I left them in place.

This aligns the RegisterNewRunnerOptions struct to the GitLab API documentation: https://docs.gitlab.com/ee/api/runners.html#register-a-new-runner

I know there have been conversations around alphabetical vs GitLab API ordering, I maintained the GitLab API ordering for this PR.

Closes #1444

Let me know if you have any questions!

### DIFF
--- a/runners.go
+++ b/runners.go
@@ -37,6 +37,7 @@ type Runner struct {
 	ID          int    `json:"id"`
 	Description string `json:"description"`
 	Active      bool   `json:"active"`
+	Paused      bool   `json:"paused"`
 	IsShared    bool   `json:"is_shared"`
 	IPAddress   string `json:"ip_address"`
 	RunnerType  string `json:"runner_type"`
@@ -47,10 +48,14 @@ type Runner struct {
 }
 
 // RunnerDetails represents the GitLab CI runner details.
+// 
+// "Active" and "Scope" have been deprecated by GitLab. Use "Paused" and "Status" respectively.
+// They will be removed in GitLab 16.0
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/runners.html
 type RunnerDetails struct {
 	Active       bool       `json:"active"`
+	Paused       bool       `json:"paused"`
 	Architecture string     `json:"architecture"`
 	Description  string     `json:"description"`
 	ID           int        `json:"id"`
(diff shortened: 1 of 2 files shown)


runners.go
```diff
@@ -384,14 +393,17 @@ func (s *RunnersService) ListGroupsRunners(gid interface{}, opt *ListGroupsRunne
...
+	Info            *RegisterNewRunnerInfoOptions `url:"info,omitempty" json:"info,omitempty"`
+	Active          *bool                         `url:"active,omitempty" json:"active,omitempty"`
+	Paused          *bool                         `url:"paused,omitempty" json:"paused,omitempty"`
+	Locked          *bool                         `url:"locked,omitempty" json:"locked,omitempty"`
+	RunUntagged     *bool                         `url:"run_untagged,omitempty" json:"run_untagged,omitempty"`
+	TagList         *[]string                     `url:"tag_list[],omitempty" json:"tag_list,omitempty"`
+	AccessLevel     *string                       `url:"access_level,omitempty" json:"access_level,omitempty"`
+	MaximumTimeout  *int                          `url:"maximum_timeout,omitempty" json:"maximum_timeout,omitempty"`
+	MaintenanceNote *string                       `url:"maintenance_note,omitempty" json:"maintenance_note,omitempty"`
 }
```
PatrickRice-KSC, line 407: Note: The reason these all flag as "changed" even though only 3 rows were added (Paused, AccessLevel, MaintenanceNote) is because I had to add one space to all the other rows to maintain formatting since the MaintenanceNote variable name was 1 character longer than Maximum Timeout 😄 

Easy to miss, so I figured I'd leave a comment.

runners.go
```diff
@@ -162,11 +168,14 @@ func (s *RunnersService) GetRunnerDetails(rid interface{}, options ...RequestOpt
 
 // UpdateRunnerDetailsOptions represents the available UpdateRunnerDetails() options.
 //
+// "Active" in deprecated, use "Paused" instead. "Active" will be removed in GitLab 16.0
```
jtymes, line 171:
```suggestion
// "Active" is deprecated, use "Paused" instead. "Active" will be removed in GitLab 16.0
```

PatrickRice-KSC: Ack, thanks for the catch @jtymes !

svanharmelen (approved):
Tweaked the deprecation messages, but looks fine otherwise 👍🏻

PatrickRice-KSC:
Thanks @svanharmelen !  - Would you mind tagging the repo so we can use this in the downstream terraform provider? 🙂

svanharmelen:
Done... `v0.62.0`

PR #1446, opened 2022-04-05
Add "bot" support for Users API

apendragon:
Since Project Access Token and Group Access Token, there is a "bot"
field in the User json structure returned by the API.

See: https://docs.gitlab.com/ee/api/users.html#single-user

Also, from 14.10 it is possible to exclude bots from search result using
the "without_project_bots" boolean request parameter.

It has been added in the ListUsersOptions struct.

### DIFF
--- a/users.go
+++ b/users.go
@@ -75,6 +75,7 @@ type User struct {
 	WebsiteURL                     string             `json:"website_url"`
 	Organization                   string             `json:"organization"`
 	JobTitle                       string             `json:"job_title"`
+	Bot                            bool               `json:"bot"`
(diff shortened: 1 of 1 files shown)


users.go
```diff
@@ -75,6 +75,7 @@ type User struct {
 	WebsiteURL                     string             `json:"website_url"`
 	Organization                   string             `json:"organization"`
 	JobTitle                       string             `json:"job_title"`
+	Bot                            bool               `json:"bot"`
```
svanharmelen, line 78: Please place the new field between `Bio` and `Location` following the order of the docs...

apendragon: Sure.

PR #1448, opened 2022-04-11
added vulnerabilities

rotiv13:


rotiv13:
Just seeing if it possible to get implemented the vulnerabilities for each repo.

svanharmelen:
Made this a draft PR as you also updated the `go.mod` file to point at your own fork... Additionally (as a premature review comment) the field types of the struct should not be `interface{}`, so please make sure they are of the correct type before finishing the PR. Thanks!

rotiv13:
> Made this a draft PR as you also updated the `go.mod` file to point at your own fork... Additionally (as a premature review comment) the field types of the struct should not be `interface{}`, so please make sure they are of the correct type before finishing the PR. Thanks!

Yep makes sense. Thank you for the comment. I am basically experimenting with this and seeing if it is possible to get the vulnerabilities of a project. I will make it better once I see that everything is ready.

rotiv13:
@svanharmelen if you can do a review when you have some time.

svanharmelen (approved):
Thanks @rotiv13! Tweaked a few small things, but looks good otherwise 👍🏻

PR #1460, opened 2022-05-03
Include assignee changes into IssueEvent

abezzub:
I see there is already support for this in [MergeEvent](https://github.com/xanzy/go-gitlab/blob/master/event_webhook_types.go#L529-L532), issues similarly include list of users who were assigned to the issue and list of users who are currently assigned to the issue. 

This PR exposes `Assignees` on `IssueEvent` the same way as it is done on the `MergeEvent`,


svanharmelen (approved):
Thanks @abezzub 👍🏻

PR #1461, opened 2022-05-04
Implementation of Agents for Kubernetes APIs

timofurrer:
This change sets adds two new services:

* `ClusterAgents`
* `ClusterAgentTokens`

The former is already available in 14.9 and the latter is anticipated to land in 15.0, see https://gitlab.com/gitlab-org/gitlab/-/merge_requests/85704/

I suggest to wait for a merge of this PR until everything is merged and available on gitlab.com.


timofurrer:
The MR on GitLab has been merged and the API is this one: https://docs.gitlab.com/ee/api/cluster_agent_tokens.html

timofurrer:
@svanharmelen ready for your review!

svanharmelen (approved):
Did quite some refactoring to bring this inline with the docs where these endpoint fall under one service and have different names. But it should be good now.. Or at least inline with the current docs...

PR #1467, opened 2022-05-09
Add ThemeID in user creation/modification

cyd01:
Add ability to set/change theme ID pour a user.


svanharmelen (approved):
Thanks @cyd01 👍🏻

PR #1472, opened 2022-05-14
Support pipeline iid

WolfspiritM:
According to docs `GET /projects/:id/pipelines` returns the project internal id via `iid` since gitlab 14.6


svanharmelen (approved):
👍🏻

PR #1488, opened 2022-06-02
Add missing ParentID to ProjectNamespace

pschwartz:


pschwartz:
@svanharmelen Noticed this one was missing while using the project to identify what namespace was it's parent.

PR #1489, opened 2022-06-03
Add support to the field failure_reason

danifbento:
Following GitLab API documentation we have:
[https://docs.gitlab.com/ee/api/jobs.html](https://docs.gitlab.com/ee/api/jobs.html)

It was added a field called `failure_reason` which returns the failure reason for a given job.

Signed-off-by: Daniela Bento <danibento@overdestiny.com>


svanharmelen (approved):
Did a little reordering, but looks good 👍🏻  Thanks!

Issue #815, opened 2020-03-30
Add support for new list pagination


klmitch:
gitlab is experimenting with a new means of performing pagination that promises to be more efficient for them; they're calling it ["Keyset-based pagination"](https://gitlab.com/help/api/README.md#keyset-based-pagination), and it uses the existing "per_page" query parameter and a new "pagination" parameter (set to the string "keyset").  It's intended to be used in conjunction with the `Link` header, which may specify some additional query parameters (their example includes "order_by", "sort", and "id_after").  It would be nice if go-gitlab could support this, so that go applications utilizing it could try out this new form of pagination.

(As of right now, it appears to only be implemented on the projects endpoint, with "order_by=id".  My read from discussions with them regarding this feature is that they intend to implement it across the board over the long term, though.)

svanharmelen:
I'm not sure if I want to add support for this now already. And if we do, I'm not yet sure how we should support both the old and new solutions concurrently. Guess I need to have a closer look to find out and form an opinion...

Of course if you have specific ideas and/or am willing to create a PR, I'm always open to look at that or discuss it.

klmitch:
Yeah, my purpose in opening this issue was to make sure to document it so everyone interested can begin to think about what it should look like.

k-mckinney:
I noticed today on the documentation for the [repository tree API](https://docs.gitlab.com/ee/api/repositories.html#list-repository-tree) that this endpoint has deprecated support for the `page` pagination in favor of keyset pagination, and it will be unsupported as of Gitlab 15.0.

Shocktrooper:
As of right now all I can find in terms of deprecation of the old pagination for supported endpoints in go-gitlab is what @k-mckinney mentioned in the above thread.  With knowledge at least this endpoint we should completely transition this over before 15.0 and notify maintainers of downstream projects if we can. @timofurrer I looked and didn't see anything using this endpoint in the terraform provider.  If you could do a once over when you have time it would be appreciated.

mitar:
Could we just add `Pagination` field to `ListOptions` for now? So that people can use it if they want. We can add any sugar on top later on.

timofurrer:
> I looked and didn't see anything using this endpoint in the terraform provider. If you could do a once over when you have time it would be appreciated.

@Shocktrooper Yes, we are not using this API right now.

pwlandoll:
#1827 has been merged, is that functionality sufficient to close this issue?

klmitch:
[...]

PR #1496, opened 2022-06-09
Idea for managed pagination into channels for resources

catouc:
For some context I maintain a package that does this already inside of the company I work for. The semantics have proven pretty useful and seem to integrate well in this library so I thought upstreaming this could be useful to the wider community.

I have written up a sample implementation for the projects API but this can essentially be repeated for every paginated resource.
Ultimately we can also think about implementing [!815](https://github.com/xanzy/go-gitlab/issues/815) behind the scenes in this down the line :thinking: 

But before I go implement loads of endpoints I thought it's worth checking if this is desired to have this or if this should be a separate more specialised package?

Happy to have a discussion!

### DIFF
--- a/examples/projects.go
+++ b/examples/projects.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"fmt"
 	"log"
 
 	"github.com/xanzy/go-gitlab"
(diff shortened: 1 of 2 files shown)


catouc:
Can' t really see why the tests are failing :thinking: unless they use the ` examples`  and are unhappy.

svanharmelen:
@catouc thanks for your PR, yet I need to think a little bit about this one. Until now this package tries to follow the GitLab API as close as possible, but this PR introduces a custom helper function that doesn't exist as a GitLab API. So I'm not sure yet if I want to open the doors for such functions...

catouc:
I had an idea that we could put this into a sub-package called `streaming` or something? Then it's logically separated more from the core library :thinking:

timofurrer:
@svanharmelen @catouc I've done some small experiment to make this somewhat sustainable, see https://github.com/xanzy/go-gitlab/pull/1498

catouc:
I maintain something similar in my internal library which has a function called `getResourcePage` that basically just accepts the resource URI, that made it somewhat sustainable without generics, of course still a lot of duplication since every List call gets its own wrapper, but I'm not sure if that cost is much different from the current cost of calls :thinking: 
Overall this looks pretty neat as an idea I think. Though the interface gets a bit confusing from the user end right now I feel and if we wrap this implementation with specific calls again, how much have we gained?

But yeah the essentially benefit of this is that you can use channels because it allows you to work page by page on what you need to do, which starts to matter once you have 10k+ resources and need to do stuff across all of them.

catouc:
One passing thought, it's probably worth looking at making the transition over to keyset pagination as smooth as possible, which does not deal with page numbers and offsets the same way.

I have a [bare bones implementation](https://github.com/catouc/gitlab-ci-crawler/blob/main/internal/gitlab/gitlab.go#L72) ([link header stuff](https://github.com/catouc/gitlab-ci-crawler/blob/main/internal/gitlab/linkheader.go)) that's been working pretty okay so far but most of the client call need to be able to call full URLs because that's what the keyset pagination `Next` headers return, thought I'd drop that here additionally while I still remember 😄

svanharmelen:

PR #1497, opened 2022-06-12
Add support for merged YAML return from linter

robbydyer:
The lint API requires setting the `include_merged_yaml` option to true to in order for it to return the merged YAML. Since the existing `Lint` function doesn't take options as a parameter, this adds a new function that sets that option to true. https://docs.gitlab.com/ee/api/lint.html#validate-the-ci-yaml-configuration

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -29,6 +29,15 @@ type ValidateService struct {
 	client *Client
 }
 
+// LintOptions represent options for the lint API
+//
+// Gitlab API docs: https://docs.gitlab.com/ee/api/lint.html#validate-the-ci-yaml-configuration
+type LintOptions struct {
+	Content           string `url:"content,omitempty" json:"content,omitempty"`
+	IncludeMergedYAML bool   `url:"include_merged_yaml,omitempty" json:"include_merged_yaml,omitempty"`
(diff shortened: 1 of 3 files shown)


validate.go
```diff
@@ -73,6 +73,31 @@ func (s *ValidateService) Lint(content string, options ...RequestOptionFunc) (*L
...
+	}
+
+	l := new(LintResult)
+	resp, err := s.client.Do(req, l)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return l, resp, nil
+}
```
svanharmelen, line 99: Instead of making a dedicated function like this the general approach is to make an option struct and pass that into the existing `Lint` function.

I do notice that the existing `Lint` function already does this different, but the best way forward is to correct that (now that it caught my eye 😉) instead of repeating it...

svanharmelen (changes requested):

validate.go
robbydyer: Would you prefer I make an entirely new function to maintain backward compatibility? Like `LintWithOptions` or something?

svanharmelen (approved):
Thanks @robbydyer! I updated the PR so your version replaces the incomplete `Lint` function that already existed.

I understand this is now a breaking change, but the old version was clearly not implemented correctly (not following the actual API) so in this case I am good with that...

PR #1498, opened 2022-06-13
Add pagination helper to easily collect all items

timofurrer:
🚧  **THIS IS MERELY AN EXPERIMENT FOR NOW!** 🚧 

The https://github.com/xanzy/go-gitlab/pull/1496 Pull Request triggered
some thinking for how to do this in a more "generic" way.

Thus, I drafted this `gitlab.Collect()` function, which can be used to
"easily" collect all items of a specific "list" API (I've implement a
*collect* function which eagerly collects all items, but something
similar can be implemented for channels to stream).

The only downside of the current implementation so far is that Go
doesn't really support generics / interfaces with direct field access as
of now ... effectively, you'll always need methods.
This means that, because the `ListOptions` field is somewhat
inconsistently (sometimes as subfield, sometimes it's promoted)
implemented over this package every `gitlab.List*Options` struct needs
it's own `GetPage() int` and `SetPage(int)` implementation ...

One improvement for this could be that we split out the `ListOptions`
from the API endpoint specific options and make it an additional
argument to the list methods - so that `Collect()` could profit from
this and `GetPage() int` and `SetPage(int)` would only need to be
implemented once.

Another one would be to just use `gitlab.ListOptions` for endpoints
which don't have additional parameters instead of doing this `type
ListInstanceVariablesOptions ListOptions`.

However, I think best would be to separate the generic `ListOptions` to
a separate argument in the long term ...

@svanharmelen what do you think?

### DIFF
--- a/examples/pagination_helper.go
+++ b/examples/pagination_helper.go
@@ -1,0 +1,67 @@
+//
+// Copyright 2021, Timo Furrer <tuxtimo@gmail.com>
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package main
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/xanzy/go-gitlab"
+)
+
+func main() {
+	git, err := gitlab.NewClient("yourtokengoeshere")
+	if err != nil {
+		log.Fatal(err)
+	}
+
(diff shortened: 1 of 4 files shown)


2785:
A while back I actually wrote a flavor of this, driven by generics & gammazero's deque that vaguely works like this

```go
func NewProjectMRIterator(ctx context.Context, pid string, opts gitlab.ListProjectMergeRequestsOptions) *ListItemIterator[*gitlab.MergeRequest] {
	return NewListItemIterator(ctx, func(ctx context.Context, page int) (results []*gitlab.MergeRequest, nextPage int, err error) {
		glClient, err := GetGlobalGitlabClient()
		if err != nil {
			return nil, 0, err
		}

		opts.Page = page

		mrs, resp, err := glClient.MergeRequests.ListProjectMergeRequests(pid, &opts, gitlab.WithContext(ctx))
		if err != nil {
			return nil, 0, err
		}

		return mrs, resp.NextPage, nil
	})
}
```

(if you ignore the global client business)

It's basically

```go
type Iterator[T any] interface {
	Next() (T, error)
}
```

but backed by a queue & goes to fetch a new batch with a user supplied function, when the queue is empty.

Can put a PR up if yall fancy.

svanharmelen:

PR #1499, opened 2022-06-14
Add skip_confirmation option to add email API

cyd01:
According to the [Gitlab API](https://docs.gitlab.com/ee/api/users.html#add-email-for-user) I propose to add `skip_confirmation` parameter to
- AddEmail
- AddEmailForUser

`UsersService` functions.

### DIFF
--- a/users.go
+++ b/users.go
@@ -814,10 +814,10 @@ func (s *UsersService) GetEmail(email int, options ...RequestOptionFunc) (*Email
 
 // AddEmailOptions represents the available AddEmail() options.
 //
-// GitLab API docs: https://docs.gitlab.com/ce/api/projects.html#add-email
+// GitLab API docs: https://docs.gitlab.com/ce/api/users.html#add-email
 type AddEmailOptions struct {
 	Email            *string `url:"email,omitempty" json:"email,omitempty"`
-	SkipConfirmation *bool   `url:"skiconfirmation,omitempty" json:"skiconfirmation,omitempty"`
+	SkipConfirmation *bool   `url:"skip_confirmation,omitempty" json:"skip_confirmation,omitempty"`
 }
 
 // AddEmail creates a new email owned by the currently authenticated user.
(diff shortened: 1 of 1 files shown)


users.go
```diff
@@ -816,7 +816,8 @@ func (s *UsersService) GetEmail(email int, options ...RequestOptionFunc) (*Email
 //
 // GitLab API docs: https://docs.gitlab.com/ce/api/projects.html#add-email
```
svanharmelen, line 817:
```suggestion
// GitLab API docs: https://docs.gitlab.com/ce/api/users.html#add-email
```

users.go
```diff
@@ -816,7 +816,8 @@ func (s *UsersService) GetEmail(email int, options ...RequestOptionFunc) (*Email
...
 type AddEmailOptions struct {
-	Email *string `url:"email,omitempty" json:"email,omitempty"`
+	Email            *string `url:"email,omitempty" json:"email,omitempty"`
+	SkipConfirmation *bool   `url:"skiconfirmation,omitempty" json:"skiconfirmation,omitempty"`
```
svanharmelen, line 820:
```suggestion
	SkipConfirmation *bool   `url:"skip_confirmation,omitempty" json:"skip_confirmation,omitempty"`
```

svanharmelen (changes requested):
Thanks @cyd01, but it does need a fix (and also a fix for something you didn't add if you don't mind 😉) before I can merge it.

users.go
cyd01: I'm really sorry. I don't known what i was thinking at when i've committed it !
But it is very weird. It works !
Anyway, I'll fix it in a minute.

users.go
cyd01: Just a question (I'm new in Golang): I've propose few pull requests in last weeks. I modified some source code, but I've never proposed documentation update. Is it auto-generated, or must I do something ?

svanharmelen (approved):
Thanks @cyd01 👍🏻

users.go
svanharmelen: Docs are not generated, just added and maintained by hand...

cyd01: Ok
So, what can i do to help you to maintain the doc?

svanharmelen: Thanks, but not much to be honest. Of course make sure then when making a PR you add the correct links, but other then that not much...

PR #1500, opened 2022-06-27
Including the tag_message field for annotated tag releases

LewisTheobald:


releases.go
```diff
@@ -126,6 +126,7 @@ func (s *ReleasesService) GetRelease(pid interface{}, tagName string, options ..
 type CreateReleaseOptions struct {
 	Name        *string               `url:"name" json:"name"`
 	TagName     *string               `url:"tag_name" json:"tag_name"`
+	TagMessage  *string               `url:"tag_message" json:"tag_message"`
```
svanharmelen, line 129: Hmm... Could you add `omitempty` for all these fields (`Name`, `tagName`, `TagMessage` and `Description`) so they are consistent and inline with the rest again? Not sure when this happened, but in option structs this should be the standard...

LewisTheobald: Added `omitempty` onto all of them!

svanharmelen (approved):
Thanks @LewisTheobald 👍🏻

PR #1506, opened 2022-07-16
Fully support project and group avatars

timofurrer:
This change set adds support for project, group and user avatars in the
same fashion this is already implemented for topics.

Currently it's not possible to remove avatars from those API - I'm
exploring possible solutions
[here](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/92604).

Nonetheless, I think this PR can already be merged.

### DIFF
--- a/groups.go
+++ b/groups.go
@@ -18,10 +18,13 @@ package gitlab
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
 	"time"
+
+	retryablehttp "github.com/hashicorp/go-retryablehttp"
 )
 
 // GroupsService handles communication with the group related methods of
@@ -88,6 +91,15 @@ type Group struct {
 type GroupAvatar struct {
 	Filename string
 	Image    io.Reader
+}
+
+// MarshalJSON implements the json.Marshaler interface.
+func (a *GroupAvatar) MarshalJSON() ([]byte, error) {
+	if a.Filename == "" && a.Image == nil {
+		return []byte(`""`), nil
+	}
+	type alias GroupAvatar
+	return json.Marshal((*alias)(a))
 }
 
 // LDAPGroupLink represents a GitLab LDAP group link.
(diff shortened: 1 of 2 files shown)


groups.go
```diff
@@ -90,6 +93,15 @@ type GroupAvatar struct {
...
 }
 
+// MarshalJSON implements the json.Marshaler interface.
+func (a *GroupAvatar) MarshalJSON() ([]byte, error) {
+	if a.Filename == "" && a.Image == nil {
+		return []byte(`""`), nil
+	}
+	type alias GroupAvatar
+	return json.Marshal((*alias)(a))
+}
```
svanharmelen, line 103: Can you explain why this is needed?

groups.go
```diff
@@ -420,6 +448,7 @@ func (s *GroupsService) TransferSubGroup(gid interface{}, opt *TransferSubGroupO
 type UpdateGroupOptions struct {
 	Name                                 *string                     `url:"name,omitempty" json:"name,omitempty"`
 	Path                                 *string                     `url:"path,omitempty" json:"path,omitempty"`
+	Avatar                               *GroupAvatar                `url:"-" json:"avatar,omitempty"`
```
svanharmelen, line 451: Shouldn't this be:
```suggestion
	Avatar                               *GroupAvatar                `url:"-" json:"-"`
```

svanharmelen:
Do you still need or want this? If so could you maybe cleanup the PR and reply to my questions? Then we can try to get it sorted...

groups.go
timofurrer: If we wouldn't have this the field wouldn't be sent. In the current implementation it will send an empty string which effectively means to remove the avatar, e.g.:

```
curl --request PUT --header "PRIVATE-TOKEN: <your_access_token>" "https://gitlab.example.com/api/v4/groups/22" \
     --data "avatar="
```

groups.go
timofurrer: hmmm, good question. I think in case of `UploadRequest()` the options are any ways never JSON encoded, but just form fields ... So, probably it doesn't matter. I have to double check.

PR #1509, opened 2022-07-24
Fix some GPG key APIs

yan12125:
Noticed when I'm working on
https://github.com/gitlabhq/terraform-provider-gitlab/issues/1166

PR #1512, opened 2022-07-27
Fix another GPG API

yan12125:
Following up https://github.com/xanzy/go-gitlab/pull/1509. This is also noticed when I'm working on https://github.com/gitlabhq/terraform-provider-gitlab/issues/1166. Probably GPG APIs should be properly tested.


timofurrer (approved):
Did you verify if any other GPG related APIs are also wrong?

yan12125:
I only check by eyes and they look correct.

svanharmelen (approved):
Thanks @yan12125!

PR #1514, opened 2022-08-02
update github.com/stretchr/testify to v1.8.0

lufia:
Though it is issue on Kubernetes, but Dependabot alerts to `gopkg.in/yaml.v2 < 2.2.8` versions: CVE-2019-11254.
**yaml.v2** is imported by only **stretchr/testify**, so I updated it.


svanharmelen:
Thanks for pointing out that we needed some updates! Updated all deps using PR #1516 so these changes are now already in master. Thanks again, but this PR on its own is now not needed anymore so I'll close it in favor of #1516

lufia:
:+1:

PR #1522, opened 2022-08-09
Add DeletePackageFile in PackagesService

cyd01:
Add DeletePackageFile function in PackagesService.

Here is the [Gitlab API documentation](https://docs.gitlab.com/ee/api/packages.html#delete-a-package-file).


svanharmelen (approved):
Thanks @cyd01 👍🏻

PR #1525, opened 2022-08-15
Updated User struct to support pagination

6fears7:
Updated to be able to support queries a la, "per_page=30". Couldn't figure out for the life of me why users weren't returning


svanharmelen:
Hi @6fears7, the `ListUsersOptions` struct already has an embedded struct called `ListOptions` which contains the needed fields to do pagination. So adding this one specific field is not needed.

Issue #1526, opened 2022-08-16
Support for project feature flags


doriandekoning:
It seems like gitlab feature flags are supported but not the project feature flags: https://docs.gitlab.com/ee/api/feature_flags.html.

These would be very useful, am I missing them or are they not supported yet? In that case I could open a pull request if it will be accepted.

svanharmelen:
PRs are accepted and welcome! Please do look at the existing code and use the same logic and style! Thanks...

svanharmelen:
This one is already solved by #1528

PR #1528, opened 2022-08-17
feat: add support for project feature flags

doriandekoning:
As discussed in https://github.com/xanzy/go-gitlab/issues/1526


svanharmelen (approved):
This PR is not inline with the rest of the package. I'll merge it and will then push a commit to straighten things out. Thanks...

PR #1531, opened 2022-08-23
Fix SAML Group Links Access Level type

timofurrer:
While working on https://github.com/gitlabhq/terraform-provider-gitlab/pull/1215
I have noticed that the access level type introcued in
https://github.com/xanzy/go-gitlab/pull/1527
is wrong.

@svanharmelen Do you mind creating a bugfix release with this in it?


svanharmelen (approved):
Thanks, bugfix release is coming up...

mhodgson:
Thanks, this was changed after the API was released and deployed to production.

PR #1537, opened 2022-08-30
enable setting job_variables_attributes when running a job

ctkeyser:
param seems to have been added more recently https://docs.gitlab.com/ee/api/jobs.html#run-a-job


svanharmelen (approved):
Tweaked the code just a little, but looks good otherwise! Thanks 👍🏻

Issue #1538, opened 2022-08-31
Mismatch between `public_builds` and `public_jobs`


mtneug:
The projects API will respond with a `public_jobs` field, while `public_builds` should be used when setting this field according to the documentation. This is already known upstream. It appears that `public_jobs` can be used for setting and retrieving.

See <https://gitlab.com/gitlab-org/gitlab/-/issues/329725>
See <https://github.com/gitlabhq/terraform-provider-gitlab/issues/1233>

svanharmelen:
Feel free to create a PR to fix it...

PR #1541, opened 2022-09-01
Expose project public_jobs field

gravis:
This MR fixes https://github.com/xanzy/go-gitlab/issues/1538, at least temporary, until https://gitlab.com/gitlab-org/gitlab/-/issues/329725 is fixed. Nothing more is actually needed since this field will be populated as expected now in [GetProject](https://github.com/xanzy/go-gitlab/blob/4256ad62c1044a01b0eab2fd863f53807d562876/projects.go#L521), but the field `public_builds` is still the one being used for [updates](https://github.com/xanzy/go-gitlab/blob/4256ad62c1044a01b0eab2fd863f53807d562876/projects.go#L843) and [creations](https://github.com/xanzy/go-gitlab/blob/4256ad62c1044a01b0eab2fd863f53807d562876/projects.go#L647).

We can fix those when GitLab fixes their API.

refs https://github.com/xanzy/go-gitlab/issues/1538

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -83,6 +83,7 @@ type Project struct {
 	StarCount                                 int                        `json:"star_count"`
 	RunnersToken                              string                     `json:"runners_token"`
 	PublicBuilds                              bool                       `json:"public_builds"`
+	PublicJobs                                bool                       `json:"public_jobs"`
 	AllowMergeOnSkippedPipeline               bool                       `json:"allow_merge_on_skipped_pipeline"`
 	OnlyAllowMergeIfPipelineSucceeds          bool                       `json:"only_allow_merge_if_pipeline_succeeds"`
 	OnlyAllowMergeIfAllDiscussionsAreResolved bool                       `json:"only_allow_merge_if_all_discussions_are_resolved"`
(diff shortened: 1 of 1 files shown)


projects.go
```diff
@@ -83,6 +83,7 @@ type Project struct {
 	StarCount                                 int                        `json:"star_count"`
 	RunnersToken                              string                     `json:"runners_token"`
 	PublicBuilds                              bool                       `json:"public_builds"`
+	PublicJobs                                bool                       `json:"public_jobs"`
```
svanharmelen, line 86: I guess we should move `PublicBuilds` to a `// Deprecated members` block as done [here](https://github.com/xanzy/go-gitlab/blob/754673b7bf0a19511b5363ccf7b60fd07ed0ebf0/runners.go#L88) for example...

svanharmelen (changes requested):
Looks good, but have a small suggestion...

projects.go
gravis: Actually, since the field isn't returned anymore by the API, what do you think of removing it completely? It's always `false` anyway, so it can lead to confusion.  
Otherwise, I'm fine moving it to a `// Deprecated members` block as you suggested.

svanharmelen: It's still needed if you want to talk to an older version of GitLab, so I prefer to keep it...

svanharmelen (approved):
Thanks @gravis 👍🏻

PR #1542, opened 2022-09-04
add OnFinished to gitlab.Client

lzhseu:
In this PR, a hook function OnFinished is added in gitlab client. It will be invoked at the end of each HTTP request, whether the result succeeds or fails. This feature will be very useful, which allows users to do some extra thing with the response and error. By the way, our team really needs this feature urgently. We want to monitor the request for gitlab server, and alarm when the call failure rate is greater than the threshold.

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -64,6 +64,11 @@ const (
 	PrivateToken
 )
 
+// OnFinished allows running a function on each HTTP request.
+// This function will be invoked at the end of every HTTP
+// request executed whether the result succeeds or fails
+type OnFinished func(response *Response, err error)
+
 // A Client manages communication with the GitLab API.
 type Client struct {
 	// HTTP client used to communicate with the API.
@@ -98,6 +103,10 @@ type Client struct {
 
 	// User agent used when communicating with the GitLab API.
 	UserAgent string
+
+	// OnFinished allows a user-supplied function to be called
+	// with the response and error at the end of each HTTP request executed.
+	OnFinished OnFinished
 
 	// Services used for talking to different parts of the GitLab API.
 	AccessRequests          *AccessRequestsService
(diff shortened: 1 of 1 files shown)


svanharmelen:
Hmm... I understand what you are trying to do, but I doubt that this kind of logic should be inside of the GitLab client. I try to keep the client in sync with the functionality described in the docs and not add any custom logic.

I'm pretty sure you can achieve something like this by either wrapping or embedding the client or by configuring the client with a custom HTTP client that has some extra logic for something like this.

I'll think about it during the weekend, but I don't think I'm wiling to merge this. Sorry...

lzhseu:
Thank you very much for your reply~  In my opinion, this feature has no intrusion into the original code logic, but enhance the functionality.

svanharmelen:
You can embed the client is a struct of your own and then "overwrite" the `Do()` method with a custom method which then in turn calls the clients' `Do()` method.

> In my opinion, this feature has no intrusion into the original code logic, but enhance the functionality.

That is true, but it adds non-GitLab logic/functionality to the client which I like to prevent. If you open the door for something additional/custom, then where do you draw the line when the next PR comes in? So I like to draw the line by only trying to support all GitLab API documented functionality.

lzhseu:
OK，I know your concerns. Thank you for your reply and your suggestions. Thank you very much!

lzhseu:
Sorry, I still have some questions.
> You can embed the client is a struct of your own and then "overwrite" the Do() method with a custom method which then in turn calls the clients' Do() method.
I don't understand what this mean?  I can define a client struct that wraps the gitlab client, and "overwrite" `Do()` method which then in turn calls gitlab clients' `Do()` method. However, in the existing API, it calls gitlab client's `Do()` method. Where to call the "overwrite" `Do()` method? 

I don't know how to use this project(go-gitlab) directly, or do you mean I should modify this project first, and then use it. 

I would appreciate it if you could answer my questions!

svanharmelen:
Not tested, but something like this should work:

```go
package main

import (
	"log"

	"github.com/hashicorp/go-retryablehttp"
	"github.com/xanzy/go-gitlab"
)

type myClient struct {
	*gitlab.Client
}

func (c *myClient) Do(req *retryablehttp.Request, v interface{}) (response *gitlab.Response, err error) {
	defer func() {
		// Do something after the request
	}()
	return c.Client.Do(req, v)
}

// This example shows how to create a client with username and password.
func main() {
	client, err := gitlab.NewClient("yourtokengoeshere")
	if err != nil {
		log.Fatal(err)
	}

	git := myClient{client}

	// List all projects
	projects, _, err := git.Projects.ListProjects(nil)
	if err != nil {
		log.Fatal(err)
	}

	log.Printf("Found %d projects", len(projects))
}
```

svanharmelen:
Hmm... No that will not work as the calls to `Do()` are made from within the client itself. So that will call the `Do()` method of the client and not the custom one of your wrapper... So you probably want another solution...

svanharmelen:
You can use your own custom HTTP client... We use the `go-retryablehttp` client, so maybe you can use this: https://pkg.go.dev/github.com/hashicorp/go-retryablehttp#ResponseLogHook

lzhseu:
I noticed ResponseLogHook, but it is invoked when the request is successful, so it doesn't help. Hmm... maybe I have to find other solutions.  Anyway, thank you very much for your time!

svanharmelen:
Yeah, sorry I don't have an off-the-shelf solution but as already discussed I would like to limit the functionality we add to what is described in the GitLab docs. So I'm going to close this PR as it introduces custom logic that is not part of the API specs. Hope you can find a good solution for your usecase!

PR #1548, opened 2022-09-14
when deleting a job artifact the api returns a 204 no content

### LINKED ISSUE #1547: EOF error on delete job artifacts
Hi,

im getting a EOF error on `client.Jobs.DeleteArtifacts(projectId, jobId)`. The API returns a 204 no content (https://docs.gitlab.com/ee/api/job_artifacts.html#delete-job-artifacts). The go-gitlab code seems to expect a job content and therefore decoding the missing body produces the EOF. Is this a leftover from v3 API?

A workaround is to check for io.EOF and ignore this error until fixed. It looks pretty similar to #195

vgropp:
fixes #1547

### DIFF
--- a/jobs.go
+++ b/jobs.go
@@ -540,10 +540,5 @@ func (s *JobsService) DeleteArtifacts(pid interface{}, jobID int, options ...Req
 		return nil, err
 	}
 
-	resp, err := s.client.Do(req, nil)
-	if err != nil {
-		return resp, err
-	}
-
-	return resp, err
+	return s.client.Do(req, nil)
 }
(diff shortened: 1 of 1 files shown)


jobs.go
```diff
@@ -528,23 +528,22 @@ func (s *JobsService) PlayJob(pid interface{}, jobID int, options ...RequestOpti
...
 
 	req, err := s.client.NewRequest(http.MethodDelete, u, nil, options)
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 
-	job := new(Job)
-	resp, err := s.client.Do(req, job)
+	resp, err := s.client.Do(req, nil)
```
svanharmelen, line 543: You can change this in just `return s.client.Do(req, nil)`

svanharmelen (changes requested):
Think we can clean it up just a little...

svanharmelen (approved):
Thanks, looks fine now 👍🏻

PR #1260, opened 2021-10-02
add rest of tests in discussions

venky333:
* add ListMergeRequestDiscussions, GetMergeRequestDiscussion, CreateMergeRequestDiscussion, ResolveMergeRequestDiscussion, AddMergeRequestDiscussionNote, UpdateMergeRequestDiscussionNote, DeleteMergeRequestDiscussionNote in discussions

PR #1551, opened 2022-09-16
fix(commit discussions): Make commit ids strings (close #1546)

### LINKED ISSUE #1546: `*CommitDiscussion` functions: how to convert a commit SHA from `string` to `int`?
Hi! I was investigating upgrading the `lab` project to a newer version of `go-gitlab` (we're currently on 0.51) and ran into an issue about translating commit SHAs (as strings) into ids (as ints) for `CreateCommitDiscussion()`, which led me down a rabbit hole. I'm not very strong in Go, so I'm fumbling around a little bit, but PR 1260 changed all of the `*CommitDiscussion` and `*CommitDiscussionNote` functions to take a `commit int` instead of a `commit string` (which I think was released in 0.56), and I haven't been able to figure out how to convert SHA strings into an int id that GL will understand.

Any pointers?

(FWIW, GL API docs for commits don't mention int ids, in fact they are all explicitly strings; and even the docs for commit discussions have different values for different endpoints: most say it's an `int`, but at least 1 says it's a `string`. Beyond that, I've even been able to pass string SHAs to those other endpoints w/o issue. 🤷 )

claytonrcarter:
Hi, this resolves my issue at #1546 by reverting some of the changes from #1260 and turning the commit id back into a string for the `*CommitDiscussion` functions. Note that -- at this time -- the API docs are still incorrectly stating that commit ids are ints, but [MR 97586](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/97586) (which corrects the docs) ~~appears to be set for merge.~~ (**Update**: that MR has been merged.) Quoting one of the reviewers on that MR: "The assertion that these IDs are in fact SHA strings is correct."

I believe that this is safe to merge now, before that MR is merged and/or released. As is, these functions appear to unusable unless you happen to be working with discussions on the small set of commits whose SHAs can be represented as ints. 😄 

Thanks for your consideration, and for this great library!

### DIFF
--- a/discussions.go
+++ b/discussions.go
@@ -917,12 +917,12 @@ type ListCommitDiscussionsOptions ListOptions
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ce/api/discussions.html#list-project-commit-discussion-items
-func (s *DiscussionsService) ListCommitDiscussions(pid interface{}, commit int, opt *ListCommitDiscussionsOptions, options ...RequestOptionFunc) ([]*Discussion, *Response, error) {
+func (s *DiscussionsService) ListCommitDiscussions(pid interface{}, commit string, opt *ListCommitDiscussionsOptions, options ...RequestOptionFunc) ([]*Discussion, *Response, error) {
 	project, err := parseID(pid)
 	if err != nil {
 		return nil, nil, err
 	}
-	u := fmt.Sprintf("projects/%s/repository/commits/%d/discussions",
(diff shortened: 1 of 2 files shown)


claytonrcarter:
I just want to bump this to clarify that the [GitLab API docs](https://docs.gitlab.com/ee/api/discussions.html#commits) have been updated by the linked MR and this project is now out of date w/ the current API docs, as outlined in the PR description.

svanharmelen (approved):
Sorry for the delay, but looks OK to me.. So I'll take it as is. Thanks 👍🏻

Issue #1536, opened 2022-08-29
LineRange on NotePosition should include omitempty


jcroall:
When trying to create a Merge Request Discussion and Note, the LineRange should be allowed to be empty for notes that occur on a single line not a range.

svanharmelen:
Feel free to create a PR to fix it...

PR #1555, opened 2022-09-22
Allow setting LineRange to null in NotePosition

KqLLL:
Fix [Issue 1536](https://github.com/xanzy/go-gitlab/issues/1536)

### DIFF
--- a/notes.go
+++ b/notes.go
@@ -82,7 +82,7 @@ type NotePosition struct {
 	NewLine      int        `json:"new_line,omitempty"`
 	OldPath      string     `json:"old_path,omitempty"`
 	OldLine      int        `json:"old_line,omitempty"`
-	LineRange    *LineRange `json:"line_range"`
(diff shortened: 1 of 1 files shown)


notes.go
```diff
@@ -73,6 +73,8 @@ type Note struct {
 }
 
 // NotePosition represents the position attributes of a note.
+// GitLab API docs:
+// https://docs.gitlab.com/ce/api/discussions.html#create-new-merge-request-thread
```
svanharmelen, line 77: Lets omit these lines as this struct is not used only for the API call.

svanharmelen (changes requested):
Please remove the additional comments, the code change itself is fine 👍🏻

notes.go
KqLLL: 👌

svanharmelen (approved):
Thanks 👍🏻

PR #1556, opened 2022-09-22
add seperate cache setting to projects

wind0r:
added missing fields from `/settings/ci_cd#js-general-pipeline-settings`


projects.go
```diff
@@ -38,6 +38,7 @@ type ProjectsService struct {
 // GitLab API docs: https://docs.gitlab.com/ce/api/projects.html
 type Project struct {
 	ID                                        int                        `json:"id"`
+	BuildAllowGitFetch                        bool                       `json:"build_allow_git_fetch"`
```
svanharmelen, line 41: I don't see this one in the docs, can you show me something (in the docs maybe) that proves this field exists? Thanks!

svanharmelen (changes requested):

projects.go
wind0r: Dropped it. This should be done via `build_git_strategy`
`build_allow_git_fetch` is used by the webinterface.

PR #1559, opened 2022-09-28
Add enforce_auth_checks_on_uploads field

epropulsion-tech:


projects.go
```diff
@@ -820,6 +820,7 @@ type EditProjectOptions struct {
 	DefaultBranch                             *string                              `url:"default_branch,omitempty" json:"default_branch,omitempty"`
 	Description                               *string                              `url:"description,omitempty" json:"description,omitempty"`
 	EmailsDisabled                            *bool                                `url:"emails_disabled,omitempty" json:"emails_disabled,omitempty"`
+	EnforceAuthChecksOnUploads                *bool                                `url:"enforce_auth_checks_on_uploads,omitempty" json:"enforce_auth_checks_on_uploads,omitempty"`
```
svanharmelen, line 823: Only adding it here is not enough. It should also be a field in the `Projects` struct and it's also a valid option that you can set when creating a project. So if you can add it there as well that would be great!

svanharmelen (changes requested):

svanharmelen (approved):
Added the missing fields...

PR #1560, opened 2022-09-30
Implement API to disable two factor authentication for user

hjweddie:
implement api for [Disable two factor](https://docs.gitlab.com/ee/api/users.html#delete-authentication-identity-from-user)


svanharmelen (approved):
Thanks 👍🏻

PR #1561, opened 2022-10-07
Add specific type for issue relations

tobischo:
Fields selected based on described fields/examples from docs

Added `references` even though it is not in the docs, as gitlab does return that field

This picks up where #1386 previously left off


svanharmelen (approved):
LGTM! Thanks!

PR #1562, opened 2022-10-09
Align query options for listing issues

jamietanna:
As per 15.5 GitLab documentation, we can make sure that all the queries
include the relevant options for listing issues.


svanharmelen (approved):
Thanks! LGTM 👍🏻

PR #1563, opened 2022-10-10
CheckResponse errors include any non-JSON data

rosenhouse:
My GitLab is throwing errors where the body is not JSON, and using this library I'm finding that the error message coming back from `CheckResponse` wasn't very helpful, just `failed to parse unknown error format`.

This PR will include the original response body in the Go error, when that body cannot be parsed as JSON.

I hope you're amenable to this change.  Thanks for maintaining such a useful project!


svanharmelen (approved):
Looks good to me 👍🏻  Thanks!

PR #1565, opened 2022-10-19
 Add `ref` to deployment webhook 

vntw:
Hey 👋,

`ref` seems to be undocumented in the [deployment webhook event docs](https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html#deployment-events), but looking at the code, it's always present: https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/gitlab/data_builder/deployment.rb#L44. 

Just as another example, deploying from sth. else than a tag will contain sth. like `"ref":"refs/merge-requests/1/merge"`. I tested this with the GitLab SaaS version.

I also updated the docs URL for the job event since the TODO comment was from 2019 🙂

### DIFF
--- a/event_webhook_types.go
+++ b/event_webhook_types.go
@@ -151,6 +151,7 @@ type DeploymentEvent struct {
 		SSHURL            string  `json:"ssh_url"`
 		HTTPURL           string  `json:"http_url"`
 	} `json:"project"`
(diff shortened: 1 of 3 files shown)


svanharmelen (approved):
Thanks 👍🏻  LGTM!

PR #1575, opened 2022-11-15
Add ConfirmedAt to Email

mycrEEpy:
The `confirmed_at` field is currently missing in the `Email` struct.

See https://docs.gitlab.com/ce/api/users.html#list-emails


svanharmelen (approved):
Thanks 👍🏻

PR #1576, opened 2022-11-15
Add ID parameter to FeatureFlagStrategyOptions

r8ygun:
As per the [Project Feature Flag documentation](https://docs.gitlab.com/ee/api/feature_flags.html#update-a-feature-flag) it can be expected to be able to pass an existing `ID` parameter as part of the `strategies` struct when updating a Project Feature Flag.

This change simply adds the ability to do this. This will be used to enable the terraform provider to update feature flag strategies instead of creating them anew.

Related to #1526.


r8ygun:
@svanharmelen May I ask when the next release will be?

svanharmelen:
Just released `v0.75.0` 👍🏻

PR #1579, opened 2022-11-22
Treat ExpiresAt as optional when editing member options

jelsdon:
This patch allows allows changing access levels without existing knowledge of the value and inadvertently removing the account's expiry. This currently occurs as non defined ExpiresAt renders to `"expires_at": null`

`expires_at` is a non-required field https://docs.gitlab.com/ee/api/members.html#edit-a-member-of-a-group-or-project


svanharmelen:
Thanks @jelsdon 👍🏻

Issue #1580, opened 2022-11-24
ListGroupsOptions bug (SkipGroups)


larscom:
Hi,

There seems to be a bug in `ListGroupsOptions` with the `SkipGroups` property.

When i provide `[37,38,39]` as value to `SkipGroups` it will only skip the group with Id: `39` (the last one in the array)

I expect it to skip all groupIds in the array, not only the last one.

![Screenshot_20221124_120731](https://user-images.githubusercontent.com/1215195/203770207-5eb24a26-a086-49e1-8265-9af1f60b0d7a.png)

svanharmelen:
Does this work with a cURL command? As I would almost think this is an API thingy...

larscom:
> Does this work with a cURL command? As I would almost think this is an API thingy...

Yes with curl it does work as expected.

This curl commands skips groups 37,38,39

```shell
curl https://gitlab.com/api/v4/groups?page=1&per_page=100&skip_groups=37,38,39
```

larscom:
![Screenshot_20221126_130711](https://user-images.githubusercontent.com/1215195/204088082-3270ee5b-2bab-4474-8eb2-647098052879.png)

The raw query is constructed wrong it seems like, it should be a comma seperated string.

See: https://github.com/xanzy/go-gitlab/blob/12d29bcd03d59321d80f17ac36896ad89de9525d/gitlab.go#L588

2785:
this is google/go-querystring, it looks like https://go.dev/play/p/SPdLfOD-M7l

2785:
```go
type Thing struct {
	IDs []int `url:"ids" del:","`
}
```

looks like the thing takes a delimiter tag. This prints `ids=1%2C2%2C3` as opposed to `ids=1&ids=2&ids=3`. 

I'm not sure if gitlab API appreciates the path escaped commas, though

larscom:
@2785 it looks like gitlab **does** accept it

This works: `/api/v4/groups?page=1&per_page=100&skip_groups=37%2C38%2C39`

2785:
That's probably our solution then. I can PR unless you want to.

larscom:
Thanks

> That's probably our solution then. I can PR unless you want to.

👍

PR #1581, opened 2022-11-26
fix: add delimiter tag into list group options skip_groups query param

2785:
resolves https://github.com/xanzy/go-gitlab/issues/1580


svanharmelen (approved):
Nice one! Thanks 👍🏻

Issue #1588, opened 2022-12-02
Add support for the IP Restriction Ranges added in 15.4


PatrickRice-KSC:
Required for terraform provider: https://gitlab.com/gitlab-org/terraform-provider-gitlab/-/issues/1332

PR #1589, opened 2022-12-02
Add support for group ip restriction ranges

PatrickRice-KSC:
Closes: #1588

### DIFF
--- a/groups.go
+++ b/groups.go
@@ -84,6 +84,7 @@ type Group struct {
 	PreventForkingOutsideGroup     bool             `json:"prevent_forking_outside_group"`
 	MarkedForDeletionOn            *ISOTime         `json:"marked_for_deletion_on"`
 	CreatedAt                      *time.Time       `json:"created_at"`
+	IPRestrictionRanges            string           `json:"ip_restriction_ranges"`
 }
 
 // GroupAvatar represents a GitLab group avatar.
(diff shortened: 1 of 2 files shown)


PatrickRice-KSC:
@svanharmelen  - I think this is all set for your review!

groups.go
```diff
@@ -84,6 +84,7 @@ type Group struct {
 	PreventForkingOutsideGroup     bool             `json:"prevent_forking_outside_group"`
 	MarkedForDeletionOn            *ISOTime         `json:"marked_for_deletion_on"`
 	CreatedAt                      *time.Time       `json:"created_at"`
+	IPRestrictionRanges            string           `json:"ip_restriction_ranges"`
```
svanharmelen, line 87: I think this would deserve a custom type like `Labels`, so something like `type IPRestrictionRanges []string` with a custom unmarshal func that parses the comma separated list into a `[]string`

But for now this will do as well...

svanharmelen (approved):

groups.go
PatrickRice-KSC: @svanharmelen  - I'm happy to implement something like that, but is there a way to do that for just the IPRestrictionRanges attribute? I see how to create a custom unmarshaller/marshaller for the whole object, but I'm a bit loathe to do that for the whole `Group` object given how big it is (and how often it changes compared to other objects 😞 )

svanharmelen: No agree it should not be on the whole `Group` struct! I think this is a good example of what I mean: https://github.com/xanzy/go-gitlab/blob/master/issues.go#L169-L193

PatrickRice-KSC: @svanharmelen  - Do you think it would make sense to create a single "CommaSeparatedString" type that we could use for other circumstances like this? I'd be happy to implement that.

svanharmelen: While the name is a bit long 😉 I guess it makes sense if we have more then one field that is needs it. Do you know of any other fields that use this same format?

PatrickRice-KSC: @svanharmelen  - Hm, there are at least 3 fields that use this syntax that I can find easily (2 of them are already implemented in `go-gitlab` with their own unique types). Seems like the primary reason would be to be future compatible. I'll leave it as-is for now, and we can re-visit later if we want to.

PatrickRice-KSC: Also, would you mind tagging the repo so we can pull the new version into the provider? We appreciate your help!

svanharmelen: Sounds like its worth the efforts then...

And I'll create a new release this morning 👍🏻

PR #1590, opened 2022-12-02
Add missing mr_default_target_self edit project attribute

timofurrer:
It's an option for forks, see [the docs](https://docs.gitlab.com/ee/api/projects.html#edit-project:~:text=in%20a%20project.-,mr_default_target_self,project.%20If%20false%2C%20the%20target%20will%20be%20the%20upstream%20project.,-name)


svanharmelen (approved):
Thanks! LGTM 👍🏻

PR #1593, opened 2022-12-09
Add WithErrorHandler client option

jokly:
Added the ability to change the error handler for `retryablehttp.Client`

### DIFF
--- a/client_options.go
+++ b/client_options.go
@@ -94,6 +94,14 @@ func WithCustomRetryWaitMinMax(waitMin, waitMax time.Duration) ClientOptionFunc 
 	}
 }
 
+// WithErrorHandler can be used to configure a custom error handler.
+func WithErrorHandler(handler retryablehttp.ErrorHandler) ClientOptionFunc {
+	return func(c *Client) error {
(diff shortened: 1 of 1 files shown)


client_options.go
```diff
@@ -118,6 +118,14 @@ func WithResponseLogHook(hook retryablehttp.ResponseLogHook) ClientOptionFunc {
 	}
 }
 
+// WithErrorHandler can be used to configure a custom error handler.
+func WithErrorHandler(handler retryablehttp.ErrorHandler) ClientOptionFunc {
+	return func(c *Client) error {
+		c.client.ErrorHandler = handler
+		return nil
+	}
+}
```
svanharmelen, line 127: These options are ordered alphabetically, so if you can update the location of the new option that would be great...

svanharmelen (changes requested):
One small request, after that is updated I'll merge this one 👍🏻

client_options.go
jokly: Thanks for the quick response! Fixed this :)

svanharmelen (approved):
Thanks 👍🏻

PR #1596, opened 2022-12-11
New GetLatestPipeline() client method

armsnyder:
Added a client method for https://docs.gitlab.com/ee/api/pipelines.html#get-the-latest-pipeline

### DIFF
--- a/pipelines.go
+++ b/pipelines.go
@@ -264,6 +264,37 @@ func (s *PipelinesService) GetPipelineTestReport(pid interface{}, pipeline int, 
 	return p, resp, err
 }
 
+// GetLatestPipelineOptions represents the available GetLatestPipeline() options.
+//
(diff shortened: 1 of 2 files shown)


pipelines.go
```diff
@@ -216,6 +216,37 @@ func (s *PipelinesService) GetPipeline(pid interface{}, pipeline int, options ..
...
+	}
+
+	p := new(Pipeline)
+	resp, err := s.client.Do(req, p)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return p, resp, err
+}
```
svanharmelen, line 248: Code looks fine, but please place it in the correct location (following the docs, so below `GetPipelineTestReport`)...

armsnyder: Done

svanharmelen (approved):
Thanks 👍🏻

PR #1598, opened 2022-12-11
Fix slow tests caused by retry backoff

armsnyder:
I disabled retry backoffs in tests to reduce the overall test run time from 25 seconds to under 1 second.

Before:

```shell
% ~/go/bin/gotestsum --jsonfile out.json
✓  . (25.685s)
% ~/go/bin/gotestsum tool slowest --jsonfile out.json
github.com/xanzy/go-gitlab TestInstanceClustersService_AddCluster_StatusInternalServerError 12.08s
github.com/xanzy/go-gitlab TestInstanceVariablesService_StatusInternalServerError 11.73s
github.com/xanzy/go-gitlab TestUploadFile_Retry 710ms
github.com/xanzy/go-gitlab TestUploadAvatar_Retry 710ms
```

After:

```shell
% ~/go/bin/gotestsum --jsonfile out.json
✓  . (425ms)
% ~/go/bin/gotestsum tool slowest --jsonfile out.json
```

### DIFF
--- a/gitlab_test.go
+++ b/gitlab_test.go
@@ -27,6 +27,7 @@ import (
 	"os"
 	"strings"
 	"testing"
+	"time"
 
 	retryablehttp "github.com/hashicorp/go-retryablehttp"
 )
@@ -44,7 +45,13 @@ func setup(t *testing.T) (*http.ServeMux, *httptest.Server, *Client) {
 	server := httptest.NewServer(mux)
 
 	// client is the Gitlab client being tested.
-	client, err := NewClient("", WithBaseURL(server.URL))
+	client, err := NewClient("",
+		WithBaseURL(server.URL),
+		// Disable backoff to speed up tests that expect errors.
+		WithCustomBackoff(func(_, _ time.Duration, _ int, _ *http.Response) time.Duration {
+			return 0
+		}),
+	)
 	if err != nil {
 		server.Close()
 		t.Fatalf("Failed to create client: %v", err)


gitlab_test.go
```diff
@@ -44,7 +45,13 @@ func setup(t *testing.T) (*http.ServeMux, *httptest.Server, *Client) {
 	server := httptest.NewServer(mux)
 
 	// client is the Gitlab client being tested.
-	client, err := NewClient("", WithBaseURL(server.URL))
+	client, err := NewClient("",
+		WithBaseURL(server.URL),
+		// Disable backoff to speed up tests that expect errors.
+		WithCustomBackoff(func(_, _ time.Duration, _ int, _ *http.Response) time.Duration {
+			return 0
```
svanharmelen, line 52: Maybe do `1 * time.Second` or `500 * time.Millisecond` so it at least is measurable/noticeable just in case anyone needs or expects that?

svanharmelen (changes requested):

gitlab_test.go
armsnyder: I would prefer not to add delay without a counterexample. I didn't see any tests relying on timing. The retry tests that we have rely on the number of times the handler is called. But, if you're certain, then I'd suggest we set the slow tests to be disabled by default, since it's an annoyance when you're doing TDD.

svanharmelen (approved):
I've taken another look at the involved tests and guess you are right... So I'll take it... 👍🏻

PR #1599, opened 2022-12-11
Deprecate the test teardown function

armsnyder:
See https://pkg.go.dev/testing#T.Cleanup

> Cleanup registers a function to be called when the test (or subtest) and all its subtests complete. Cleanup functions will be called in last added, first called order.

### DIFF
--- a/access_requests_test.go
+++ b/access_requests_test.go
@@ -27,8 +27,7 @@ import (
 )
 
 func TestListProjectAccessRequests(t *testing.T) {
-	mux, server, client := setup(t)
-	defer teardown(server)
+	mux, client := setup(t)
 
 	mux.HandleFunc("/api/v4/projects/1/access_requests", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, http.MethodGet)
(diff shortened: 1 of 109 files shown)


svanharmelen (changes requested):
Thanks for this PR, but I'm not going to merge something like this unless the PR "fixes" it for the whole package (to preserve consistency and maintainability). Not asking you to do that, just saying I'm not open to introduce a new pattern without cleaning up the old one...

armsnyder:
Reasonable. I didn't want to "fix" the whole project unless you were okay with the change first. I may do it then.

svanharmelen:
Check... Additionally I'm not sure if having one cleanup function that will teardown once is going to work for all tests, but I guess that is something we should, uuh test (no pun intended)...

armsnyder:
Updated. It was a very simple find-and-replace with no corner cases. Every time `setup()` was called was at the start of a test and immediately followed by `defer teardown()`.

svanharmelen (approved):
Ok... Then I'll take it :) Thanks!

PR #1600, opened 2022-12-12
Add more "changes" fields to IssueEvent and MergeEvent

armsnyder:
- Add "closed_at" to IssueEvent
- Add "state_id" to IssueEvent
- Add "updated_at" to IssueEvent
- Add "updated_at" to MergeEvent
- Modify "state_id" type from int to StateID in MergeEvent

I modified the testdata so that I could test the fields. These new fields came from webhook payloads that I observed on GitLab 15.3.4-ee.

### DIFF
--- a/event_webhook_types.go
+++ b/event_webhook_types.go
@@ -25,7 +25,8 @@ import (
 
 // StateID identifies the state of an issue or merge request.
 //
-// There are no GitLab API docs on the subject, but the mappings can be found in GitLab's codebase:
+// There are no GitLab API docs on the subject, but the mappings can be found in
+// GitLab's codebase:
 // https://gitlab.com/gitlab-org/gitlab-foss/-/blob/ba5be4989e/app/models/concerns/issuable.rb#L39-42
 type StateID int
 
--- a/event_webhook_types.go
+++ b/event_webhook_types.go
@@ -23,6 +23,20 @@ import (
 	"time"
 )
 
+// StateID identifies the state of an issue or merge request.
+//
+// There are no GitLab API docs on the subject, but the mappings can be found in GitLab's codebase:
+// https://gitlab.com/gitlab-org/gitlab-foss/-/blob/ba5be4989e/app/models/concerns/issuable.rb#L39-42
+type StateID int
+
+const (
+	StateIDNone   StateID = 0
+	StateIDOpen   StateID = 1
+	StateIDClosed StateID = 2
+	StateIDMerged StateID = 3
+	StateIDLocked StateID = 4
+)
+
 // BuildEvent represents a build event.
 //
 // GitLab API docs:
@@ -292,6 +306,18 @@ type IssueEvent struct {
 			Previous string `json:"previous"`
 			Current  string `json:"current"`
 		} `json:"title"`
+		ClosedAt struct {
+			Previous string `json:"previous"`
+			Current  string `json:"current"`
+		} `json:"closed_at"`
+		StateID struct {
+			Previous StateID `json:"previous"`
+			Current  StateID `json:"current"`
+		} `json:"state_id"`
+		UpdatedAt struct {
+			Previous string `json:"previous"`
+			Current  string `json:"current"`
+		} `json:"updated_at"`
 		UpdatedByID struct {
 			Previous int `json:"previous"`
 			Current  int `json:"current"`
@@ -580,8 +606,8 @@ type MergeEvent struct {
 			Current  int `json:"current"`
 		} `json:"source_project_id"`
 		StateID struct {
-			Previous int `json:"previous"`
-			Current  int `json:"current"`
+			Previous StateID `json:"previous"`
+			Current  StateID `json:"current"`
 		} `json:"state_id"`
 		TargetBranch struct {
 			Previous string `json:"previous"`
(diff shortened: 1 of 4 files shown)


event_webhook_types.go
```diff
@@ -1048,3 +1064,17 @@ type EventLabel struct {
...
+// https://gitlab.com/gitlab-org/gitlab-foss/-/blob/ba5be4989e/app/models/concerns/issuable.rb#L39-42
+type StateID int
+
+const (
+	StateIDNone   StateID = 0
+	StateIDOpen   StateID = 1
+	StateIDClosed StateID = 2
+	StateIDMerged StateID = 3
+	StateIDLocked StateID = 4
+)
```
svanharmelen, line 1080: Can you put these (`StateID` and the consts) at the top of the file? Thanks!

event_webhook_types.go
```diff
@@ -292,6 +292,18 @@ type IssueEvent struct {
...
+			Current  string `json:"current"`
+		} `json:"closed_at"`
+		StateID struct {
+			Previous StateID `json:"previous"`
+			Current  StateID `json:"current"`
+		} `json:"state_id"`
+		UpdatedAt struct {
+			Previous string `json:"previous"`
+			Current  string `json:"current"`
+		} `json:"updated_at"`
```
svanharmelen, line 306: Are `ClosedAt` and `UpdatedAt` not strings that should be parsed as timestamps?

svanharmelen (changes requested):

event_webhook_types.go
armsnyder: Done

event_webhook_types.go
svanharmelen: 1. We already have a custom type (`ISOTime`) for a specific time format used by Gitlab, I don't see why we couldn't make another one for the format used here?
2. While it differs from type to type it is consistent for a specific field.
3. How do you loose precision? You can parse the strings GitLab uses here without losing any precision in a time.Time just fine...

armsnyder: I'm happy to make the changes you suggest, but there are a few things I should clarify first:

1. Fair point. However `ISOTime` uses `2006-01-02` as its format, which is a date, so we would need a new type for datetime.
2. My overall point on this is that I am concerned that we are taking on more responsibility than we want to as the GitLab API is not well defined in this area. However I also need to correct what you said – if you compare the documented event payloads to actual payloads, the formats differ on the specific field. I assumed that this meant there was a breaking change in the past, but it could simply be poor documentation.

   [Documented](https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html#issue-events):

   ```json
   "labels": [{
     "id": 206,
     "title": "API",
     "color": "#ffffff",
     "project_id": 14,
     "created_at": "2013-12-03T17:15:43Z",
     "updated_at": "2013-12-03T17:15:43Z",
     "template": false,
     "description": "API related issues",
     "type": "ProjectLabel",
     "group_id": 41
   }]
   ```

   Actual:

   ```json
   "labels": [
     {
       "id": 11199,
       "title": "status::Done",
       "color": "#00b140",
       "project_id": null,
       "created_at": "2022-11-07 17:34:03 UTC",
       "updated_at": "2022-11-07 17:34:03 UTC",
       "template": false,
       "description": "",
       "type": "GroupLabel",
       "group_id": 2168
     }
   ]
   ```

   There is an open issue about time format inconsistencies generally, and they have not decided on a direction yet. https://gitlab.com/gitlab-org/gitlab/-/issues/19667
3. When I say you lose precision I mean you cannot determine the precision off of the `time.Time` object once it has been parsed. You do not know if it is a date or a datetime.

svanharmelen: 1. Yes, I already mentioned that as well. But we could of course (which I also mentioned) create another one specifically for this format.
2. I do relate to this one a little, but I still doubt if by parsing the value this would apply here.
3. Don't agree with this one, `time.Time` holds all the parsed info and doesn't truncate any info it received. So you are still able to get the date and the time from `time.Time` once you parsed it.

event_webhook_types.go
```diff
@@ -23,6 +23,20 @@ import (
 	"time"
 )
 
+// StateID identifies the state of an issue or merge request.
+//
+// There are no GitLab API docs on the subject, but the mappings can be found in GitLab's codebase:
```
svanharmelen, line 28:
```suggestion
// There are no GitLab API docs on the subject, but the mappings can be found in
// GitLab's codebase:
```

svanharmelen (changes requested):
I've got one more small request (I like to keep comments to 80 chars when possible, which of course isn't true for the long URL)...

As for our other discussion I OK with leaving at this for now.

event_webhook_types.go
armsnyder: Good on 1 and 2 but we still have a miscommunication on 3. 😅 Maybe I shouldn't have used the term "lose precision", because I don't mean to say that any data is truncated. You're correct there. I'll try to explain again, even though it sounds like we're converging on using `time.Time` anyway.

As a user if I see the string "2013-12-03" I know that I have a date that has precision down to the day and no more precise. But if I have a `time.Time` with the same value, I do not know whether it is precise down to the day or hour or second. I would have to know the original value that it was parsed from. So that's the information that is missing.

svanharmelen (approved):
Thanks @armsnyder 👍🏻

armsnyder:
@svanharmelen Hi, just so you know I hadn't committed the `time.Time` field changes you had requested.

svanharmelen:
I know, all good for now... 👍🏻

PR #1602, opened 2022-12-14
Update BillableMembers with current API specification

Nighthawk22:
Check https://docs.gitlab.com/ee/api/members.html#list-all-billable-members-of-a-group for current specification.


svanharmelen (approved):
Thanks @Nighthawk22! LGTM 👍🏻

PR #1604, opened 2022-12-22
Fix .gitignore template and GitLab CI YAML template list items

FantasyTeddy:
The two API endpoints for listing `.gitignore` templates and GitLab CI YAML templates used the same `struct` as the corresponding endpoint to retrieve a single template.
However, the two response types are not identical and the `Content` field was always empty when querying the entire list.
I added a new type that properly represents the response of a list item and is different from the type returned by a single template request.


svanharmelen (approved):
Looks good, thanks 👍🏻

PR #1606, opened 2022-12-22
Update the broadcast messages API

FantasyTeddy:
Mark the `Color` field as deprecated ([removed in GitLab 15.6](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/95829))

Add the following new fields according to the [API documentation](https://docs.gitlab.com/ee/api/broadcast_messages.html):
- `TargetAccessLevels`: Target access levels (roles) of the broadcast message.
- `TargetPath`: Target path of the broadcast message.
- `BroadcastType`: Appearance type (defaults to banner)
- `Dismissable`: Can the user dismiss the message?


svanharmelen (approved):
And another one... Thanks @FantasyTeddy LGTM 👍🏻

PR #1609, opened 2022-12-23
Update dependencies

FantasyTeddy:
I hope I did this right, since I'm not too familiar with Go's dependency management.
If I did something wrong, I hope I can learn from my mistakes.


svanharmelen (approved):
Think one new dep is released, But I'll add that one 👍🏻

PR #1611, opened 2022-12-24
Fix GitLab documentation links

FantasyTeddy:
All URLs to https://docs.gitlab.com/ce have been replaced by links to https://docs.gitlab.com/ee, since they redirected there anyways.

Furthermore, there were a few broken documentation links and several broken anchor links.


svanharmelen (approved):
You are on fire 🔥 Thanks!

PR #1615, opened 2023-01-02
Add "raw" option to group/project variable managment

cyd01:
Corresponding to the documentation:

- https://docs.gitlab.com/ee/api/group_level_variables.html
- https://docs.gitlab.com/ee/api/project_level_variables.html

there is now a `raw` setting in API variable definition to determine whether the variable is expandable.

### DIFF
--- a/group_variables.go
+++ b/group_variables.go
@@ -41,8 +41,8 @@ type GroupVariable struct {
 	VariableType     VariableTypeValue `json:"variable_type"`
 	Protected        bool              `json:"protected"`
 	Masked           bool              `json:"masked"`
-	EnvironmentScope string            `json:"environment_scope"`
 	Raw              bool              `json:"raw"`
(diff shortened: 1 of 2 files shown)


group_variables.go
```diff
@@ -42,6 +42,7 @@ type GroupVariable struct {
 	Protected        bool              `json:"protected"`
 	Masked           bool              `json:"masked"`
 	EnvironmentScope string            `json:"environment_scope"`
+	Raw              bool              `json:"raw"`
```
svanharmelen, line 45: Following the ordering from the docs... So can you please place `Raw` in between `Masked` and `EnvironmentScope` everywhere?

svanharmelen (changes requested):

svanharmelen (approved):
Thanks! 👍🏻

PR #1619, opened 2023-01-06
Add requestOptions to Client

htdvisser:
This pull request adds a `WithRequestOptions` `ClientOptionFunc` that lets you configure the client to apply request options to every request to the GitLab API.

With this client option, I don't have to keep adding the same request options every time I want to make an API call.

### DIFF
--- a/client_options.go
+++ b/client_options.go
@@ -134,10 +134,10 @@ func WithoutRetries() ClientOptionFunc {
 	}
 }
 
-// WithRequestOptions can be used to configure options that are applied to every request to the GitLab API.
+// WithRequestOptions can be used to configure default request options applied to every request.
 func WithRequestOptions(options ...RequestOptionFunc) ClientOptionFunc {
 	return func(c *Client) error {
-		c.requestOptions = append(c.requestOptions, options...)
+		c.defaultRequestOptions = append(c.defaultRequestOptions, options...)
 		return nil
 	}
 }
--- a/client_options.go
+++ b/client_options.go
@@ -133,3 +133,11 @@ func WithoutRetries() ClientOptionFunc {
 		return nil
 	}
 }
+
+// WithRequestOptions can be used to configure options that are applied to every request to the GitLab API.
+func WithRequestOptions(options ...RequestOptionFunc) ClientOptionFunc {
+	return func(c *Client) error {
+		c.requestOptions = append(c.requestOptions, options...)
+		return nil
+	}
+}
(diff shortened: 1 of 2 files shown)


gitlab.go
```diff
@@ -99,6 +99,9 @@ type Client struct {
 	// User agent used when communicating with the GitLab API.
 	UserAgent string
 
+	// Request options that are applied to every request to the GitLab API.
+	requestOptions []RequestOptionFunc
```
htdvisser, line 103: @svanharmelen do you prefer this field to be exported or unexported?

client_options.go
```diff
@@ -133,3 +133,11 @@ func WithoutRetries() ClientOptionFunc {
 		return nil
 	}
 }
+
+// WithRequestOptions can be used to configure options that are applied to every request to the GitLab API.
```
svanharmelen, line 137:
```suggestion
// WithRequestOptions can be used to configure default request options applied to every request.
```

gitlab.go
```diff
@@ -595,6 +598,14 @@ func (c *Client) NewRequest(method, path string, opt interface{}, options []Requ
...
 	}
 
+	for _, fn := range c.requestOptions {
+		if fn == nil {
+			continue
+		}
+		if err := fn(req); err != nil {
+			return nil, err
+		}
+	}
```
svanharmelen, line 608: Instead  of adding a new block, I prefer to concat the slices:

```go
for _, fn := range append(c.defaultRequestOptions, options...) {
		if fn == nil {
			continue
		}
		if err := fn(req); err != nil {
			return nil, err
		}
	}
```

gitlab.go
svanharmelen, line 103:
```suggestion
	defaultRequestOptions []RequestOptionFunc
```
And please place this one above the `UserAgent`...

svanharmelen (changes requested):
Have a few small suggestions, but I'm OK with the general idea...

client_options.go
htdvisser: Done as requested.

gitlab.go
htdvisser: I would have expected that you'd prefer avoiding the (unnecessary) allocation, but done as requested.

svanharmelen: Fair point... Yet I would hope (and expect a little bit) that the compiler is smart enough to compile this down to effective bytecode 😊

svanharmelen (approved):
Thanks @htdvisser! LGTM 👍🏻

htdvisser:
Thank you!

PR #1621, opened 2023-01-10
fix: Added missing mapping for ProjectID

linkdotnet:
The `RegistryRepository` also contains a `project_id` (see [here](https://docs.gitlab.com/ee/api/container_registry.html#within-a-group)) that wasn't mapped. 

Funnily enough, the tests had those JSON fields.


svanharmelen (approved):
Thanks 👍🏻

PR #1626, opened 2023-01-11
fix: Only use ListOptions

### LINKED ISSUE #1625: Bug: `ListGroupRegistryRepositories` uses `ListRegistryRepositoriesOptions`
The `ListGroupRegistryRepositories` call accepts a `ListRegistryRepositoriesOptions` object as option. The problem is that `Tags` and `TagsCount` is not a valid option for `ListGroupRegistryRepositories`, only for `ListRegistryRepositoryTags`

See https://docs.gitlab.com/ee/api/container_registry.html#within-a-group (It got removed with gitlab 15.0)

> [Removed](https://gitlab.com/gitlab-org/gitlab/-/issues/336912) the tags and tag_count attributes in GitLab 15.0.

The obvious solution would be to only allow `ListOptions` as a valid parameter to indicate that `tag` and `tagCount` has no impact.

EDIT: As the initial idea would make the API incompatible with older gitlab versions, we only deprecate these usage of `ListRegistryRepositoriesOptions`.

linkdotnet:
Fixes #1625

### DIFF
--- a/container_registry.go
+++ b/container_registry.go
@@ -74,7 +74,9 @@ func (s RegistryRepositoryTag) String() string {
 // https://docs.gitlab.com/ee/api/container_registry.html#list-registry-repositories
 type ListRegistryRepositoriesOptions struct {
 	ListOptions
-	Tags      *bool `url:"tags,omitempty" json:"tags,omitempty"`
+	// Deprecated: This option is deprecated only for ListGroupRegistryRepositories calls. (Removed in GitLab 15.0)
+	Tags *bool `url:"tags,omitempty" json:"tags,omitempty"`
+	// Deprecated: This option is deprecated only for ListGroupRegistryRepositories calls. (Removed in GitLab 15.0)
 	TagsCount *bool `url:"tags_count,omitempty" json:"tags_count,omitempty"`
 }
 


svanharmelen:
I understand the idea, but by using this way to fix it you make this package incompatible with older GitLab version. So I prefer to add a deprecation note instead (like this: https://github.com/xanzy/go-gitlab/blob/master/runners.go#L184)

linkdotnet:
> I understand the idea, but by using this way to fix it you make this package incompatible with older GitLab version. So I prefer to add a deprecation note instead (like this: https://github.com/xanzy/go-gitlab/blob/master/runners.go#L184)

That makes absolute sense. Thanks. I'll update the PR soon (and update the issue)

linkdotnet:
@svanharmelen  As the object is shared by two function calls (group level as well as project level, we can only mark the parameter in `ListGroupRegistryRepositories` as deprecated/obsolete. How does that work?

svanharmelen:
It's not pretty, but maybe make a new type `ListGroupRegistryRepositoriesOptions`... This is a bit of a nasty one, as in that case I would also like to rename the existing one to `ListProjectRegistryRepositoriesOptions`. Not sure if its all worth it as in the end both types will still have the same options.

So maybe just add that clarification to the deprecation message? So something like:

```go
// Deprecated: These options are deprecated only for ListGroupRegistryRepositories calls. (Deprecated in GitLab 15.0)
```

linkdotnet:
> It's not pretty, but maybe make a new type `ListGroupRegistryRepositoriesOptions`... This is a bit of a nasty one, as in that case I would also like to rename the existing one to `ListProjectRegistryRepositoriesOptions`. Not sure if its all worth it as in the end both types will still have the same options.
> 
> So maybe just add that clarification to the deprecation message? So something like:
> 
> ```go
> // Deprecated: These options are deprecated only for ListGroupRegistryRepositories calls. (Deprecated in GitLab 15.0)
> ```

Added the deprecated message on top of both fields in the options.

svanharmelen (approved):
Thanks 👍🏻

PR #1634, opened 2023-01-23
Add the ability to retrieve certificate information to Pages Domains

### LINKED ISSUE #1633: Align Pages Domain attributes to upstream GitLab API
The Pages Domain objects were missing `certificate`, `subject`, and `certificate_text` attributes. Oddly, the tests included those attributes, but they weren't exposed in the objects.

This Issue will add support for those items, along with updating the tests to support them properly.

PatrickRice-KSC:
Closes #1633

### DIFF
--- a/pages_domains.go
+++ b/pages_domains.go
@@ -42,8 +42,11 @@ type PagesDomain struct {
 	VerificationCode string     `json:"verification_code"`
 	EnabledUntil     *time.Time `json:"enabled_until"`
 	Certificate      struct {
-		Expired    bool       `json:"expired"`
-		Expiration *time.Time `json:"expiration"`
+		Subject         string     `json:"subject"`
+		Expired         bool       `json:"expired"`
+		Expiration      *time.Time `json:"expiration"`
+		Certificate     string     `json:"certificate"`
+		CertificateText string     `json:"certificate_text"`
 	} `json:"certificate"`
 }
 
--- a/pages_domains_test.go
+++ b/pages_domains_test.go
@@ -40,11 +40,17 @@ func TestPagesDomainsService_ListPagesDomains(t *testing.T) {
 		VerificationCode: "",
 		EnabledUntil:     nil,
 		Certificate: struct {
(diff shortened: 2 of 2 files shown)


PatrickRice-KSC:
@svanharmelen  - I believe this is ready for review if you wouldn't mind!

pages_domains.go
```diff
@@ -42,8 +42,11 @@ type PagesDomain struct {
 	VerificationCode string     `json:"verification_code"`
 	EnabledUntil     *time.Time `json:"enabled_until"`
 	Certificate      struct {
-		Expired    bool       `json:"expired"`
-		Expiration *time.Time `json:"expiration"`
+		Subject         string     `json:"subject"`
+		Expired         bool       `json:"expired"`
+		Expiration      *time.Time `json:"expiration"`
```
timofurrer, line 47: @PatrickRice-KSC is `Expiration` still a valid field? It's not listed in the docs: https://docs.gitlab.com/ee/api/pages_domains.html#single-pages-domain

PatrickRice-KSC: @timofurrer  - Weirdly, there are some other undocumented fields that still show up when I run test cases against a local GitLab (example: verified/verification code)... let me go ruby digging to see what the valid fields are.

PatrickRice-KSC: @timofurrer  - Looks like it is, and there may be another `enabled_until` field here as well. Given that's undocumented, I don't think we should _add_ it in the API here, but I think we should leave expiration alone even though it's undocumented:

- https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/api/entities/pages_domain_basic.rb
- https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/api/entities/pages_domain_certificate_expiration.rb

svanharmelen: In general we like to stick to whatever is in the docs and not add undocumented fields. But as this field is already in I guess this is good to go 👍🏻

svanharmelen (approved):

PR #1635, opened 2023-01-23
Add missing system hook event for user_failed_login

mycrEEpy:
This PR adds the missing system hook event for `user_failed_login` as documented in the [GitLab Docs](https://docs.gitlab.com/ee/administration/system_hooks.html).

Interestingly enough the test data for the event was already there.


svanharmelen (approved):
I'll take it 👍🏻

PR #1636, opened 2023-01-24
Add field import_url to Project

apricote:
Field is available in the API, but missing from the api docs.

https://gitlab.com/gitlab-org/gitlab/-/blob/b93c0cf331413309e44695dac7a3bc63f5bf1b5d/lib/api/entities/project.rb#L98-100

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -99,6 +99,7 @@ type Project struct {
 	MirrorTriggerBuilds                       bool                       `json:"mirror_trigger_builds"`
 	OnlyMirrorProtectedBranches               bool                       `json:"only_mirror_protected_branches"`
 	MirrorOverwritesDivergedBranches          bool                       `json:"mirror_overwrites_diverged_branches"`
+	ImportURL                                 string                     `json:"import_url"`
 	PackagesEnabled                           bool                       `json:"packages_enabled"`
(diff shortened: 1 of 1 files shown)


svanharmelen:
I'm a bit reluctant to support an undocumented field as we do not know if the field is actually part of the public API and might still change before it gets (if at all) added to the docs... So in the past we declined adding undocumented fields and I think I want to stick to that for now.

svanharmelen:
I thought about it a little more, but decided that for now I want to stick to only supporting things that are documented... Sorry!

apricote:
> I thought about it a little more, but decided that for now I want to stick to only supporting things that are documented... Sorry!

Makes sense for the project! No problem :)

timofurrer:
@apricote I went ahead and documented the field https://gitlab.com/gitlab-org/gitlab/-/merge_requests/111933

@svanharmelen any chance to re-open and merge this once the aforementioned MR is merged?

svanharmelen:
Sure thing 👍🏻

timofurrer:
@svanharmelen The MR has been merged. Can we merge and release this?

svanharmelen (approved):

svanharmelen:
Tagged `v0.80.2`

PR #1638, opened 2023-01-30
Revert the change to project member accepting ISO time

### LINKED ISSUE #1637: Revert the change to ProjectMembers API accepting a Time object - Update to accept ISO Time
Relates to #1630

Unfortunately, the update to require a `Time` object, while more precise, causes a breaking change. In addition, GitLab actually discards the additional precision, resulting in a breaking change that doesn't actually provide the extra precision to the end user.

We should update the project members API back to ISO standard so that we don't inject a breaking change. However, we should provide a convenience method for converting a `Time` object to an `ISOTime` object, so that it's easy for projects that are using the library to convert between objects. While there would be some loss of precision inherent in that method, the `ISOTime` object should only be used where that precision is lost anyway.

PatrickRice-KSC:
Closes #1637

This reverts a change from ISOTime -> time.Time type within the ProjectMembers struct. This revert fixes a breaking change, that being that ISOTime can accept YYYY-MM-DD format dates, whereas time.Time requires a full DateTime object. 

With that breaking change in place, any consumer who is relying on the time suffix not being present would break.

### DIFF
--- a/project_members_test.go
+++ b/project_members_test.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"net/http"
 	"testing"
-	"time"
 
 	"github.com/stretchr/testify/require"
 )
@@ -292,13 +291,11 @@ func TestProjectMembersService_EditProjectMember(t *testing.T) {
 			  "web_url": "http://192.168.1.8:3000/root",
 			  "access_level": 30,
 			  "email": "venkatesh.thalluri@example.com",
-			  "expires_at": "2023-10-22T14:13:35Z",
+			  "expires_at": null,
 			  "group_saml_identity": null
 			}
 		`)
 	})
-
-	ExpectedExpiry := time.Date(2023, time.October, 22, 14, 13, 35, 0, time.UTC)
 
 	want := &ProjectMember{
 		ID:          1,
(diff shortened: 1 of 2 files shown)


PatrickRice-KSC:
I think this should be a simple review since it's just a reversion @svanharmelen  😞 

Happy to chat about alternative methods, but there wasn't a lot of chatter on the issue from the original contributor, and I wanted to fix this soon-ish so other consumers don't accidentally break if they use this feature.

svanharmelen:
Thanks @PatrickRice-KSC for the updates and explanations... I'll merge this one and release a new version...

svanharmelen:
Tagged version `v0.79.1` Guess I didn't take enough time to fully understand the change in relation to how the API actually handles this field. Bit annoying that the API handles it like this and that we need two different time types for different parts of the API, but looks like this is the better solution for this one 👍🏻

PatrickRice-KSC:
Yeah, I agree completely. I'm going to see if I can make a documentation update on the GitLab side as well, because a lot of their examples have DateTime objects, even though only the Date precision is used. I suspect that's behind a lot of confusion.

Issue #1583, opened 2022-11-28
Add custom header for passthrough Google IAM


insomnius:
Currently the library won't work if the gitlab is under google IAM, need to add custom additional header to insert IAM token in the request.

theoriginalstove:
I'm wondering if this is as simple as adding a new `RequestOptionFunc` and then when creating a new client just passing it into `WithRequestOptions`.

insomnius:
Yup, just simple custom header could work. But getting the IAM header is another thing, since it's not visible in regular network console.

PR #1639, opened 2023-01-30
feat: add ability to set custom headers for client requests.

theoriginalstove:
For #1583 in order to add a custom header.


theoriginalstove:
Sorry I know the tests for these are a bit messy. I will clean them up a bit. I wasn't sure if it would be better to set a single Header or be able to pass a map  to create multiple at once so I made both methods just in case as a proof-of-concept.

PR #1645, opened 2023-02-07
Add Files at `UpdateSnippetOptions` and `UpdateProjectSnippetOptions `

blacknon:
There was Files in the options of Create, but it seems that it was not in Update, so I added it.


svanharmelen (approved):
Thanks @blacknon 👍🏻

PR #1650, opened 2023-02-10
update SnippetFile.

blacknon:
Sorry again.
When using SnippetFiles when updating, the values ​​`Action` and `PreviousPath` were required, so I added them.

### DIFF
--- a/snippets.go
+++ b/snippets.go
@@ -152,11 +152,11 @@ func (s *SnippetsService) SnippetFileContent(snippet int, ref, filename string, 
 	return b.Bytes(), resp, err
 }
 
-// CreateSnippetFile represents the object that is used to create snippets
+// CreateSnippetFileOptions represents the create snippet file options.
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/snippets.html#create-new-snippet
-type CreateSnippetFile struct {
+type CreateSnippetFileOptions struct {
 	FilePath *string `url:"file_path,omitempty" json:"file_path,omitempty"`
 	Content  *string `url:"content,omitempty" json:"content,omitempty"`
 }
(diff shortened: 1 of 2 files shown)


snippets.go
```diff
@@ -157,8 +157,10 @@ func (s *SnippetsService) SnippetFileContent(snippet int, ref, filename string,
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/snippets.html#create-new-snippet
 type SnippetFile struct {
-	FilePath *string `url:"file_path,omitempty" json:"file_path,omitempty"`
-	Content  *string `url:"content,omitempty" json:"content,omitempty"`
+	FilePath     *string `url:"file_path,omitempty" json:"file_path,omitempty"`
+	Content      *string `url:"content,omitempty" json:"content,omitempty"`
+	Action       *string `json:"action,omitempty"`        // use only update
+	PreviousPath *string `json:"previous_path,omitempty"` // use only update
```
svanharmelen, line 163: Each method that needs different options should use its own options struct... So instead of adding the fields to this struct, maybe we should rename this struct to `CreateSnippetFile` and add an `UpdateSnippetFile` struct?

svanharmelen (changes requested):

snippets.go
blacknon: I checked the parameters of the Update Snippet API and thought it could be handled by the same Strcut. 🤔
The reason is that we thought we could use the `SnippetFile` as it is, since the API json keys to be read as Struct are the same.
However, we were not able to consider the unity of the library as a whole.

> CreateSnippetFile and add an UpdateSnippetFile struct

I would like to implement this modification, if I may?

blacknon:
I fixed it. please check 🙏 .

https://github.com/xanzy/go-gitlab/pull/1650#discussion_r1102998538

svanharmelen (approved):

svanharmelen:
Thanks @blacknon 👍🏻

PR #1660, opened 2023-02-24
Update rate limiting to use a custom or default value.

### LINKED ISSUE #1659: Update rate limiter behavior to use a static value unless explicitly configured
The current behavior for the rate limiter is to make a request to "/", then read the RateLimit header on the 429 response. When users either create a client-per-request or when the client is used in integration tests, this can cause a _lot_ of noise to that endpoint.

I would suggest that we update the behavior slightly:

- Allow a custom value to be passed in during the client creation. This value is used if it's present.
- If no value is passed in, use a default value.
- Add a flag that restores the current behavior if explicitly set to `true`

cc/ @timofurrer

PatrickRice-KSC:
Closes #1659

I also refactored the rate limiter creation slightly to make it more testable. The logic of determining the limiter is now extracted out into its own func which allows us to easily test the output. I've added tests for dynamically determining the value, for statically setting the value, and for using the default behavior.

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -84,6 +84,14 @@ type Client struct {
 	// Limiter is used to limit API calls and prevent 429 responses.
 	limiter RateLimiter
 
+	// Flag to determine if dynamic rate limiting should be enabled. If enabled,
+	// rate limits will be determined by headers instead of using a default value of 1,000.
+	enableDynamicRateLimiting *bool
+
+	// Value that allows configuring the rate limit when creating the client. If this flag
+	// is set, `enableDynamicRateLimiting` will be ignored.
+	staticRateLimit *int
+
 	// Token type used to make authenticated API calls.
 	authType AuthType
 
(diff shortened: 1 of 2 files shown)


svanharmelen:
Hi @PatrickRice-KSC... I'm not too excited about this PR to be honest. The rate limiter code had been in here for quite a while and has proven to be stable and working well for >99% of the people using it (based on the fact I never get any issues about it).

And for the <1% of people where it doesn't play well with their setup, there is always the [option](https://github.com/xanzy/go-gitlab/blob/master/client_options.go#L54) to pass in a custom rate limiter, configured exactly to your needs.

So I would suggest you have a look at that approach as that should give you all the freedom you need to tweak things to your liking without the need for this PR.

timofurrer:
The problem here is not in particular our client, but the sum of all clients - they make millions of requests a day to e.g. `gitlab.com//api/v4` just to retrieve these headers (I'd argue that most of the clients will never hit any rate limiting anyways). While this may work on some instances, for potentially many it doesn't work, because there is CloudFlare in front which enforces its own rate limiting. 

This change would be a mitigation for these millions of requests.

PatrickRice-KSC:
I'll also add that I set the default based on gitlab.com's published rate limits for authenticated users of 2,000 requests/minute. The 30 rps that I chose puts a given user at 1,800 rpm, which is within that threshold with no change to end-users of the library. 

I agree that change for change's sake isn't good, but hitting the endpoint millions of times per day as Timo mentioned seems like something we may want to mitigate 😄

horjulf:
Hi @svanharmelen, I've added some metrics/graphs at https://gitlab.com/gitlab-org/terraform-provider-gitlab/-/issues/1360#note_1296021797 to highlight the impact.

The strategy becomes problematic in cases where the library is used as a one-off, either because a new HTTP client is used every time or due to single request workflows. We effectively get double the request rate for these cases.

A default rate/pacer would be ideal to protect instances of all sizes, alternatively would it make sense to start with no limit and set one as best effort after the first request ?

svanharmelen:
Hi @horjulf, sorry for my slow response (I started my own business a few months ago so I'm kind of busy 😉)! I was indeed thinking something along the lines of your last suggestion, so make the initial call without a configured limiter and use the response of the initial call to configure the limiter for any additional calls.

I have an idea how that could fit in nicely in this package, but need to find some time to add and test it. Will probably be this Friday or else this weekend. So keep you posted in a few days...

PatrickRice-KSC:
@svanharmelen  - I can try and take a crack at updating this PR to do that as well; congrats on the new business btw!

svanharmelen:
Thanks @PatrickRice-KSC but I found some time this morning to implement a simple solution that should solve the issue in a backwards compatible way that doesn't impact anything else while still preventing the limiter from needing that additional call.

Can you (@PatrickRice-KSC, @timofurrer and @horjulf) please check out #1663 and let me know (confirm) if that is a solution to the problem? Thanks!

PR #1661, opened 2023-02-26
Add missing project properties

KingCrunch:
I tried the gitlab gitlab-provider for terraform and realized, that there were some fields missing.

I compared the current fields with the api document, but maybe I missed some.


svanharmelen (approved):
Thanks @KingCrunch 👍🏻

PR #1663, opened 2023-03-08
Prevent additional call to config the rate limiter

### LINKED ISSUE #1659: Update rate limiter behavior to use a static value unless explicitly configured

svanharmelen:
Fixes #1659
Closes #1660


timofurrer:
LGTM ! Thanks ❤️

timofurrer (approved):

PatrickRice-KSC:
Looks great to me as well, thanks @svanharmelen !

PR #1665, opened 2023-03-08
Add omitempty and test

### LINKED ISSUE #1664: New Project Value causes issue when null
In #1661 , a bunch of new properties were added to the `Project` object, which is great! Unfortunately, one of the tags (`only_allow_merge_if_all_status_checks_passed`) is missing the `omitempty` json tag, which causes an API error if the value is set to `nil`, which is allowed due to it accepting a pointer.

PatrickRice-KSC:
Closes #1664 

This updates one optional project attribute to use `omitempty`, and creates one test to ensure that marshalling the object with that attribute not set doesn't include that tag anymore :)


timofurrer (approved):
Thank you very much @PatrickRice-KSC !

svanharmelen (approved):
Thanks @PatrickRice-KSC 👍🏻

PR #1667, opened 2023-03-10
Fix struct for NoteMr event

underside:
I have found that struct for NoteMr Event is not correspond to Gitlab documentation, since I replaced this stuff for my needs I propose to merge this in go-gitlab. 

I hope it will be helpful.


event_webhook_types.go
```diff
@@ -402,102 +402,127 @@ type MemberEvent struct {
...
-		PathWithNamespace string          `json:"path_with_namespace"`
-		DefaultBranch     string          `json:"default_branch"`
-		Homepage          string          `json:"homepage"`
-		URL               string          `json:"url"`
-		SSHURL            string          `json:"ssh_url"`
-		HTTPURL           string          `json:"http_url"`
-		WebURL            string          `json:"web_url"`
-		Visibility        VisibilityValue `json:"visibility"`
+	ObjectKind string `json:"object_kind"`
+	User       struct {
```
svanharmelen, line 406: Please revert this to use `*EventUser` (it has the same fields)

event_webhook_types.go
```diff
@@ -402,102 +402,127 @@ type MemberEvent struct {
...
+		CreatedAt    string      `json:"created_at"`
+		UpdatedAt    string      `json:"updated_at"`
+		ProjectID    int         `json:"project_id"`
+		Attachment   interface{} `json:"attachment"`
+		LineCode     interface{} `json:"line_code"`
+		CommitID     string      `json:"commit_id"`
+		NoteableID   int         `json:"noteable_id"`
+		System       bool        `json:"system"`
+		StDiff       interface{} `json:"st_diff"`
+		URL          string      `json:"url"`
```
svanharmelen, line 445: Are you sure all the `ObjectAttributes` fields you removed are incorrect? I cannot imagine these fields ever being added without them being in the docs.

event_webhook_types.go
```diff
@@ -402,102 +402,127 @@ type MemberEvent struct {
...
+		AuthorID        int    `json:"author_id"`
+		AssigneeID      int    `json:"assignee_id"`
+		Title           string `json:"title"`
+		CreatedAt       string `json:"created_at"`
+		UpdatedAt       string `json:"updated_at"`
+		MilestoneID     int    `json:"milestone_id"`
+		State           string `json:"state"`
+		MergeStatus     string `json:"merge_status"`
+		TargetProjectID int    `json:"target_project_id"`
+		Iid             int    `json:"iid"`
```
svanharmelen, line 462:
```suggestion
		IID             int    `json:"iid"`
```

event_webhook_types.go
```diff
@@ -402,102 +402,127 @@ type MemberEvent struct {
...
+			Color       string      `json:"color"`
+			ProjectID   interface{} `json:"project_id"`
+			CreatedAt   time.Time   `json:"created_at"`
+			UpdatedAt   time.Time   `json:"updated_at"`
+			Template    bool        `json:"template"`
+			Description interface{} `json:"description"`
+			Type        string      `json:"type"`
+			GroupID     int         `json:"group_id"`
+		} `json:"labels"`
+		Source struct {
```
svanharmelen, line 477: Please revert this to use `*Repository` (it has the same fields)

event_webhook_types.go
```diff
@@ -402,102 +402,127 @@ type MemberEvent struct {
...
+			Namespace         string      `json:"namespace"`
+			VisibilityLevel   int         `json:"visibility_level"`
+			PathWithNamespace string      `json:"path_with_namespace"`
+			DefaultBranch     string      `json:"default_branch"`
+			Homepage          string      `json:"homepage"`
+			URL               string      `json:"url"`
+			SSHURL            string      `json:"ssh_url"`
+			HTTPURL           string      `json:"http_url"`
+		} `json:"source"`
+		Target struct {
```
svanharmelen, line 493: Please revert this to use `*Repository` (it has the same fields)

event_webhook_types.go
```diff
@@ -402,102 +402,127 @@ type MemberEvent struct {
...
+			Timestamp string `json:"timestamp"`
+			URL       string `json:"url"`
 			Author    struct {
 				Name  string `json:"name"`
 				Email string `json:"email"`
 			} `json:"author"`
 		} `json:"last_commit"`
 		WorkInProgress bool `json:"work_in_progress"`
-		TotalTimeSpent int  `json:"total_time_spent"`
-		HeadPipelineID int  `json:"head_pipeline_id"`
```
svanharmelen, line 500: Are you sure all these fields are incorrect?

svanharmelen (changes requested):

event_webhook_types.go
underside: Well, I'm not 100% sure about older version of Gitlab, but I just convert latest version of json event to Go struct and add it :-)  Before I did this I have got an error.  
Also this variant already in use and as far as I see, it works flawlessly. 

I think the struct in go-git was correct for previous version, let me check this, I will back with result a bit later.

svanharmelen: Yes, but we would like this package to be compatible with older versions of the GitLab API as well. in some cases we then add a deprecation notice (search the code for examples) to follow the GitLab API, but we try to not remove it to remain backwards compatible with older versions.

svanharmelen:
I'm going to close this PR as it's outdated and most likely "fixed" by a couple of recently merged PRs (the same ones that now cause this PR to have conflicts). If you feel something is still missing, then feel free to open a new PR for that 👍🏻

PR #1668, opened 2023-03-10
Add BuildQueuedDuration to JobEvent struct

catouc:
The field is send by GitLab for a while now, the fixtures also already included it so I'm just properly passing it through. GitLabs docs are just outdated.


svanharmelen:
I generally try to support what is in the docs and omit anything not in the docs to prevent too many changes (in case things change before they are documented)... So I propose to wait with this one until it's documented. Maybe open a MR with GitLab to ask for a doc fix?

catouc:
Added an [MR](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/115163) to GitLab to update this field.

catouc:
@svanharmelen With that upstream MR merged do we have to wait until it's dropped into an actual release?

svanharmelen (approved):
Thanks @catouc 👍🏻

PR #1673, opened 2023-03-15
update gitlab project event struct

chenggui53:
It seems gitlab project event api has upgrade, and the old struct is not correct to now api, so i update the project event struct to match the new api.


oryxBuildBinary
svanharmelen, line 1: Please remove this file from the PR

projects.go
```diff
@@ -572,26 +572,32 @@ func (s *ProjectsService) GetProject(pid interface{}, opt *GetProjectOptions, op
...
-	Data           struct {
-		Before            string      `json:"before"`
-		After             string      `json:"after"`
-		Ref               string      `json:"ref"`
-		UserID            int         `json:"user_id"`
-		UserName          string      `json:"user_name"`
-		Repository        *Repository `json:"repository"`
-		Commits           []*Commit   `json:"commits"`
-		TotalCommitsCount int         `json:"total_commits_count"`
-	} `json:"data"`
```
svanharmelen, line 591: This is still a valid field.

projects.go
```diff
@@ -572,26 +572,32 @@ func (s *ProjectsService) GetProject(pid interface{}, opt *GetProjectOptions, op
...
+	Note           interface{} `json:"note"`
+	PushData       interface{} `json:"push_data"`
```
svanharmelen, line 588: I don't see a field called `push_data` in the docs...

projects.go
```diff
@@ -572,26 +572,32 @@ func (s *ProjectsService) GetProject(pid interface{}, opt *GetProjectOptions, op
...
-		Before            string      `json:"before"`
-		After             string      `json:"after"`
-		Ref               string      `json:"ref"`
-		UserID            int         `json:"user_id"`
-		UserName          string      `json:"user_name"`
-		Repository        *Repository `json:"repository"`
-		Commits           []*Commit   `json:"commits"`
-		TotalCommitsCount int         `json:"total_commits_count"`
-	} `json:"data"`
+	Note           interface{} `json:"note"`
```
svanharmelen, line 587: This needs to be a typed struct. The IDs are `interface{}`'es because they can be either a string or int, but the rest should all be typed.

projects.go
```diff
@@ -572,26 +572,32 @@ func (s *ProjectsService) GetProject(pid interface{}, opt *GetProjectOptions, op
...
 }
 
+type ProjectEventAuthor struct {
+	Name string `json:"name"`
+	Username string `json:"username"`
+	ID int `json:"id"`
+	State string `json:"state"`
+	AvatarUrl string `json:"avatar_url"`
+	WebUrl string `json:"web_url"`
+}
```
svanharmelen, line 599: I thing this matches the `EventUser`?

svanharmelen (changes requested):

svanharmelen:
@chenggui53 can you have a look and check if this is actually what you want?

chenggui53:
> @chenggui53 can you have a look and check if this is actually what you want?

@svanharmelen yes, and i have add push data field and its test from gitlab doc.

svanharmelen (approved):
Thanks @chenggui53 👍🏻

PR #1674, opened 2023-03-15
Draft: Updated Webhook types 

smit-modi:
Added missing fields in DeploymentEvent.

Since GitLab doesn't send time data in RFC3339 format, currently, it is being unmarshalled as a string, I have added `customTime` as https://github.com/go-playground/webhooks/blob/master/gitlab/payload.go#L12 is doing.

For now only updated DeploymentEvent, planning to update other events as well soon.

### DIFF
--- a/event_webhook_types.go
+++ b/event_webhook_types.go
@@ -197,6 +197,23 @@ type DeploymentEvent struct {
 	CommitTitle string     `json:"commit_title"`
 }
 
+// FeatureFlagEvent represents a feature flag event
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html#feature-flag-events
+type FeatureFlagEvent struct {
+	ObjectKind       string     `json:"object_kind"`
+	Project          *Project   `json:"project"` // TODO check which Project struct to use Project or ProjectEvent
+	User             *EventUser `json:"user"`
+	UserURL          string     `json:"user_url"`
+	ObjectAttributes struct {
+		ID          int    `json:"id"`
+		Name        string `json:"name"`
+		Description string `json:"description"`
+		Active      bool   `json:"active"`
+	} `json:"object_attributes"`
+}
+
 // IssueCommentEvent represents a comment on an issue event.
 //
 // GitLab API docs:
--- a/event_webhook_types.go
+++ b/event_webhook_types.go
@@ -671,10 +671,22 @@ type MergeEvent struct {
 	Repository *Repository   `json:"repository"`
 	Labels     []*EventLabel `json:"labels"`
 	Changes    struct {
+		Assignees struct {
+			Previous []*EventUser `json:"previous"`
+			Current  []*EventUser `json:"current"`
+		} `json:"assignees"`
+		Reviewers struct {
+			Previous []*EventUser `json:"previous"`
+			Current  []*EventUser `json:"current"`
+		} `json:"reviewers"`
 		Description struct {
 			Previous string `json:"previous"`
 			Current  string `json:"current"`
 		} `json:"description"`
+		Labels struct {
+			Previous []*EventLabel `json:"previous"`
+			Current  []*EventLabel `json:"current"`
+		} `json:"labels"`
 		LastEditedAt struct {
 			Previous string `json:"previous"`
 			Current  string `json:"current"`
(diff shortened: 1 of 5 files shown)


smit-modi:
@svanharmelen What are your thoughts on changing the fields with time data, from string to customTime type?

svanharmelen:
Hmm... We already have ISOTime as a custom time value, I'm not sure if I want to add more custom time objects.

Yet I do like your idea to try and parse multiple different formats, but that will only help parsing from the `string` to `time.Time`, not the other way around (for queries that take a time).

It's indeed a shame the GitLab API doesn't use a consistent time format, yet until now we seemed to be able to get away with just `ISOTime` and `time.Time`.

Guess I want to think about it a little and have a closer look when back behind a desk again (I'm away for a few days).

event_webhook_types.go
```diff
@@ -87,15 +87,15 @@ type BuildEvent struct {
 	ProjectName       string     `json:"project_name"`
 	User              *EventUser `json:"user"`
 	Commit            struct {
-		ID          int    `json:"id"`
-		SHA         string `json:"sha"`
-		Message     string `json:"message"`
-		AuthorName  string `json:"author_name"`
-		AuthorEmail string `json:"author_email"`
-		Status      string `json:"status"`
-		Duration    int    `json:"duration"`
```
svanharmelen, line 70: I cannot find this specific field in the docs. Can you show me something that shows this needs to be a `float64`?

event_webhook_types.go
```diff
@@ -340,10 +339,10 @@ type IssueEvent struct {
 		TimeEstimate        int      `json:"time_estimate"`
 		TotalTimeSpent      int      `json:"total_time_spent"`
 		TimeChange          int      `json:"time_change"`
-		HumanTotalTimeSpent string   `json:"human_total_time_spent"` // TODO Check type
-		HumanTimeEstimate   string   `json:"human_time_estimate"`    // TODO Check type
-		HumanTimeChange     string   `json:"human_time_change"`      // TODO Check type
-		Weight              string   `json:"weight"`                 // TODO Check type
```
svanharmelen: So did you check/confirm this? Can you share your findings? Same question goes for all places where you removed similar comments.

event_webhook_types.go
```diff
@@ -426,28 +425,28 @@ type JobEvent struct {
...
+		Duration    float64 `json:"duration"`
+		StartedAt   string  `json:"started_at"`
+		FinishedAt  string  `json:"finished_at"`
 	} `json:"commit"`
 	Repository *Repository `json:"repository"`
 	Runner     struct {
 		ID          int      `json:"id"`
 		Active      bool     `json:"active"`
 		RunnerType  string   `json:"runner_type"`
-		Shared      bool     `json:"is_shared"` // TODO variable name should be IsShared
```
svanharmelen: Does this break backwards compatibility? As in, was this field called `Shared` in older versions?

event_webhook_types.go
```diff
@@ -672,15 +671,15 @@ type MergeEvent struct {
 	Repository *Repository   `json:"repository"`
 	Labels     []*EventLabel `json:"labels"`
 	Changes    struct {
-		Description struct { // TODO GitLab doc doesn't mention this field
```
svanharmelen: Same question, were you able to confirm this? Can you share where/how?

svanharmelen (changes requested):

svanharmelen:
OK, I just noticed I commented on a specific commit instead of the whole PR and that the PR is still labelled as `Draft` 😊 So you should probably take my comments with a grain of salt...

I'll check again when the `Draft` label is removed and please also note that you can actually create a draft PR instead of just adding the text `Draft` as a label...

event_webhook_types.go
smit-modi: Yes, it would break backward compatibility. It was called `Shared` in older versions

event_webhook_types.go
smit-modi: My bad, it was int. I tested in local.

smit-modi:
> I thought ya'll had an alternative to `time.Time` that you were using, but I only saw `time.Time` in these structures. The name `customTime` is actually coming from other code that we're currently using. Is this something that we could consider adding to the `ISOTime`? Or could we perhaps add some additional functionality to `ISOTime` or `customTime` to track what format was read, for re-marshaling later. (Though we'd have to have some sort of functionality to map it to 3339, as that's what our existing code is intended to do...)

@svanharmelen waiting for your response on this.

GaikwadPratik:
@smit-modi,

For your reference,
https://gitlab.com/gitlab-org/gitlab/-/issues/394785
https://gitlab.com/gitlab-org/gitlab/-/issues/19567

So we will need more than one formats to be handled, as I don't see any of the above ticket being resolved sooner or at all. They kept the same format as `webhooks` in `graphql` response for some reason when using `time.RFC3339` is closest to being perfect format when used in UTC.

timofurrer:
@smit-modi  we've migrated the repository over to [GitLab.com](https://gitlab.com/gitlab-org/api/client-go).

PR #1676, opened 2023-03-21
fix(stats): fix project issue stats api milestone parameter

alexjx:
The project issue statistics api requires milestone string, but it currently is a struct. This patch fixed this parameter to make this api works.


svanharmelen (approved):
Interesting finding... But looks like your right (according to the docs). So thanks 👍🏻

PR #1680, opened 2023-03-23
Fix uptade protected branch

### LINKED ISSUE #1657: ProtectedBranches.UpdateProtectedBranch doesn't seem to work
Hey there. I'm using v0.80.1
I think UpdateProtectedBranch method is broken.

I call it and I get an error
```
devPermission := gitlab.DeveloperPermissions
maintainerPermission := gitlab.MaintainerPermissions
_, _, err = c.gl.ProtectedBranches.UpdateProtectedBranch(project.ID, project.DefaultBranch, &gitlab.UpdateProtectedBranchOptions{
	AllowedToPush:  &[]*gitlab.BranchPermissionOptions{{AccessLevel: &maintainerPermission}},
	AllowedToMerge: &[]*gitlab.BranchPermissionOptions{{AccessLevel: &devPermission}},
})
```
`SetMergePermissions error PATCH: 400 {error: allowed_to_push is invalid, allowed_to_merge is invalid}`

RawQuery for the request is `allowed_to_merge={<nil> <nil> <nil> 0xc0003f24d0}&allowed_to_push={<nil> <nil> <nil> 0xc0003f24d8}`

I've tried to override NewRequest and send the payload in the body, but marshaling result was `<unsupported type: []uint8>`

Also, as far as I understood from docs, you have to send permission ID in the request. But BranchPermissionOptions struct lacks the field.

bmsareias:
Adds missing elements for the full API functionality of UpdateProtectedBranch() method.

As consequence of this, it's mandatory by API to ship this data via body to the API, therefore I've added the case to the already existing switch-case.

This should have more extensive tests via the testing suit. I've only tested this for my own needs.

This PR fixes #1657

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -568,7 +568,7 @@ func (c *Client) NewRequest(method, path string, opt interface{}, options []Requ
 
 	var body interface{}
 	switch {
-	case method == http.MethodPost || method == http.MethodPut || method == http.MethodPatch:
+	case method == http.MethodPatch || method == http.MethodPost || method == http.MethodPut:
 		reqHeaders.Set("Content-Type", "application/json")
 
 		if opt != nil {
--- a/gitlab.go
+++ b/gitlab.go
@@ -568,7 +568,7 @@ func (c *Client) NewRequest(method, path string, opt interface{}, options []Requ
 
 	var body interface{}
 	switch {
-	case method == http.MethodPost || method == http.MethodPut:
+	case method == http.MethodPost || method == http.MethodPut || method == http.MethodPatch:
 		reqHeaders.Set("Content-Type", "application/json")
 
 		if opt != nil {
--- a/project_managed_licenses_test.go
+++ b/project_managed_licenses_test.go
@@ -124,10 +124,7 @@ func TestEditManagedLicenses(t *testing.T) {
 
 	mux.HandleFunc("/api/v4/projects/1/managed_licenses/3", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, http.MethodPatch)
-		approvalStatus := r.URL.Query().Get("approval_status")
-		if approvalStatus != "blacklisted" {
-			t.Errorf("query param approval_status should be blacklisted but was %s", approvalStatus)
-		}
+		testBody(t, r, `{"approval_status":"blacklisted"}`)
 		mustWriteHTTPResponse(t, w, "testdata/edit_project_managed_license.json")
 	})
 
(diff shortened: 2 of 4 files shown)


svanharmelen:
Did you try to get it working without using the body (so with using query params instead)? Can you post the result/error when you tried that?

svanharmelen:
FYI I'm asking because if you add `http.MethodPatch` to the list of calls that should use the request body instead of query params, you change this for all calls that currently use `PATH` requests.

So to avoid potential issues with other calls I prefer to not change this. Unless you can show that you tested all other impacted calls and they do not break...

MUlt1mate:
This fix works for me as well. @bmsareias thank you

As for using query params - it doesn't work. It will generate `RawQuery="allowed_to_merge={<nil> <nil> <nil> <nil> 0xc000032bb0 <nil>}&allowed_to_push={<nil> <nil> <nil> <nil> 0xc000032bb8 <nil>}"`

bmsareias:
> This fix works for me as well. @bmsareias thank you
> 
> As for using query params - it doesn't work. It will generate `RawQuery="allowed_to_merge={<nil> <nil> <nil> <nil> 0xc000032bb0 <nil>}&allowed_to_push={<nil> <nil> <nil> <nil> 0xc000032bb8 <nil>}"`

@MUlt1mate glad to hear! Yes from my tests all calls that want to UPDATE/DELETE (and to some extent CREATE) new entries on any of those 3 array/list/slices results in 400 - Invalid <Attribute_Name>.

svanharmelen:
Thanks for the responses! Funny enough I don't have access to a GitLab instance myself, but looking through the code I see 3 other API's (implemented by this package) that currently use `MethodPatch`:

- https://docs.gitlab.com/ee/api/users.html#disable-two-factor-authentication 
- https://docs.gitlab.com/ee/api/managed_licenses.html#edit-an-existing-managed-license
- https://docs.gitlab.com/ee/api/error_tracking.html#enable-or-disable-the-error-tracking-project-settings

If you can verify that these 3 calls also work when using the request body, then I am good to merge this PR...

bmsareias:
The Users API only applies to Self-Managed Instances (which I don't have one accessible at this point)
The Managed License API is already marked as deprecated so I don't see any benefit in really testing it.
The Error Tracking I do not have access to a project that has this feature configured in it.

Can try to test the Users API if I manage to have the time over the weekend, and will ask around for a project that has the Error Tracking feature that I (or proxied through someone else) has maintainer access on it.

svanharmelen:
OK... Well I guess in this case I'm good with moving this one forward and fixing things forward when it turned out one of those 3 APIs does end up causing issues...

svanharmelen (approved):
I'll take it... Thanks @bmsareias and @MUlt1mate 👍🏻

PR #1692, opened 2023-04-11
Updated Job event

smit-modi:
Added RunnerType, Tags in Runner struct. Added Environment.

Renamed Shared to IsShared to follow the standard guideline.


svanharmelen (approved):
Thanks @smit-modi, LGTM!

PR #1697, opened 2023-04-12
fix: `SystemOutput` in the `PipelineTestCase`

### LINKED ISSUE #1696: Fix SystemOutput type in pipelinetescases
`SystemOutput` is not a string type but a struct having properties `type` and `mesage`.

GaikwadPratik:
It is either a null or struct. Not a string.
Fixes #1696.


pipelines.go
```diff
@@ -115,12 +115,17 @@ type PipelineTestCases struct {
...
 	File           string         `json:"file"`
 	ExecutionTime  float64        `json:"execution_time"`
-	SystemOutput   string         `json:"system_output"`
+	SystemOutput   SystemOutput   `json:"system_output"`
 	StackTrace     string         `json:"stack_trace"`
 	AttachmentURL  string         `json:"attachment_url"`
 	RecentFailures RecentFailures `json:"recent_failures"`
 }
 
+type SystemOutput struct {
```
svanharmelen, line 124: Please add a simple comment, like `// SystemOutput describes the pipeline test results.`

pipelines.go
```diff
@@ -121,6 +121,7 @@ type PipelineTestCases struct {
 	RecentFailures RecentFailures `json:"recent_failures"`
 }
 
+// SystemOutput contains information about test cases when it fails
```
svanharmelen, line 124:
```suggestion
// SystemOutput contains information about test cases when it fails.
```
Following the Go coding guides and the style in this package 😏

PR #1701, opened 2023-04-14
Replaced Visibility with VisibilityLevel in Webhooks

smit-modi:
GitLab doesn't return Visibility with string type anymore. It returns VisibilityLevel with int type.

### DIFF
--- a/types.go
+++ b/types.go
@@ -738,6 +738,20 @@ func Visibility(v VisibilityValue) *VisibilityValue {
 	return p
 }
 
+// VisibilityLevelValue represents a visibility level within GitLab in int type.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/
+type VisibilityLevelValue int
+
+// List of available visibility levels.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/
+const (
+	PrivateVisibilityLevel  VisibilityLevelValue = 0
+	InternalVisibilityLevel VisibilityLevelValue = 10
+	PublicVisibilityLevel   VisibilityLevelValue = 20
+)
+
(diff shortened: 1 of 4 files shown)


smit-modi:
@svanharmelen Can you please review this?

types.go
```diff
@@ -738,6 +738,20 @@ func Visibility(v VisibilityValue) *VisibilityValue {
...
+
+// List of available visibility levels.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/
+const (
+	PrivateVisibilityLevel  VisibilityLevelValue = 0
+	InternalVisibilityLevel VisibilityLevelValue = 10
+	PublicVisibilityLevel   VisibilityLevelValue = 20
+)
+
```
svanharmelen, line 754:
```go
// VisibilityLevel is a helper routine that allocates a new
// VisibilityLevelValue to store v and returns a pointer to it.
func VisibilityLevel(v VisibilityLevelValue) *VisibilityLevelValue {
	p := new(VisibilityLevelValue)
	*p = v
	return p
}
```

Can you add something like this ☝🏻 I try to add such a helper function for all custom types to make it easier for people to get a pointer to such a value (which is needed in option structs).

svanharmelen (changes requested):
Besides the one minor comment, I'm a little bit unsure what this change means for backwards compatibility. Can you tell from what version GitLab changed this behavior?

I guess the easiest way to keep supporting older versions is to not change the existing field, but instead add the new field next to the existing one...

Or do you have any other thoughts?

timofurrer:
@smit-modi we've migrated the repository over to [GitLab.com](https://gitlab.com/gitlab-org/api/client-go).

PR #1702, opened 2023-04-15
feat: Add `email` field to `GroupMember`

candiduslynx:
This is required for https://github.com/cloudquery/cloudquery/issues/10117.
Although the `email` attribute is only visible to group Owners for any [enterprise user](https://docs.gitlab.com/ee/user/enterprise_user/index.html) (see [limitations](https://docs.gitlab.com/ee/api/members.html#limitations)), it'd be nice to have this included.

Might be safely ditched in favor of #1703 (I presume #1703 is a better one)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ _testmain.go
 *.iml
 *.swp
 *.swo
+
+# vendor
+vendor
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,3 @@ _testmain.go
 *.iml
 *.swp
 *.swo
-
-# vendor
-vendor
--- a/group_members.go
+++ b/group_members.go
@@ -54,6 +54,7 @@ type GroupMember struct {
 	CreatedAt         *time.Time               `json:"created_at"`
 	ExpiresAt         *ISOTime                 `json:"expires_at"`
 	AccessLevel       AccessLevelValue         `json:"access_level"`
+	Email             string                   `json:"email,omitempty"`
 	GroupSAMLIdentity *GroupMemberSAMLIdentity `json:"group_saml_identity"`
 }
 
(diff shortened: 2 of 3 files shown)


candiduslynx:
@svanharmelen could you please take a look?

.gitignore
```diff
@@ -28,3 +28,6 @@ _testmain.go
 *.iml
 *.swp
 *.swo
+
+# vendor
+vendor
```
svanharmelen, line 33: Please revert this change...

svanharmelen (changes requested):

.gitignore
candiduslynx: Any reason why this should be left included?

candiduslynx:
@svanharmelen I've reverted vendor change + added the typo fix in here

.gitignore
svanharmelen: Because this package doesn't have a `vendor` directory checked into git, so excluding it is actually no-op 😉 If you do anything with a vendor directory in your code, then it makes much more sense to exclude the directory in your own gitignore file.

svanharmelen (approved):
Thanks 👍🏻

.gitignore
candiduslynx: go mod vendor -> filled in the vendor dir -> need to actually check what files you want to add instead of all...

candiduslynx:
I'd still prefer to go through with #1703, though...

candiduslynx:
@svanharmelen any ETA on the release that will include this fix?

svanharmelen:
Maybe first add a PR for the vendor stuff? Or are you good without that?

candiduslynx:
Yep, in a minute!

candiduslynx:
@svanharmelen #1704 is there.

svanharmelen:
Cool, noticed on minor issue with the additional. Will merge and release when that is fixed/updated 👍🏻

candiduslynx:
updated

svanharmelen:
Released `v0.83.0`

PR #1703, opened 2023-04-16
fix: Merge `ProjectMember`, `GroupMember` & `BillableGroupMember` into a single `Member` struct

candiduslynx:
This is a more correct alternative to #1702 and it does the following:
1. Merges the structs described in the same API doc into a single implementation (fields are just a union of the `BillableGroupMember`, `GroupMember` & `ProjectMember`)
2. Fixes the typo in `removable` field (see the [docs](https://docs.gitlab.com/ee/api/members.html) to see that it is really a typo in the codebase ATM).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ _testmain.go
 *.iml
 *.swp
 *.swo
+
+# vendor
+vendor
--- a/group_members.go
+++ b/group_members.go
@@ -19,7 +19,6 @@ package gitlab
 import (
 	"fmt"
 	"net/http"
-	"time"
 )
 
 // GroupMembersService handles communication with the group members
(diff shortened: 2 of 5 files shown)


candiduslynx:
@svanharmelen this is alternative to #1702 and it does the following:
1. Merges the structs described in the same API doc into a single implementation (fields are just a union of the `BillableGroupMember`, `GroupMember` & `ProjectMember`)
2. Fixes the typo in `removable` field (see the [docs](https://docs.gitlab.com/ee/api/members.html) to see that it is really a typo in the codebase ATM).

svanharmelen:
Hi @candiduslynx 👋🏻 Thanks for opening this PR, but I prefer to follow the GitLab API as close as possible. So that means that I prefer to have named struct with fields that match the API responses over more generic structs that combine entities. So I suggest we close this PR and merge (when the comment is resolved) #1702

candiduslynx:
> I prefer to follow the GitLab API as close as possible

@svanharmelen that's exactly the point of this PR.
The issue is, that the 3 types you have are, in fact, the same entity in the GitLab API.
See that it's done so in the Java SDK https://github.com/gitlab4j/gitlab4j-api/blob/main/src/main/java/org/gitlab4j/api/ProjectApi.java#L1454

svanharmelen:
Can you show me where in the GitLab code this is one type? As the link you gave is a Java API client. So the fact that they choose to use one type for this, doesn't mean the GitLab API does so too.

candiduslynx:
@svanharmelen 
https://gitlab.com/gitlab-org/gitlab-foss/-/blob/master/app/models/members/project_member.rb
https://gitlab.com/gitlab-org/gitlab-foss/-/blob/master/app/models/members/group_member.rb

Both of those refer to https://gitlab.com/gitlab-org/gitlab-foss/-/blob/master/app/models/member.rb
(that actually defines the fields)

svanharmelen:
Yeah so `ProjectMembers` and `GroupMembers` _could_ be merged into one type, but `BillingGroupMember` seems to be a different type (which also seems to be indicated by the different fields). So does it fix an actual problem? If not I still prefer to keep it as is.

candiduslynx:
Could you point to the code that shows that `BillableGroupMember` is different? It's just a filtered `GroupMember`, or am I wrong in this one?

svanharmelen:
Well I'm not going to dive too deep into this one (not going to dive into the GitLab codebase) as I simply don't have time for that at the moment...

But I'm mainly looking at the docs (which I try to follow as guideline so that people of this package can look at the docs to understand what they can expect). And based on that `BillableGroupMember` exposes fields that `ProjectMember` and `GroupMember` don't.

So even if GitLab unifies some types in their code and then adds some logic to expose only certain fields for certain APIs, I think it makes more sense for us as API client to follow the exposed API and expose types matching the exposed APIs...

PR #1707, opened 2023-04-22
refactor: created separate struct for `LastPipeline`

GaikwadPratik:
A pipeline schedule can be created but not used so the `LastPipeline` property can be `nil`. This change will prevent panic or using `reflection` to check if it's defined or not.

### DIFF
--- a/pipeline_schedules.go
+++ b/pipeline_schedules.go
@@ -35,23 +35,26 @@ type PipelineSchedulesService struct {
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/pipeline_schedules.html
 type PipelineSchedule struct {
-	ID           int        `json:"id"`
-	Description  string     `json:"description"`
-	Ref          string     `json:"ref"`
-	Cron         string     `json:"cron"`
-	CronTimezone string     `json:"cron_timezone"`
-	NextRunAt    *time.Time `json:"next_run_at"`
-	Active       bool       `json:"active"`
-	CreatedAt    *time.Time `json:"created_at"`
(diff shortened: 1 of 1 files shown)


pipeline_schedules.go
```diff
@@ -35,23 +35,26 @@ type PipelineSchedulesService struct {
...
+	Active       bool                          `json:"active"`
+	CreatedAt    *time.Time                    `json:"created_at"`
+	UpdatedAt    *time.Time                    `json:"updated_at"`
+	Owner        *User                         `json:"owner"`
+	LastPipeline *PipelineScheduleLastPipeline `json:"last_pipeline"`
+	Variables    []*PipelineVariable           `json:"variables"`
+}
+
+// PipelineScheduleLastPipeline represents the last pipeline ran by schedule
+type PipelineScheduleLastPipeline struct {
```
svanharmelen, line 53:
```suggestion
type LastPipeline struct {
```
I guess that should be long enough 😏

svanharmelen (changes requested):

pipeline_schedules.go
GaikwadPratik: Sounds good. I am in the habit of usually prepending the parent name to have uniqueness. :)

svanharmelen (approved):
Thanks! LGTM 👍🏻

pipeline_schedules.go
svanharmelen: Yeah fully understand, but when not needed I prefer to keep things a little shorter 😊

PR #1709, opened 2023-04-23
Return Environment on stop.

hsmade:
This removes manual parsing of the response json in clients, when they want to know the state, for example.

### DIFF
--- a/environments.go
+++ b/environments.go
@@ -84,7 +84,7 @@ func (s *EnvironmentsService) ListEnvironments(pid interface{}, opts *ListEnviro
 		return nil, resp, err
 	}
 
-	return envs, resp, err
+	return envs, resp, nil
 }
 
 // GetEnvironment gets a specific environment from a project.
@@ -109,7 +109,7 @@ func (s *EnvironmentsService) GetEnvironment(pid interface{}, environment int, o
 		return nil, resp, err
 	}
 
-	return env, resp, err
+	return env, resp, nil
 }
 
 // CreateEnvironmentOptions represents the available CreateEnvironment() options.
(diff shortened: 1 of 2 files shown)


timofurrer (approved):
Awesome, thanks!

environments.go
```diff
@@ -208,17 +208,19 @@ func (s *EnvironmentsService) DeleteEnvironment(pid interface{}, environment int
...
 	req, err := s.client.NewRequest(http.MethodPost, u, nil, options)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
-	return s.client.Do(req, nil)
+	env := new(Environment)
+	resp, err := s.client.Do(req, env)
+	return env, resp, err
```
svanharmelen, line 225:
```suggestion
	env := new(Environment)
	resp, err := s.client.Do(req, env)
	if err != nil {
		return nil, resp, err
	}
	
	return env, resp, nil
```
For consistency... Also noticed this in this file the returns always return `err` even when the `err` is already checked. So feel free (please do 😉 ) to update the other return statements as well 😊

svanharmelen (changes requested):

environments.go
hsmade: Hmm not only this file I see. 2 other random files I checked also always return err, even when it's already checked. 
Want me to do a search replace in a new PR maybe?

svanharmelen: Sure... I prefer to keep things consistent, so please do 😊

hsmade: https://github.com/xanzy/go-gitlab/pull/1710

svanharmelen (approved):
Thanks, LGTM 👍🏻

svanharmelen:
@hsmade can you check and fix the test? When it passes I'll merge it...

hsmade:
Took me a bit of debugging. Which is fun because it just works when you run it using delve. But in the end the fix makes sense actually.  It now returns an object, so it needs some json from the API.

PR #1710, opened 2023-04-25
Return nil if no error.

hsmade:
If err was already checked, return nil for readability.


svanharmelen (approved):
WOW!! I had no idea there were this much instances of this 🙈 Thanks a lot for this PR @hsmade! Love it ❤️

PR #1714, opened 2023-05-01
Update global lint method (deprecated)

ltickett:
The global lint endpoint was deprecated in GitLab 16.0.


svanharmelen (approved):
Thanks @ltickett 👍🏻

PR #1716, opened 2023-05-09
update ListProjectVisibleEventsOptions according to the latest doc

HugoNgai:
According to the latest doc https://docs.gitlab.com/ee/api/events.html#list-a-projects-visible-events, 
ListProjectVisibleEventsOptions should contain action, targetType ... etc.

image

### DIFF
--- a/events.go
+++ b/events.go
@@ -196,7 +196,13 @@ func (s ProjectEvent) String() string {
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/events.html#list-a-projects-visible-events
-type ListProjectVisibleEventsOptions ListOptions
+type ListProjectVisibleEventsOptions struct {
+	Action     *EventTypeValue       `url:"action,omitempty" json:"action,omitempty"`
+	TargetType *EventTargetTypeValue `url:"target_type,omitempty" json:"target_type,omitempty"`
+	Before     *ISOTime              `url:"before,omitempty" json:"before,omitempty"`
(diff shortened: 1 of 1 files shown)


events.go
```diff
@@ -196,7 +196,13 @@ func (s ProjectEvent) String() string {
...
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/events.html#list-a-projects-visible-events
-type ListProjectVisibleEventsOptions ListOptions
+type ListProjectVisibleEventsOptions struct {
+	Action     *EventTypeValue       `url:"action,omitempty" json:"action,omitempty"`
+	TargetType *EventTargetTypeValue `url:"target_type,omitempty" json:"target_type,omitempty"`
+	Before     *ISOTime              `url:"before,omitempty" json:"before,omitempty"`
+	After      *ISOTime              `url:"after,omitempty" json:"after,omitempty"`
+	Sort       *string               `url:"sort,omitempty" json:"sort,omitempty"`
+}
```
svanharmelen, line 205: You need to embed the `ListOptions` as well. Please check other list option structs for an example...

svanharmelen, line 205: You need to embed the `ListOptions` struct as well...

svanharmelen (changes requested):

events.go
HugoNgai: seems `ListOptions` not shown in API doc, not sure whether it will support `ListOptions`...

svanharmelen (approved):
Thanks @HugoNgai 👍🏻  LGTM!

Issue #1715, opened 2023-05-04
Search: Can't unmarshal Blob because Blob.ID can apparently be a string


BSchwetzel:
Hello,

I have written a program that uses the search api to find blobs in projects. The field `Blob.ID` is expected to be an int: https://github.com/xanzy/go-gitlab/blob/master/search.go#L251

Unfortunately, I found that for all my search results, either on gitlab.com or on our own instance, the ID returned by the GitLab API is either `null` or a hex string that looks like a SHA-1 hash, given that it's 40 characters long. 

The null value is apparently expected, it's shown like that in the API documentation: https://docs.gitlab.com/ee/api/search.html#scope-blobs-premium-2
Also, it just works :-)

But when a hex string is returned it obviously doesn't work:
`2023/05/04 09:51:43 Error: json: cannot unmarshal string into Go struct field Blob.id of type int`

So, maybe this field is supposed to be a string now?

```json
[
  {
    "basename": "redacted",
    "data": "redacted",
    "path": "redacted",
    "filename": "redacted",
    "id": "a9fdaa66911a2b2c9b8fc3f4f4feb5e3dfd2542d",
    "ref": "master",
    "startline": 1,
    "project_id": 1
  }
]
```

I'm using v0.83.0

svanharmelen:
Open to a PR (should be a very easy fix)... But I am curious to know if this has never worked, or that GitLab changed something in their API?

BSchwetzel:
From what I could gather, being totally unfamiliar with the GitLab codebase and Ruby in general, it was always supposed to be a string while mostly being nil. As an example, in the Elasticsearch indexer (https://gitlab.com/gitlab-org/gitlab-elasticsearch-indexer/-/blob/main/indexer/blob.go?ref_type=heads#L27) it was supposed to be a string since the very beginning. 

Additionally, the search functionality seems to return a FoundBlob, which in turn seems to load a Gitlab::Git::Blob (https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/gitlab/search/found_blob.rb#L93), which, from my understanding, actually means the git objects. This would explain the SHA-1, at least.

But please take these observations with a grain of salt :-)

svanharmelen:
Thanks for the context! Merged the PR...

PR #1718, opened 2023-05-10
change blob id type to string

BSchwetzel:
As stated in #1715 unmarshalling search results is not possible due to the API returning null or strings as blob IDs. This PR changes the data type to string.


svanharmelen (approved):
Thanks, I'll take it 👍🏻

PR #1719, opened 2023-05-15
Add missing Project lint options

robbydyer:
The two different project lint calls were missing some options from https://docs.gitlab.com/ee/api/lint.html#validate-a-projects-ci-configuration


svanharmelen (approved):
Thanks @robbydyer! LGTM 👍🏻

PR #1736, opened 2023-06-13
personal_access_tokens.go: Implement Rotate

prarit:
GitLab version 16.0 introduced the ability to 'rotate' a token.  Via the API a user can specify a current token ID, mark it revoked, and create a new one-week expiry token. [1]

Implement Rotate token functionality.

[1] https://docs.gitlab.com/ee/api/personal_access_tokens.html#rotate-a-personal-access-token


svanharmelen:
Let me know when this is ready for review...

prarit:
@svanharmelen this can be reviewed now.

personal_access_tokens.go
```diff
@@ -134,3 +134,23 @@ func (s *PersonalAccessTokensService) RevokePersonalAccessToken(token int, optio
 
 	return s.client.Do(req, nil)
 }
+
+// RotatePersonalAccessToken revokes a token and returns a new token that expires in one week.
```
svanharmelen, line 138:
```suggestion
// RotatePersonalAccessToken revokes a token and returns a new token that
// expires in one week.
```

personal_access_tokens.go
```diff
@@ -134,3 +134,23 @@ func (s *PersonalAccessTokensService) RevokePersonalAccessToken(token int, optio
...
+	}
+
+	pat := new(PersonalAccessToken)
+	resp, err := s.client.Do(req, pat)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return pat, resp, nil
+}
```
svanharmelen, line 156: Can you place this method above `RevokePersonalAccessToken` so it matches the ordering in the docs?

personal_access_tokens_test.go
```diff
@@ -237,3 +237,16 @@ func TestRevokePersonalAccessToken(t *testing.T) {
...
+
+	mux.HandleFunc("/api/v4/personal_access_tokens/0/rotate", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, http.MethodPost)
+	})
+
+	_, _, err := client.PersonalAccessTokens.RotatePersonalAccessToken(0)
+	if err != nil {
+		t.Errorf("PersonalAccessTokens.RotatePersonalAccessToken returned error: %v", err)
+	}
+}
```
svanharmelen, line 252: Can you place this method above `TestRevokePersonalAccessToken` so it matches the ordering in the docs?

svanharmelen (changes requested):
Few tiny comments, rest looks fine 👍🏻

personal_access_tokens.go
prarit: Fixed in new push.

personal_access_tokens_test.go
prarit: Fixed in new push.

personal_access_tokens.go
prarit: Fixed in new push.

prarit:
@svanharmelen thanks for the review.  Fixed the three issues you commented on.

svanharmelen (approved):
Thanks @prarit! LGTM 👍🏻

svanharmelen:
It seems the tests broke, can you fix that before I merge this one?

prarit:
@svanharmelen here's the problem with the test.  The two tokens that are available for rotating are either already revoked or expired.  This results in the error we see in the test I already have.  I could add a token but that requires administrator access and I suspect that's why that particular feature has yet to be added to go-gitlab.  My thinking is now that revoke should not have a test as it is currently untestable.  What do you think?

prarit:
@svanharmelen I've removed the test.

PR #1740, opened 2023-06-21
Add admin permission const

### LINKED ISSUE #1739: Add AdminLevel const to Types
Some APIs, namely https://docs.gitlab.com/ee/api/protected_branches.html , allow "admin" as a possible value. This issue adds that to const just so that there is a comprehensive listing of values that could be used in downstream APIs.

PatrickRice-KSC:
Closes #1739

This PR simply adds a single const value to the permission const values so that the list is comprehensive for use in downstream APIs.

### DIFF
--- a/types.go
+++ b/types.go
@@ -66,6 +66,7 @@ const (
 	DeveloperPermissions     AccessLevelValue = 30
 	MaintainerPermissions    AccessLevelValue = 40
 	OwnerPermissions         AccessLevelValue = 50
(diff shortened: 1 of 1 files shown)


PatrickRice-KSC:
@svanharmelen  - Very simply PR here, ready for your review when you have some time :)

svanharmelen (approved):
Thanks @PatrickRice-KSC 👍🏻

PR #1741, opened 2023-06-21
Added pagination helper

Roemer:
Added a pretty flexible pagination helper.

It allows to abort on certain criterias and also to process the items after each page if wanted.

Open for discussion!

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -950,3 +950,48 @@ func parseError(raw interface{}) string {
 		return fmt.Sprintf("failed to parse unexpected error type: %T", raw)
 	}
 }
+
+// Helper method to get all items with pagination.
+// Allows passing an abort function to define early aborts
+// eg. for only certain pages or until a condition (id > xxx) is met.
+// Also allows passing a process method that allows to process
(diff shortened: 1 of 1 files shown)


timofurrer:
I've implemented something similar in #1498 🤔

timofurrer:
I guess iterators would also be nice for this once they land 🤔

Roemer:
Any feedback here? I still think this would be a good addition.

JordanP:
There's not a lot of Gitlab specific things in there. It increases the API for something that could live outside of this package. For some reason I prefer the aesthetics of https://github.com/xanzy/go-gitlab/pull/1498 Also maybe we should wait for "official" support of Iterator in Go ?

costela:
O somehow totally missed this PR and duplicated some of the work here: #1875
Apologies for that :sweat: 

That PR also includes initial [`rangefunc`](https://go.dev/wiki/RangefuncExperiment) support, now that iterators are a thing (albeit experimentally).

svanharmelen:

Issue #1743, opened 2023-06-23
cannot unmarshal string into Go struct field TodoTarget.target.id of type int


josephburnett:
When mentioned on a commit, the target id will be the commit hash which is a string. The [`TodoTarget.ID`](https://github.com/xanzy/go-gitlab/blob/0886867b3ba7ab3554169fab2b418868c9f016b5/todos.go#L61) is an int so unmarshaling fails.

```json
        "action_name": "mentioned",
        "target_type": "Commit",
        "target": {
            "id": "6a49b920e067e123703578ac38f4819f8e5a2823",
```

```
cannot unmarshal string into Go struct field TodoTarget.target.id of type int
```

josephburnett:
The [GitLab TODO API page](https://docs.gitlab.com/ee/api/todos.html) pretty clearly shows an integer ID. So a string is unexpected.

```json
    "action_name": "marked",
    "target_type": "MergeRequest",
    "target": {
      "id": 34,
```

josephburnett:
Opened an issue with GitLab: https://gitlab.com/gitlab-org/gitlab/-/issues/416180

josephburnett:
Unfortunately we cannot change the type of `ID` because it would be a breaking change. So instead I've implemented a workaround that creates a derived field `IDString`: https://github.com/xanzy/go-gitlab/pull/1744

The GitLab `id` will be routed to either `ID` or `IDString` based on it's JSON type. And either `ID` or `IDString` will be emitted as `id`. So the effect is entirely contained in the library and no visible in the JSON output.

svanharmelen:
Fixed by #1744

PR #1744, opened 2023-06-23
Derive TodoTarget.IDString

josephburnett:
Create derived field IDString to work around
https://github.com/xanzy/go-gitlab/issues/1743.

Sometimes GitLab will return a string id instead of an int (e.g. when target is a commit). This change intercepts a string id before it causes unmarshaling to fail and redirects it to the new IDString field. Conversely when marshaling it routes either ID or IDString into the expected JSON "id" field.


svanharmelen:
Thanks for your PR, but I'm not a big fan of this solution.

If the GitLab API sometimes returns an `int` and sometimes a `string`, then for now the best solution is probably to make the field type `interface{}`.

Also not ideal, but at least we are not introducing arbitrary fields that do not exists in the actual APIs...

josephburnett:
Yeah, it should have been an `interface{}` to begin with if GitLab was going to return different types. However changing it now is a breaking change. Do you plan to release the type change with a major version bump?

svanharmelen:
The API changes quite a lot.  Generally the changes are backwards compatible, but sometimes they are not.

One of the reasons I'm still using 0.x.x versions. So I'm OK with changing this it it "fixes" a bug...

svanharmelen:
Do you intent to update this PR following our conversation, or should it be closed?

svanharmelen (approved):
Update according to our discussion... So LGTM now 👍🏻

aelgasser:
@svanharmelen I think there is a typo in the last commit: 

```go
	ID                   int                    `json:"id"`
+	IID                  interface{}            `json:"iid"`
```

You set the `interface{}` type on the json `iid` field where the issue is on the `id` (with only 1 I) (actually maybe on both, better safe than sorry, but I don't see an `iid` field in my test below).

I just tested it on my project: 

```json
    {
        "id": 89908,
        "project": {
            "id": 63,
            "description": "",
            "name": "REDACTED",
            "name_with_namespace": "REDACTED",
            "path": "platform",
            "path_with_namespace": "REDACTED",
            "created_at": "2018-06-21T15:30:47.435Z"
        },
        "author": {
            "id": 63,
            "username": "abderraouf.elgasser",
            "name": "Abderraouf El Gasser",
            "state": "active",
            "locked": false,
            "avatar_url": "REDACTED",
            "web_url": "REDACTED"
        },
        "action_name": "directly_addressed",
        "target_type": "Commit",
        "target": {
            "id": "1d76d1b2e3e886108f662765c97f4687f4134d8c", // THE BAD ID IS HERE
            "short_id": "1d76d1b2",
            "created_at": "2024-01-29T11:17:05.000+01:00",
            "parent_ids": [
                "66ca971205388f39905d0f3fb48538dce7932bc5"
            ],
            "title": "fix: updated relay schemas",
            "message": "fix: updated relay schemas\n",
            "author_name": "Abderraouf El Gasser",
            "author_email": "REDACTED",
            "authored_date": "2024-01-29T11:17:05.000+01:00",
            "committer_name": "Abderraouf El Gasser",
            "committer_email": "REDACTED",
            "committed_date": "2024-01-29T11:17:05.000+01:00",
            "trailers": {},
            "web_url": "REDACTED"
        },
        "target_url": "REDACTED",
        "body": "@abderraouf.elgasser test on commit",
        "state": "pending",
        "created_at": "2024-01-30T17:14:46.520Z",
        "updated_at": "2024-01-30T17:14:46.520Z"
    },
```

Could you please make a change so I can tell to the mattermost plugin team [here](https://github.com/mattermost/mattermost-plugin-gitlab/issues/164) to use the latest version to fix the problem there.

svanharmelen:
Fixed in cd61482

aelgasser:
Thank you!

PR #1745, opened 2023-06-25
enable paginating ssh keys

jchorl:
Thanks for the great lib.

I cut https://gitlab.com/gitlab-org/gitlab/-/merge_requests/124539 to try to get the docs change in. I was building some automation over my gitlab and found that this works.

### DIFF
--- a/users.go
+++ b/users.go
@@ -411,7 +411,7 @@ type SSHKey struct {
 
 // ListSSHKeysOptions represents the available ListSSHKeys options.
 //
-// These are undocumented as of 2023/06/25 but empirically work.
+// GitLab API docs: https://docs.gitlab.com/ee/api/users.html#list-ssh-keys
 type ListSSHKeysOptions ListOptions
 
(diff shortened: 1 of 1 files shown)


users.go
```diff
@@ -409,11 +409,16 @@ type SSHKey struct {
 	ExpiresAt *time.Time `json:"expires_at"`
 }
 
+// ListSSHKeysOptions represents the available ListSSHKeys options.
+//
+// These are undocumented as of 2023/06/25 but empirically work.
```
svanharmelen, line 414:
```suggestion
// GitLab API docs: https://docs.gitlab.com/ee/api/users.html#list-ssh-keys
```

svanharmelen (changes requested):
I'm normally against adding undocumented features, but this one seems very obvious... Yet please just reference the docs to keep things clean 😊

svanharmelen (approved):
LGTM, thanks 👍🏻

PR #1747, opened 2023-06-27
Add support for user associations_count

Ikke:
https://docs.gitlab.com/ee/api/users.html#list-associations-count-for-user


svanharmelen (approved):
Tweaked things a tiny bit, LGTM now 👍🏻

PR #1748, opened 2023-06-27
add support for get-user-by-fingerprint-of-ssh-key and get-user-by-de…

hjweddie:
Hope that we can add support for these API:

GitLab API docs:
https://docs.gitlab.com/ee/api/keys.html#get-user-by-fingerprint-of-ssh-key
https://docs.gitlab.com/ee/api/keys.html#get-user-by-deploy-key-fingerprint


svanharmelen (approved):
Thanks @hjweddie 👍🏻

PR #1754, opened 2023-07-14
fix: project repository schedule returns single ProjectRepositoryStorageMove

billiechar:
ScheduleStorageMoveForProject should only return a single ProjectRepositoryStorageMove, Currently this will fail to unmarshal


billiechar:
@svanharmelen this is ready for a review. This was mistakenly set to return a slice, but the API will only ever return a single.

svanharmelen (approved):

PR #1756, opened 2023-07-18
Add ID to TagAccessDescription

mitar:
ID is now available and it is visible in [example responses in documentation](https://docs.gitlab.com/ee/api/protected_tags.html).


svanharmelen (approved):
Thanks @mitar 👍🏻

PR #1759, opened 2023-07-19
Add missing `housekeeping_optimize_repository_period` setting

PatrickRice-KSC:
This PR adds one missing setting to the `Settings` struct (for Application Settings) so that it can be updated and read appropriately


svanharmelen (approved):
Thanks @PatrickRice-KSC 👍🏻

PR #1765, opened 2023-07-28
Bugfix: add name parameter to group badges

PascalRoose:
I noticed that the `name` parameter was missing for the group badges API. This was implemented for project badges, but strangely not for group badges yet.

https://docs.gitlab.com/ee/api/group_badges.html


svanharmelen (approved):
Thanks @PascalRoose! Looks good 👍🏻

PR #1769, opened 2023-08-04
Add new runner api support

PatrickRice-KSC:
This MR adds support for the new user-based Gitlab Runner API, documented here: https://docs.gitlab.com/ee/api/users.html#create-a-runner

It's worth noting that it's in the `users.go` file even though it's creating a runner, because it's in the `/user` api space. I could be convinced to move it to the `runners.go` file pretty easily, but figured the default stance was to align to the upstream API space.

It's also worth noting that there is no `/users/runner` DELETE api, because it uses the existing APIs within `runners.go` to delete by ID.

Let me know if you have any questions! I'd appreciate cutting a new version after merging this MR so that I can start working on the tf provider resource :)

### DIFF
--- a/users.go
+++ b/users.go
@@ -1415,16 +1415,17 @@ func (s *UsersService) DisableTwoFactor(user int, options ...RequestOptionFunc) 
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/users.html#create-a-runner
 type CreateUserRunnerOptions struct {
-	RunnerType     string   `json:"runner_type"`
-	GroupID        int      `json:"group_id"`
-	ProjectID      int      `json:"project_id"`
-	Description    string   `json:"description"`
-	Paused         bool     `json:"paused"`
-	Locked         bool     `json:"locked"`
-	RunUntagged    bool     `json:"run_untagged"`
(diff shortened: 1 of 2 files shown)


PatrickRice-KSC:
@svanharmelen  - This should be ready for review when you have a moment!

PatrickRice-KSC:
@metanerd 

> I just found your PR, because I am also looking forward to this feature in the terraform provider. To my knowledge working on the tf provider resource should already work with `go mod replace`. With this message, it is not my intention to put pressure on this or the tf provider PR.

You're right, I could actually compile the library, use it locally, the develop that way. Sander is usually pretty quick on the review though, and between that an Renovate it means it's usually just cleaner to write it cleanly the first time :) . No worries on the pressure on the provider, I know it's a highly anticipated resource!

svanharmelen (approved):
Hey @PatrickRice-KSC 👋🏻 Take a look to the things I changed in the last commit and please try to follow the same style and pattern next time 😉 

Thanks!

PatrickRice-KSC:
Will do; thanks @svanharmelen !

PR #1770, opened 2023-08-10
Add last used at to impersonation token

mgh87:
Gitlab has the not documented field in the response of the [impersonation token api](https://docs.gitlab.com/ee/api/users.html#get-all-impersonation-tokens-of-a-user) 

I added this property to the ImpersonationToken struct.

Example for API response for impersonation token
```[{
    "id": 244,
    "name": "VSBOT_IT_PS1",
    "revoked": false,
    "created_at": "2023-07-13T15:55:45.728+02:00",
    "scopes": [
      "api",
      "admin_mode"
    ],
    "user_id": 51,
    "last_used_at": "2023-08-10T05:40:46.999+02:00",
    "active": true,
    "expires_at": "2024-07-12",
    "impersonation": true
  }]
```

P.s.: I found no tests for this struct, so I was unsure how to add tests without bloating the test harness.

### DIFF
--- a/users.go
+++ b/users.go
@@ -1159,14 +1159,15 @@ func (s *UsersService) RejectUser(user int, options ...RequestOptionFunc) error 
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/users.html#get-all-impersonation-tokens-of-a-user
 type ImpersonationToken struct {
-	ID        int        `json:"id"`
-	Name      string     `json:"name"`
-	Active    bool       `json:"active"`
-	Token     string     `json:"token"`
-	Scopes    []string   `json:"scopes"`
-	Revoked   bool       `json:"revoked"`
-	CreatedAt *time.Time `json:"created_at"`
-	ExpiresAt *ISOTime   `json:"expires_at"`
+	ID         int        `json:"id"`
+	Name       string     `json:"name"`
+	Active     bool       `json:"active"`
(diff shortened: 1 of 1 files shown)


mgh87:
Hello @SvenDowideit,
is there anything I'm missing to get this MR checked?

Our instance where I produced this result is on version 16.0.2 of gitlab.

svanharmelen:
I will merge this one, but as an exception to this rule: https://github.com/xanzy/go-gitlab/blob/master/CONTRIBUTING.md#contributing-code

```
This is not needed when fixing a bug or adding an enhancement, as long as the enhancement you are trying
to add can be found in the public GitLab API docs as this project only supports what is in the public API docs.
```

mgh87:
Tyvm.

I read this part of the contribution guidelines and thought for this change it is applicable that one should not open an issue.
If you prefer open issue + merge request I will do this the next time.

P.s.: ty for the awesome library.

svanharmelen:
No it's not about requiring an issue, there is no need for that! It's about the fact that this field is not documented and this package only supports documented APIs (and fields).

mgh87:
I got the gitlab docs updated.
https://gitlab.com/gitlab-org/gitlab/-/merge_requests/129745
Probably gets live with the next release.

svanharmelen:
Nice 👍🏻  Thanks!

PR #1779, opened 2023-08-21
Add support for Group Protected Environments

beekeep:
Add support for Group Protected Environments, similar to Project Protected Environments.

https://docs.gitlab.com/ee/api/group_protected_environments.html

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,8 +4,7 @@ We want to make contributing to this project as easy as possible.
 
 ## Reporting Issues
 
-If you have an issue, please report it on the [issue tracker](
-https://github.com/xanzy/go-gitlab/issues).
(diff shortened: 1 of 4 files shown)


beekeep:
@svanharmelen This is ready for your review.

svanharmelen (approved):
I'm not sure if copying over all structs is the best solution here (as most of them are a one-2-one match), but for now I'll take it... Thanks 👍🏻

beekeep:
@svanharmelen Thanks for the review.  I had the initial thought of re-using the project protected environment structs, but I wasn't sure if it would be better to create structs specific to groups, in case they do change in projects.  In the future, would you prefer re-use if the structs are the same?

svanharmelen:
I don't know... Generally I prefer to re-use structs if it makes sense, but in this case I think this is OK.

PR #1785, opened 2023-08-31
Add Draft boolean attribute to MergeType.ObjectAttributes

CescHolly:
Earlier today I added the Draft attribute to the Changes of a MergeEvent webhook type. I also saw this attribute is not present in the ObjectAttributes itself.

So this is a small correction of that (should've added it to the earlier PR but I didn't notice ;) ).

### DIFF
--- a/event_webhook_types.go
+++ b/event_webhook_types.go
@@ -618,7 +618,6 @@ type MergeEvent struct {
 		MilestoneID              int          `json:"milestone_id"`
 		StateID                  StateID      `json:"state_id"`
 		State                    string       `json:"state"`
-		Draft                    bool         `json:"draft"`
 		MergeStatus              string       `json:"merge_status"`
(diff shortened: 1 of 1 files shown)


event_webhook_types.go
```diff
@@ -618,6 +618,7 @@ type MergeEvent struct {
 		MilestoneID              int          `json:"milestone_id"`
 		StateID                  StateID      `json:"state_id"`
 		State                    string       `json:"state"`
+		Draft                    bool         `json:"draft"`
```
svanharmelen, line 621: Can you please place the field below `WorkInProgress` following the ordering of the docs?

svanharmelen (changes requested):

event_webhook_types.go
CescHolly: True!

svanharmelen (approved):

PR #1787, opened 2023-09-03
fix: squash is bool parameter not string

GaikwadPratik:
@svanharmelen , I am not sure how I missed this, but I apologize for this. It’s a small correction.


svanharmelen (approved):
👍🏻

PR #1788, opened 2023-09-04
Add Ptr function to use generics for allocating pointers

mycrEEpy:
* update minimum go version to 1.19
* deprecate all other functions which were allocating pointers before


svanharmelen:
I'm a little busy these days, but I really like this approach. So I'll certainly get back on this one, but it might take a little...

svanharmelen (approved):
Looks good tome 👍🏻  Thanks for this @mycrEEpy!

PR #1791, opened 2023-09-11
Add repositories archive path support

ChistiyKot:
According to [Gitlab Repositories API](https://docs.gitlab.com/ee/api/repositories.html#get-file-archive) it is possible to set subpath of the repository to download (optional `path` field).

> The subpath of the repository to download. If an empty string, defaults to the whole repository.

This PR adds `path` field to Repositories Archive request options.


svanharmelen (approved):
Thanks @ChistiyKot 👍🏻

PR #1800, opened 2023-09-21
Added references property to TodoTarget

iamphill:
For issues `references` gets exposed instead of just `reference`

### DIFF
--- a/todos.go
+++ b/todos.go
@@ -88,6 +88,7 @@ type TodoTarget struct {
 	MergeWhenPipelineSucceeds bool         `json:"merge_when_pipeline_succeeds"`
 	MergeStatus               string       `json:"merge_status"`
 	Reference                 string       `json:"reference"`
+	References                []string     `json:"references"`
 	Reviewers                 []*BasicUser `json:"reviewers"`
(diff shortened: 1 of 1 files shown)


svanharmelen:
I don't see the field in the docs, can you point me to a location that shows this fields exists? And if _only_ `references` is exposed, then I suggest to delete `reference` (no point in keeping it if it isn't exposed by the API).

iamphill:
@svanharmelen I'm not sure why it isn't in the docs but it is `reference` is deprecated https://gitlab.com/gitlab-org/gitlab/-/merge_requests/20354

svanharmelen:
Check... This instead we should probably deprecate `reference` like [this](https://github.com/xanzy/go-gitlab/blob/master/merge_requests.go#L104)

In addition (looking at the link to the MR you provided, it seems like `references` returns a structs instead of a slice of strings. Can you double check that your changes indeed match the new API?

svanharmelen:
Any updates @iamphill? If not I'm going to close this one...

svanharmelen:
Closing due to no response...

PR #1805, opened 2023-09-28
Support `shared_runners_setting` in Group response

takp:
Add `shared_runners_setting` in Group response, and removed `shared_runners_enabled` because it was not included in the response. (`shared_runners_enabled` is returned in the Project resources, not Group.)

I confirmed with the `curl` response and had a look at the gitlab source code as well. In addition, I created a MR to add `shared_runners_setting` to the API doc: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/132831

Ref:
- API doc: https://docs.gitlab.com/ee/api/groups.html
- [Options for shared_runners_setting](https://docs.gitlab.com/ee/api/groups.html#options-for-shared_runners_setting)

### DIFF
--- a/groups.go
+++ b/groups.go
@@ -68,7 +68,7 @@ type Group struct {
 	MentionsDisabled        bool                       `json:"mentions_disabled"`
 	RunnersToken            string                     `json:"runners_token"`
 	SharedProjects          []*Project                 `json:"shared_projects"`
-	SharedRunnersEnabled    bool                       `json:"shared_runners_enabled"`
+	SharedRunnersSetting    SharedRunnersSettingValue  `json:"shared_runners_setting"`
 	SharedWithGroups        []struct {
 		GroupID          int      `json:"group_id"`
 		GroupName        string   `json:"group_name"`
--- a/types.go
+++ b/types.go
@@ -607,8 +607,11 @@ type SharedRunnersSettingValue string
 // https://docs.gitlab.com/ee/api/groups.html#options-for-shared_runners_setting
 const (
 	EnabledSharedRunnersSettingValue                  SharedRunnersSettingValue = "enabled"
-	DisabledWithOverrideSharedRunnersSettingValue     SharedRunnersSettingValue = "disabled_with_override"
+	DisabledAndOverridableSharedRunnersSettingValue   SharedRunnersSettingValue = "disabled_and_overridable"
 	DisabledAndUnoverridableSharedRunnersSettingValue SharedRunnersSettingValue = "disabled_and_unoverridable"
+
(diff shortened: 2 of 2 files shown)


types.go
```diff
@@ -607,7 +607,7 @@ type SharedRunnersSettingValue string
 // https://docs.gitlab.com/ee/api/groups.html#options-for-shared_runners_setting
 const (
 	EnabledSharedRunnersSettingValue                  SharedRunnersSettingValue = "enabled"
-	DisabledWithOverrideSharedRunnersSettingValue     SharedRunnersSettingValue = "disabled_with_override"
```
timofurrer, line 610: I wonder if it makes sense to leave the deprecated setting in here just to be more graceful and be backwards compatible. You could mark the setting as deprecated:

```golang
// Deprecated: DisabledWithOverrideSharedRunnersSettingValue is deprecated in favor of DisabledAndOverridableSharedRunnersSettingValue  
DisabledWithOverrideSharedRunnersSettingValue       SharedRunnersSettingValue = "disabled_with_override"
DisabledAndOverridableSharedRunnersSettingValue   SharedRunnersSettingValue = "disabled_and_overridable"
```

takp: That's a good point! I added it as deprecated 👍

groups.go
```diff
@@ -68,7 +68,7 @@ type Group struct {
 	MentionsDisabled        bool                       `json:"mentions_disabled"`
 	RunnersToken            string                     `json:"runners_token"`
 	SharedProjects          []*Project                 `json:"shared_projects"`
-	SharedRunnersEnabled    bool                       `json:"shared_runners_enabled"`
+	SharedRunnersSetting    SharedRunnersSettingValue  `json:"shared_runners_setting"`
```
svanharmelen, line 71: Guess `SharedRunnersEnabled`should also be deprecated instead of removed right?

takp: Hi @svanharmelen, in my understanding, `SharedRunnersEnabled` is not an attribute of `Group` resource in the first place. `SharedRunnersEnabled` is an output of `Project` resource. So, I think we should remove `SharedRunnersEnabled` here.  cc: @timofurrer

timofurrer: @svanharmelen @takp it seems like I've introduced this field a while back (https://github.com/xanzy/go-gitlab/pull/1429) - wrongly 😰

svanharmelen: Check 👍🏻

svanharmelen (approved):
Thanks 👍🏻

timofurrer:
@svanharmelen once again, it would be awesome to get a release of this asap to unblock provider development 😇

svanharmelen:
Done...

takp:
Thanks! 😄

PR #1807, opened 2023-09-28
Add ObjectAttributes.Name to PipelineEvent

CescHolly:
Quoting the documentation [here](https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html#pipeline-events):

"
In [GitLab 16.1](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/123639) and later, pipeline webhooks started to expose object_attributes.name.
"

For anyone using a previous version, the unmarshaling of the event payload will just leave ObjectAttributes.Name with an empty string value (""). While for applications connecting to gitlab.com or a hosted version of gitlab with version 16.1 or later will be able to use this field.

### DIFF
--- a/event_webhook_types.go
+++ b/event_webhook_types.go
@@ -794,6 +794,7 @@ type PipelineEvent struct {
 	ObjectAttributes struct {
 		ID             int      `json:"id"`
 		IID            int      `json:"iid"`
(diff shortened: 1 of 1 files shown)


svanharmelen (approved):
Thanks 👍🏻

PR #1808, opened 2023-09-29
fix: correction in rest end point path

GaikwadPratik:
library adds `/` at the beginning of the path by default, no need to add it manually


svanharmelen (approved):
Thanks... 👍🏻

PR #1809, opened 2023-10-02
Align project notification events

PatrickRice-KSC:
This PR adds support for several notification events that are present on the group/project level per the API documentation here: https://docs.gitlab.com/ee/api/notification_settings.html#update-groupproject-level-notification-settings

I also added a test for the project API testing that nested events returned properly.


PatrickRice-KSC:
@svanharmelen  - This should be ready for review when you have a moment. Just adds a couple of events to the existing APIs (matched the current alphabetical order style), and added a couple tests for Update and Get.

svanharmelen (approved):
Thanks 👍🏻

PR #1813, opened 2023-10-04
Make sure response is fully consumed

mitar:
This fully consumes a response body. I have learned through some other experiences that this is important to do in Go. This is especially useful if `nil` is provided as `v`, because then response body is just closed. But also when JSON decoder does not decode all data (and there is some extra bytes after the decoded JSON response) - JSON decoder does not necessary consume full input, just enough that `v` is decoded. Any errors returned by `io.Copy` are ignored on purpose.

See similar thing in go-github library: https://github.com/google/go-github/pull/317


svanharmelen (approved):
I'll take it, thanks 👍🏻

PR #1821, opened 2023-10-11
Added NotAuthorUsername to List*MergeRequestsOptions

nedenwalker:
Added NotAuthorUsername following pattern established before with NotLabels. Added unit tests validating both AuthorUsername and Not[AuthorUsername].


nedenwalker:
totally. thanks for reducing the testdata, yeah, didn't need to be so big!

PR #1822, opened 2023-10-12
fix LinePosition fields for multiline CreateMergeRequestDiscussion API call

johnybx:
Hi,
I am trying to use this module to create multiline comments on merge request but API calls ends with error 
```
{message: 400 Bad request - Note {:position=>["must be a valid json schema"]}}
```
The problem is that `LinePosition` contains fields `OldLine` and `NewLine` which are not defined in [documentation](https://docs.gitlab.com/ee/api/discussions.html#parameters-for-multiline-comments) for merge request comments. I added `omitempty` so that it can be used while keeping backward compatibility.

### DIFF
--- a/notes.go
+++ b/notes.go
@@ -95,8 +95,8 @@ type LineRange struct {
 type LinePosition struct {
 	LineCode string `json:"line_code"`
 	Type     string `json:"type"`
-	OldLine  int    `json:"old_line"`
-	NewLine  int    `json:"new_line"`
+	OldLine  int    `json:"old_line,omitempty"`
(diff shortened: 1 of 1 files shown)


svanharmelen:
@johnybx I think this is a better solution (#1824) as we should use option structs for requests instead or reusing the response types. This should prevent the kind of issues you mentioned and are also in line with the rest of the package.

Can you test/verify if this works as expected?

johnybx:
sure, I tested it and it works as expected :+1: . Thanks.

svanharmelen:
Cool, thanks for testing!

PR #1825, opened 2023-10-20
fix: application settings marshalling

stevexuereb:
What
---
Update `Settings.PerformanceBarAllowedGroupID` to be an `int` instead `string`

Why
---
The API returns an int because it's a group ID:

```
curl -s --header "PRIVATE-TOKEN: TOKEN" "https://gitlab.example.com/api/v4/application/settings" | jq -r '.performance_bar_allowed_group_id'
1234
```

The [documentation](https://gitlab.com/gitlab-org/gitlab/-/blob/784401ac8b5bf5ae3056cc84a95fb2be37c83934/doc/api/settings.md#L92) specifies that it's an `int`, and it's also stored as an int in the [database](https://gitlab.com/gitlab-org/gitlab/-/blob/784401ac8b5bf5ae3056cc84a95fb2be37c83934/db/structure.sql#L11472).

### DIFF
--- a/settings.go
+++ b/settings.go
@@ -238,7 +238,7 @@ type Settings struct {
 	PasswordSymbolRequired                                bool              `json:"password_symbol_required"`
 	PasswordUppercaseRequired                             bool              `json:"password_uppercase_required"`
 	PasswordLowercaseRequired                             bool              `json:"password_lowercase_required"`
-	PerformanceBarAllowedGroupID                          string            `json:"performance_bar_allowed_group_id"`
+	PerformanceBarAllowedGroupID                          int               `json:"performance_bar_allowed_group_id"`
 	PerformanceBarAllowedGroupPath                        string            `json:"performance_bar_allowed_group_path"`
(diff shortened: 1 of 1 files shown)


settings.go
```diff
@@ -238,7 +238,7 @@ type Settings struct {
 	PasswordSymbolRequired                                bool              `json:"password_symbol_required"`
 	PasswordUppercaseRequired                             bool              `json:"password_uppercase_required"`
 	PasswordLowercaseRequired                             bool              `json:"password_lowercase_required"`
-	PerformanceBarAllowedGroupID                          string            `json:"performance_bar_allowed_group_id"`
+	PerformanceBarAllowedGroupID                          int               `json:"performance_bar_allowed_group_id"`
```
stevexuereb, line 241: **question:** Should I also update the tests here, or should I leave it?

svanharmelen (approved):
Thanks @stevexuereb 👍🏻

settings.go
svanharmelen: No that's fine...

PR #1827, opened 2023-10-24
Add keyset pagination options

pwlandoll:
This is one possible implementation of keyset-based pagination, as suggested in #815.

The basic flow of offset-based pagination is that the client requests the first page, and the server returns information about the page in headers. The flow for keyset-based pagination differs in that the information for building a request for the next page is not returned in pieces, but in a pre-formatted URL in the `Link` header. The documentation recommends using that link to make the request, which doesn't seem like it will work cleanly with the existing code here. In particular, the link includes a number of query parameters that differ among endpoints that support keyset-based pagination.

So, the changes here provide a way to directly use all of the query parameters returned in the header when building the next request. `ListOptions` now includes fields that are shared among all keyset-based paginated endpoints. Other required parameters for paginated endpoints can be added by means of a `RequestOptionFunc`, for which there is a builder function included.

An alternative would be to have `ListOptions` enumerate each possible query parameter that could be included in a "next Link" header, and return them in the `Response` struct in the same way as existing header values. This would mean that the end-user code for each endpoint using this pagination might look different, i.e. there would be no single way to do all keyset-based paginated endpoints. But, it would make the code less opaque, and avoid using a "magic black box function" to add `RequestOptionFunc`s. 

This PR is mostly a suggestion/WIP to get the conversation moving about how this package might support keyset-based paginated endpoints, so please let me know how it looks!

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -227,11 +227,20 @@ type Client struct {
 // ListOptions specifies the optional parameters to various List methods that
 // support pagination.
 type ListOptions struct {
-	// For paginated result sets, page of results to retrieve.
+	// For keyset-based paginated result sets, name of the column by which to order
+	OrderBy string `url:"order_by,omitempty" json:"order_by,omitempty"`
+
+	// For offset-based paginated result sets, page of results to retrieve.
 	Page int `url:"page,omitempty" json:"page,omitempty"`
 
-	// For paginated result sets, the number of results to include per page.
+	// For keyset-based paginated result sets, the value must be `"keyset"`
+	Pagination string `url:"pagination,omitempty" json:"pagination,omitempty"`
+
+	// For offset-based and keyset-based paginated result sets, the number of results to include per page.
 	PerPage int `url:"per_page,omitempty" json:"per_page,omitempty"`
+
+	// For keyset-based paginated result sets, sort order (`"asc"`` or `"desc"`)
+	Sort string `url:"sort,omitempty" json:"sort,omitempty"`
 }
 
 // RateLimiter describes the interface that all (custom) rate limiters must implement.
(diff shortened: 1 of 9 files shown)


gitlab.go
```diff
@@ -720,12 +729,23 @@ type Response struct {
...
+
+	PreviousLink string
+	NextLink     string
+	FirstLink    string
+	LastLink     string
+
+	// The "next" link in the Link header includes query parameters for the
+	// next paginated request, but the relevant parameters vary among API
+	// endpoints.
+	NextLinkParameters url.Values
```
stanhu, line 741: Minor question: I wonder if this field should be omitted, so that `BuildKeysetPaginatedRequestOptionFunc` can just take in the `NextLink` and lazily parse it.

stanhu:
Thanks for doing this! 

This looks like a pretty good approach to me. I just had one minor point for consideration.

stanhu:
@svanharmelen Would you mind looking at this? We noticed that on GitLab.com there are clients such as [`gitlab-ci-pipelines-exporter`](https://github.com/mvisonneau/gitlab-ci-pipelines-exporter) that are iterating through a large number of CI jobs with offset pagination, when keyset pagination should be used.

gitlab.go
pwlandoll: I think that's a good idea. It seems like the goal of the library right now is more "provide a clean interface without being too opinionated" rather than "provide as little as possible" or "all batteries included." I'll see if I can streamline that a little better.

stanhu:
Thanks, this looks good to me!

@svanharmelen Could you merge?

pwlandoll:
Great, thanks! I was able to pull the latest version and run code like the included example.

PR #1831, opened 2023-11-06
Support deleting labels by label-id

### LINKED ISSUE #1823: Use of deprecated Group/Project Label API
The current API methods for Delete and Update labels for projects and groups are using the deprecated api to reference the label by name rather than label ID. 

It looks like this would be a breaking change to go-gitlab to modify in place since the respective options parameters would need an id field added to the type. I'm not sure how this project handles breaking changes. Perhaps for now, new update/delete functions could be created (i.e. DeleteGroupLabelById) and the old ones removed at a later date?

ysmilda:
Closes #1823 

This implements the Gitlab Label API for deleting a label as described in https://docs.gitlab.com/ee/api/labels.html#delete-a-label

### DIFF
--- a/group_labels.go
+++ b/group_labels.go
@@ -136,21 +136,32 @@ func (s *GroupLabelsService) CreateGroupLabel(gid interface{}, opt *CreateGroupL
 	return l, resp, nil
 }
 
-// DeleteGroupLabel deletes a group label given by its name.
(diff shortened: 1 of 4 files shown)


svanharmelen (approved):
I changed things a bit to make it work for both use cases so we can keep supporting older versions as well.

PR #1840, opened 2023-11-22
Correct Json marshaling for labels of issues

tkschmidt:
Labels for issues should be marshalled to a list of strings instead of single (comma-seperated) string. This is according to https://docs.gitlab.com/ee/api/issues.html (22-11-2023)


svanharmelen (approved):
While more involved and backwards incompatible, I think this is the only "correct" solution for this one...

aaronjheng:
Why was `Description` field from `CreateMergeRequestOptions` changed to `*LabelOptions`?

svanharmelen:
Thats a mistake... Thanks for spotting it, will change it in a bit!

svanharmelen:
Fixed in `v0.95.1`

PR #1844, opened 2023-12-06
Added missing fields for flexibleRollout feature flag strategy

korotin:
Feature flag model is missing a few fields for `flexibleRollout` strategy.

These fields aren't documented in Gitlab API docs for some reason, but they are described in [Unleash docs](https://docs.getunleash.io/reference/activation-strategies#gradual-rollout) and [Gitlab documentation](https://docs.gitlab.com/ee/api/feature_flags.html#create-a-feature-flag) has link to it (see `strategies:name`).


svanharmelen (approved):
I'll take it, thank you 👍🏻

PR #1845, opened 2023-12-07
feat(options): Add unidiff option

puskunalis:
Present diffs in the [unified diff](https://www.gnu.org/software/diffutils/manual/html_node/Detailed-Unified.html) format. Default is false. [Introduced](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/130610) in GitLab 16.5.


svanharmelen (approved):
Thanks @puskunalis! Looks good 👍🏻

PR #1846, opened 2023-12-07
Add delete group milestone functionality and tests

eranor:
Added a new function, DeleteGroupMilestone, to the GroupMilestonesService which allows for the deletion of a specified group milestone.

Ref: https://docs.gitlab.com/ee/api/group_milestones.html#delete-group-milestone


svanharmelen (commented):
Looks good, thanks @eranor 👍🏻

PR #1855, opened 2023-12-14
Reorder issue

inputvalidation:
This PR implements support for the reorder issue endpoint in the REST API (https://docs.gitlab.com/ee/api/issues.html#reorder-an-issue). You can use it to update the relative position to other issues, which is then reflected on the issue board (but also allows you to sort by "relative position" in the GraphQL API).

### DIFF
--- a/issues.go
+++ b/issues.go
@@ -786,3 +786,35 @@ func (s *IssuesService) GetParticipants(pid interface{}, issue int, options ...R
 
 	return bu, resp, nil
 }
+
+// ReorderIssueOptions represents the available ReorderIssue() options.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/issues.html#reorder-an-issue
+type ReorderIssueOptions struct {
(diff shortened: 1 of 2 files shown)


issues.go
```diff
@@ -786,3 +786,35 @@ func (s *IssuesService) GetParticipants(pid interface{}, issue int, options ...R
...
+	}
+
+	i := new(Issue)
+	resp, err := s.client.Do(req, i)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return i, resp, nil
+}
```
svanharmelen, line 820: The file is ordered inline with the order in the docs, so could you please place this method between the `DeleteIssue` and `MoveIssue` methods? Goes for the test files as well... Thanks!

inputvalidation: Thanks Sander, I have reordered the reorder functions :)

svanharmelen:
Thanks @inputvalidation 👍🏻

PR #1869, opened 2024-01-18
Fix an issue with environment names not being encoded

PatrickRice-KSC:
This PR fixes an issue where environments that have slashes in their name cause an error because the URL isn't encoded. 

Upstream TF provider issue: https://gitlab.com/gitlab-org/terraform-provider-gitlab/-/issues/6198


svanharmelen (approved):
I'll take it... Thanks @PatrickRice-KSC

Issue #1320, opened 2021-12-15
Pagination helper


mitar:
It seems I am writing many boilerplate pagination loops. Maybe this package could provide a pagination helper to do this automatically?

It could help with #815, too, so that we could update pagination helper and users would automatically get benefits of keyset pagination when it is available.

svanharmelen:
Can you share an example of such boilerplate?

mitar:
It is already [in the repository](https://github.com/xanzy/go-gitlab/blob/master/examples/pagination.go). The only difference is that I am `append`ing results into a slice and then returning it at the end (instead of printing them out).

svanharmelen:
Check... But looking at that code it looks like it would mean that each own endpoint (type) would need it's own custom function (loop). I'm not too excited (read not really open) to add that to this package as I try to keep the load on maintaining this package reasonable by limiting the functionality 1:1 to the GitLab API.

Maybe when generics have landed we can see if we can write a single function that can work for all of the list calls, but until then I don't want to add a helper function for each list call.

Roemer:
I just stumbled upon this closed issue and also tried to get rid of some boilerplate with generics.
The best method I came up with for now is with the following:
```go
func getPaginated[I any, O gitlab.ListOptions](searchFunc func(pageNum int) ([]I, *gitlab.Response, error)) ([]I, *gitlab.Response, error) {
	items := []I{}
	pageNum := 1
	for {
		tempItems, response, err := searchFunc(pageNum)
		if err != nil {
			return nil, response, err
		}
		items = append(items, tempItems...)
		if response.NextPage == 0 {
			break
		}
		pageNum = response.NextPage
	}
	return items, nil, nil
}
```
The idea is that you basically use your existing method that needs pagination and you get the information which page you need and you need to add this as list option.
So for example:
```go
gitlabBranches, _, err := getPaginated(func(pageNum int) ([]*gitlab.Branch, *gitlab.Response, error) {
	return git.Branches.ListBranches(myProjectId, &gitlab.ListBranchesOptions{
		ListOptions: gitlab.ListOptions{Page: pageNum},
	})
})
```
or
```go
treeNodes, _, err := getPaginated(func(pageNum int) ([]*gitlab.TreeNode, *gitlab.Response, error) {
	return git.Repositories.ListTree(myProjectId, &gitlab.ListTreeOptions{
		Path:        gitlab.String(basePath),
		Recursive:   gitlab.Bool(false),
[...]

PR #1875, opened 2024-02-05
feat: add AllPages pagination helper

costela:
This small PR finally addresses #1320, picking up where the discussion left off.

Naming is - of course - up for discussion! :)

**Update**: the PR now also includes an experimental iterator based on the [`rangefunc` experiment](https://go.dev/wiki/RangefuncExperiment).


costela:
Oh, this is embarrassing, I totally missed #1741 when searching for previous work :confused: 

This implementation is considerably different though, so I'll leave the PR open for now.

Roemer:
Looks cool but what I am missing from my PR (https://github.com/xanzy/go-gitlab/pull/1741) is the ability to 1. somehow allow the pagination to abort on some criteria and 2. the ability to immediately process each result. Both are a bit of edge cases but can come in handy in a few situations.

Roemer:
Ah true, I misread that. Seems that the iterator approach could really do those things. That would be fantastic and way less clunky than my previous version.

costela:
Just for the record: note that the iterator doesn't technically require the `rangefunc` experiment or go1.22, but I think we should still "hide" it behind the experiment to limit API breakage to those explicity interested.

If go1.23 turns out to break the `rangefunc` experiment API, we can then change the iterator with minimal user impact.

JordanP:
It took me at least 5min to understand this line `func PageIterator[O, T any](f Paginatable[O, T], opt *O) iter.Seq2[*T, error] {` but it's beautiful !

costela:
Just 5min? Then you were faster than me! 😅

Seriously though, I think renaming the type parameters to something more verbose may help?

And I think as long as the usage is straightforward (which IMHO it still is), then the implementation is allowed to look a bit gnarly.

As a side-note, I'm personally still a bit torn on the rangefunc stuff. Powerful and elegant (at least from a runtime perspective) but twists my brain in ways I'm not yet comfortable with.

JordanP:
> As a side-note, I'm personally still a bit torn on the rangefunc stuff. Powerful and elegant (at least from a runtime perspective) but twists my brain in ways I'm not yet comfortable with.

Yep ! For me, that feeling started with the introduction of generics. As a "user"/consumer, they are a joy to use. As a code author/maintainer, they are a pain.

----

Some functions take an additional `pid interface{}` as first argument (like `Pipelines.ListProjectPipelines(...)`). So I need a function like
```go
func XXYYY[O, T any](pid interface{}, f func(interface{}, *O, ...gitlab.RequestOptionFunc) ([]*T, *gitlab.Response, error)) Paginatable[O, T] {
	return func(opt *O, optionFunc ...gitlab.RequestOptionFunc) ([]*T, *gitlab.Response, error) {
		return f(pid, opt, optionFunc...)
	}
}
```

It's a function that returns a closure, whose signature matches `type Paginatable[O, T any]`

Would that make sense to include such function in that `gitlab` package ?Is there a better way ?

iterator.go
```diff
@@ -0,0 +1,41 @@
...
+//		if err != nil {
+//			// handle error
+//		}
+//		// process individual user
+//	}
+func PageIterator[O, T any](f Paginatable[O, T], opt *O) iter.Seq2[*T, error] {
+	return func(yield func(*T, error) bool) {
+		nextLink := ""
+		for {
+			page, resp, err := f(opt, WithKeysetPaginationParameters(nextLink))
```
JordanP, line 25: The function `f` accepts a variable number of `RequestOptionFunc` arguments, but in this call here, it passes none of them.

JordanP: Maybe something like that
```go
//	for user, err := range gitlab.PageIterator(gl.Users.List, nil, nil) {
//		if err != nil {
//			// handle error
//		}
//		// process individual user
//	}
func PageIterator[O, T any](f Paginatable[O, T], opt *O, optionFunc ...gitlab.RequestOptionFunc) iter.Seq2[*T, error] {
	return func(yield func(*T, error) bool) {
		nextLink := ""
		for {
			page, resp, err := f(opt, append(optionFunc, gitlab.WithKeysetPaginationParameters(nextLink))...)
```

costela: ah yes, good point! Thankfully a quick fix. PTAL

costela:
> It's a function that returns a closure, whose signature matches type `Paginatable[O, T any]`
>
> Would that make sense to include such function in that gitlab package ?Is there a better way ?

Also a good point. However, I'm not sure about the closure approach. It's elegant, but I fear it may not be directly clear for consumers.
I've added an alternative suggestion to the code: `*ForID` variants to the pagination helpers. IMHO that's slightly more ergonomic for users, but I'll gladly refactor into the closure approach if that's the preferred way.

The two versions unfortunately only cover around 60% of list-like methods: if my grep-foo is correct, there are currently 187 list-like methods in total, 25 of the "simple" opt+optFunc type and 85 of the id+opt+optFunc type.
Hopefully that's still enough of an improvement to warrant including it?

costela:
Friendly reminder :innocent: 

Now that the `iter` package is apparently being released as GA in 1.23 (at least it's in [master](https://pkg.go.dev/iter@master) and mentioned in [the current draft release notes](https://tip.golang.org/doc/go1.23#iterators)), it may be a good idea to reconsider this approach?

costela:
And now 1.23 is out _with_ `rangefunc` support :tada: 

Any feedback here? :innocent:

jakubbortlik:
> And now 1.23 is out _with_ `rangefunc` support 🎉
> 
> Any feedback here? 😇

Yeah, it would indeed be very nice if a pagination helper was implemented. Either this one or #1741.

It would also make it possible to close one more MRs - #1498 :-)

svanharmelen:
If we can clean this PR up a little (mainly the comments I think) I'm good with merging this one...

costela:
@svanharmelen I see only [this](https://github.com/xanzy/go-gitlab/pull/1875#discussion_r1484171551) open thread, which has been addressed. Am I missing something? :thinking: 

Also: should we keep the non-rangefunc variant, or should we now go "all-in" and just provide the a `rangefunc` variant of `AllPages`? I personally would go this way, to not add unnecessary ambiguity to the API.

svanharmelen:
I was looking at these:

image

But also noticed the comments don't follow the contributing guidelines (keep comment lines below 80 chars).

svanharmelen:
I must admit that maybe I was a bit too quick yesterday, as I now realize that choosing this solution would actually make this package require at least Go 1.23.x which might cause issues for some of the people using this package.

costela:
@svanharmelen cleaned the comments and removed the non-rangefunc variant. Now this PR focus *only* on the iterator support. Also added some basic tests. PTAL :pray: 

Reminder: as mentioned above, the two variants introduced in this PR cover only about [60% of the list-like methods in the API](https://github.com/xanzy/go-gitlab/pull/1875#issuecomment-1944854078). There's probably some room to refactor the other methods and make them all a bit more similar, so that in the future more lists can be paginated with the same iterator.

costela:
@svanharmelen friendly ping :innocent: 
anything I missed from the reviews above?

svanharmelen:
Thanks for the ping and sorry for taking so long with this one... This just falls outside of "standard" fixes and enhancements and I want to take some time to play with it myself. Yet I'm not using GitLab myself anymore (for a long long time by now 🙈) so sitting myself down to actually do that in between "work" turns out to be something that doesn't happen by itself 😏 

I will try to make some time for it in the coming week(end)-ish 😉 And again sorry for all the delays and thank you for your patience!

Roemer:
One more thing I noticed with this variant: Things like `ListPipelineJobs` do not work currently as this method takes two ids whereas this helper currently only support zero or one id. And there are probably even more combinations of required parameters which are also not covered yet. This is probably the one thing my proposal (https://github.com/xanzy/go-gitlab/pull/1741) did better: It is completely flexible in terms of parameters.

timofurrer:
@costela  we've migrated the repository over to [GitLab.com](https://gitlab.com/gitlab-org/api/client-go).

PR #1878, opened 2024-02-07
The snippets API was failing due to improper URL encoding of the filename parameter

Sild:
GitLab api error:

```
GET https://gitlab.ZZZ.com/api/v4/snippets/YYY/files/HEAD/version 1/raw: 404 {message: 404 File Not Found}'
```

Root Cause: 
 
The filename contains spaces which need to be properly encoded in the URL path. url.QueryEscape was being used instead of PathEscape.

Resolution:  

Use PathEscape when encoding the filename since it is part of the URL path rather than query parameters. After applying this change, the snippets API works correctly.

Please let me know if you have any other questions!

### DIFF
--- a/snippets.go
+++ b/snippets.go
@@ -20,7 +20,6 @@ import (
 	"bytes"
 	"fmt"
 	"net/http"
-	"net/url"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


svanharmelen (approved):
Thanks 👍🏻

PR #1886, opened 2024-02-24
Add missing fields to SetDroneCIServiceOptions

jmajaca:
This PR adds a missing `push_events`, `merge_requests_events` and `tag_push_events` to `SetDroneCIServiceOptions`.

GitLab API docs: https://docs.gitlab.com/ee/api/integrations.html#set-up-drone

PR #1885, opened 2024-02-24
Jenkins integration enable ssl verification

jmajaca:
This PR adds a missing `EnableSSLVerification` to`SetJenkinsCIServiceOptions`.

GitLab API docs: https://docs.gitlab.com/ee/api/integrations.html#set-up-jenkins

PR #1887, opened 2024-02-28
Update GET response for jenkins and drone integrations

jmajaca:
I recently created https://github.com/xanzy/go-gitlab/pull/1886 and https://github.com/xanzy/go-gitlab/pull/1885 PRs.

But I missed that GET response also changed for those integrations, so I updated them now:
- removed `Token` from `DroneCIServiceProperties`
- added `EnableSSLVerification` to `JenkinsCIServiceProperties`
- updated GitLab API links for those two integrations

Also, I saw that other integrations/services are not in sync with current GitLab API and I would like to update them. Do you prefer me opening PR for each integration/service individually or grouping them in few PRs?

### DIFF
--- a/services.go
+++ b/services.go
@@ -383,7 +383,7 @@ func (s *ServicesService) DeleteDiscordService(pid interface{}, options ...Reque
 // DroneCIService represents Drone CI service settings.
 //
 // GitLab API docs:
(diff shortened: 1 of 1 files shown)


svanharmelen (approved):
Thanks @jmajaca 👍🏻

PR #1893, opened 2024-03-06
Add group_inheritance_type attribute to project/group protected environments

beekeep:
Added support for `group_inheritance_type` attribute to both project and group protected environment resources, specifically under `deploy_access_levels` where it was missing.

https://docs.gitlab.com/ee/api/protected_environments.html#group-inheritance-types


svanharmelen (approved):
LGTM @beekeep 👍🏻  Thanks!

PR #1905, opened 2024-03-29
Add support for "draft" notes

harrisoncramer:
This MR adds supports for draft notes, which are special notes that are part of a review and can be "published."

Documentation for this resource type is here: https://docs.gitlab.com/ee/api/draft_notes.html


harrisoncramer:
Implements #1905

PR #1906, opened 2024-04-01
Adding Cookie client for embed code into Gitlab site tree

aborche:
When you embed your code into gitlab tree, with this patch you can use all session cookies for API requests.

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -63,6 +63,7 @@ const (
 	JobToken
 	OAuthToken
 	PrivateToken
+	GitlabCookie
 )
 
 // A Client manages communication with the GitLab API.
@@ -87,6 +88,9 @@ type Client struct {
 
 	// Token type used to make authenticated API calls.
 	authType AuthType
+
+	// Auth Cookie for internal requests
+	cookie []*http.Cookie
 
 	// Username and password used for basic authentication.
 	username, password string
@@ -270,6 +274,18 @@ func NewBasicAuthClient(username, password string, options ...ClientOptionFunc) 
 	client.authType = BasicAuth
 	client.username = username
 	client.password = password
+
+	return client, nil
+}
+
+func NewCookieClient(cookie []*http.Cookie, options ...ClientOptionFunc) (*Client, error) {
+	client, err := newClient(options...)
+	if err != nil {
+		return nil, err
+	}
+
+	client.authType = GitlabCookie
+	client.cookie = cookie
 
 	return client, nil
 }
@@ -851,6 +867,13 @@ func (c *Client) Do(req *retryablehttp.Request, v interface{}) (*Response, error
 	case PrivateToken:
 		if values := req.Header.Values("PRIVATE-TOKEN"); len(values) == 0 {
 			req.Header.Set("PRIVATE-TOKEN", c.token)
+		}
+	case GitlabCookie:
+		sendCookie := c.cookie
+		if len(sendCookie) > 0 {
+			for _, value := range sendCookie {
+				req.Header.Add("Cookie", value.String())
+			}
 		}
 	}
 


svanharmelen:
Can you elaborate on the use case for this? Is gitlab tree a tool I don't know about? 😅 

Additionally I wonder if you could solve your use case using https://github.com/xanzy/go-gitlab/blob/main/client_options.go#L137? It allows you to pass in a func the would update each request. So I guess you could add a func here that would set the header you need/want?

aborche:
> Can you elaborate on the use case for this? Is gitlab tree a tool I don't know about? 😅

This is a dirty hack.
```
nginx['custom_gitlab_server_config'] = "location /deployhelper {\n proxy_pass http://localhost:8099;\n}\n"
``` 
after this you can embed any external code in your gitlab site tree and use gitlab api with current logged user context.
For example, you can automate some hard subroutines or embed own analytics portal which process many api requests in one page/step.

PS: also i know how to add a custom menu item in gitlab without contributing code in main gitlab tree. For example custom "Deploy Helper" application
image

svanharmelen:
Hmm... I see.. So isn't it possible to use https://github.com/xanzy/go-gitlab/blob/main/client_options.go#L137 instead to accomplish the same result?

I'm a bit worried that adding this (and people starting using it) might lead to issues that I cannot oversee. For example what happens when a cookie gets updated or changed after you created the client? There is no way the client would recover from that I guess.

So I think I would prefer to not add this as a client type, but that you instead enable this using the option func I mentioned.

aborche:
i have no idea how i can use _WithRequestOptions_. Can you provide an example for my case?
```go

func GetCtxBaseURL(c *gin.Context, config Config) string {
	scheme := "http"
	if c.Request.Header.Get("X-Forwarded-Proto") != "" {
		scheme = c.Request.Header.Get("X-Forwarded-Proto")
	}
	GitlabAPIHost := c.Request.Host
	if !config.GitlabUrlFromHost {
		GitlabAPIHost = config.CustomGitlabURL
	}
	ctxBaseURL := scheme + "://" + GitlabAPIHost + "/api/v4/"
	return ctxBaseURL
}

func GetCtxCookies(c *gin.Context) (GitlabCookies []*http.Cookie) {
	GitlabAuthCookies := []string{"known_sign_in", "_gitlab_session"}
	for _, value := range c.Request.Cookies() {
		if slices.Contains(GitlabAuthCookies, value.Name) {
			GitlabCookies = append(GitlabCookies, value)
		}
	}
	return GitlabCookies
}

func APIClient(c *gin.Context, cfg conf.Config) (*gitlab.Client, error) {
	ctxBaseUrl := conf.GetCtxBaseURL(c, cfg)
	ctxSessionCookies := conf.GetCtxCookies(c)
	git, err := gitlab.NewCookieClient(ctxSessionCookies, gitlab.WithBaseURL(ctxBaseUrl))
	if err != nil {
		return nil, fmt.Errorf("Failed to create client: %v", err)
	}
	if len(ctxSessionCookies) < 2 {
		return nil, errors.New("no requested cookies found")
	}
	return git, nil
}

func GetUserInfo(c *gin.Context, cfg conf.Config) (*gitlab.User, error) {
	git, err := APIClient(c, cfg)
	if err != nil {
		return nil, err
	}
	users, _, err := git.Users.CurrentUser()
	if err != nil {
		return nil, err
	}
	return users, err
}

```

aborche:
> Hmm... I see.. So isn't it possible to use https://github.com/xanzy/go-gitlab/blob/main/client_options.go#L137 instead to accomplish the same result?

No, we can't use clientOptions. When we work with cookie methods, no other authentication methods can be used. The go-gitlab client does not have initial authentication information from the user. It uses existing authentication tokens from the upstream request.

RicePatrick:
Even if no other authentication method is supported, you should still be able to use `WithRequestOptions` to remove the existing authentication header and add your cookies. Here's an example of doing so:

```golang

client, err := gitlab.NewOAuthClient("", gitlab.WithRequestOptions(func(req *retryablehttp.Request) error {
		// Remove "Authorization" header
		req.Header.Del("Authorization")

		// Add a new cookie instead
		cookie := &http.Cookie{
			Name:     "gitlab_token",
			Value:    c.Token,
			Path:     "/",
			HttpOnly: true,
			Secure:   true,
		}
		req.AddCookie(cookie)

		return nil
	}))
```

You would need to tailor what you're removing/adding, but every API call would have that function run on it before making the call which should inject the cookie you need.

I agree with Sander that supporting this in the base client is not something that we should really add since you've need to use a separate authentication method normally to _get_ a cookie value, with you can then use for subsequent calls, making it a non-standard programmatic way of interacting with the API. Instances where that isn't true would be something like yours where you have a "wrapper" around GitLab, which are difficult for us to support.

I'll go ahead and close the PR for now, and we can discuss under and issue if you'd like, but I think the above should work for you!

PR #1909, opened 2024-04-03
Add support for EmailsEnabled to groups

PatrickRice-KSC:
This PR adds support for the `emails_enabled` attribute to the Group APIs. This allows us to support that attribute in the Terraform provider prior to its removal in GitLab 17.0


PatrickRice-KSC:
@svanharmelen  - This is ready for review. If you wouldn't mind cutting a new tag after this is merged so that we can update the TF provider, I'd appreciate it :)

svanharmelen (approved):
Thanks, look good 👍🏻

PR #1913, opened 2024-04-11
New events webhooks

Poulpatine:
Hi,

I'm adding support for new events webhooks:
- Resource Access Token events project and group webhook: https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html#project-and-group-access-token-events
- Member Events group webhook: https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html#group-member-events

Thanks.

PR #1916, opened 2024-04-15
Add new project and group access token events to webhook event types

vntw:
👋 

This PR complements #1913 by also supporting the new events when parsing webhooks.

**Docs/Examples:** https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html#project-and-group-access-token-events

Let me know if you have any questions :)

Thanks!

### DIFF
--- a/event_parsing.go
+++ b/event_parsing.go
@@ -27,24 +27,25 @@ type EventType string
 
 // List of available event types.
 const (
-	EventConfidentialIssue EventType = "Confidential Issue Hook"
-	EventConfidentialNote  EventType = "Confidential Note Hook"
-	EventTypeBuild         EventType = "Build Hook"
-	EventTypeDeployment    EventType = "Deployment Hook"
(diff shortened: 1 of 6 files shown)


event_parsing.go
```diff
@@ -252,6 +253,20 @@ func ParseWebhook(eventType EventType, payload []byte) (event interface{}, err e
...
 	case EventTypeRelease:
 		event = &ReleaseEvent{}
+	case EventTypeResourceAccessToken:
+		data := map[string]interface{}{}
+		err := json.Unmarshal(payload, &data)
+		if err != nil {
+			return nil, err
+		}
+
+		if _, ok := data["project"]; ok {
```
vntw, line 263: The payload is mostly the same for both types. Depending on the actual type, either `group_id`/`group` or `project_id`/`project` will be present. A bit of a change compared to other events… 🤷

svanharmelen (approved):
Thanks @vntw 👍🏻

PR #1919, opened 2024-04-16
Ignore response body during HEAD request error handling

### LINKED ISSUE #1047: GetFileMetaData errors on missing file
When calling the `RepositoryFilesService.GetFileMetaData()` function on a file that doesn't exist, but the request was successful, I expect the returned `*File` and `error` to be nil.
Instead I get an `ErrorResponse` containing `HEAD http://127.0.0.1:49996/api/v4/projects/1234/repository/files/foobar.yaml: 404 failed to parse unknown error format`.

This is due to the fact that CheckResponse() (called by Client.Do(), called by RepositoryFilesService.GetFileMetaData()) tries  to parse the empty body as JSON.
I think it fails here:
```
data, err := ioutil.ReadAll(r.Body)
	if err == nil && data != nil {
```
`data` doesn't seem to be nil here, maybe it contains `\r\n`? As I can see, this endpoint that came with #511 is the only one within the codebase using HEAD, which is probably why it never popped up before.
Nevertheless I think that a HEAD request that returns 404 should either be not an error or some specific error (e.g. NotFoundError) to be catchable.

vntw:
Fixes #1047

This improves the error handling of a HEAD request response. It avoids trying to parse the body entirely. To handle the error, you can still check for a `*gitlab.ErrorResponse`:

```go
_, _, err := gitlabClient.RepositoryFiles.GetFileMetaData(1, "not-there.txt", &gitlab.GetFileMetaDataOptions{Ref: gitlab.Ptr("main")})
if err != nil {
	var errResp *gitlab.ErrorResponse
	if errors.As(err, &errResp) {
		log.Println(errResp.Response.StatusCode)
	} else {
		log.Println(err)
	}
}
```

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -953,8 +953,13 @@ type ErrorResponse struct {
 
 func (e *ErrorResponse) Error() string {
 	path, _ := url.QueryUnescape(e.Response.Request.URL.Path)
-	u := fmt.Sprintf("%s://%s%s", e.Response.Request.URL.Scheme, e.Response.Request.URL.Host, path)
-	return fmt.Sprintf("%s %s: %d %s", e.Response.Request.Method, u, e.Response.StatusCode, e.Message)
+	url := fmt.Sprintf("%s://%s%s", e.Response.Request.URL.Scheme, e.Response.Request.URL.Host, path)
+	str := fmt.Sprintf("%s %s: %d", e.Response.Request.Method, url, e.Response.StatusCode)
+	if e.Message != "" {
+		str += " " + e.Message
+	}
+
+	return str
 }
 
 // CheckResponse checks the API response for errors, and returns them if present.
@@ -965,6 +970,11 @@ func CheckResponse(r *http.Response) error {
 	}
 
 	errorResponse := &ErrorResponse{Response: r}
+
+	if r.Request.Method == http.MethodHead {
+		return errorResponse
+	}
+
 	data, err := io.ReadAll(r.Body)
 	if err == nil && data != nil {
 		errorResponse.Body = data
--- a/gitlab.go
+++ b/gitlab.go
@@ -21,6 +21,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"math/rand"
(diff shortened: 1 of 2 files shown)


gitlab.go
```diff
@@ -965,6 +970,11 @@ func CheckResponse(r *http.Response) error {
 	}
 
 	errorResponse := &ErrorResponse{Response: r}
+
+	if r.Request.Method == http.MethodHead {
+		return errorResponse
+	}
```
svanharmelen, line 976: I fail to see how this fixes #1047? This will still return an error, the only difference being that the additional parsing error message is not added.

The main suggestion in #1047 was either not returning an error at all, of returning a specific error which could be found with things like `errors.Is()`.

Any additional feedback or insights on that?

vntw: Hm maybe I misunderstood but this returns exactly what the GitLab API returns but without the futile parsing attempt and the misleading error message (which in my eyes was the main issue).

I don't see why a `HEAD` request should behave differently for this use case/status code or why there's even a need for a new error type (what's the benefit?) – and alternatively, if this returns `nil` for both values, how do I know what happened? 

The aim of this PR is to allow for the same predictable error handling that other resources share.

Sorry if I don't get it but that's my reasoning for it, the proposed solutions just didn't make sense to me.

svanharmelen: See the commit I just pushed. I think that is a (somewhat) similar solution which _does_ give users to distinguish between a 404 error and a functional/logical error.

The updated check to decide to parse the bode or not should fix issues with "empty" bodies that do not contain readable characters.

vntw: Oh yea thanks for your time and the example, I got what you meant with the custom error, I just don't see a reason to introduce this when everything that's needed to handle these cases is already in the library.

But I don't want to take up more of your time with basically the same argument. If you feel that the custom type approach is better, please do go ahead and use it (or close this PR and see if someone has a better approach in the future, totally fine) 🙂 

You're _a little bit_ more familiar with the library and what does/doesn't fit in the picture 😄

svanharmelen: I get your point, but I'll go ahead and merge this anyway as it does make checking for 404 errors a lot easier and more idiomatic without causing any backwards incompatibility issues.

PR #1923, opened 2024-04-26
Add missing attribute to group push rules

beekeep:
The `commit_committer_name_check` was missing from the group push rules.  Similar to the recent update made to project push rules.


svanharmelen (approved):
Thanks 👍🏻

PR #1924, opened 2024-04-30
Add SAML group link support for custom roles

heidiberry:
As a GitLab Ultimate customer, we would like to be able to use the custom role feature through SAML and terraform. This change intends to add the ID for the custom role (member_role_id) as an option to the List, Get and Add Group SAML Links commands. Our intention is to then add support to the terraform provider.


svanharmelen (approved):

svanharmelen:
Thanks @heidiberry 👍🏻  LGTM!

PR #1925, opened 2024-05-02
Update integrations 1/2

jmajaca:
I updated first 10 (out of 20) integrations that are currently implemented in the source code - [Integrations API](https://docs.gitlab.com/ee/api/integrations.html).

`gitlab-ci` and `hipchat` are no longer in the API docs so I deleted them.

My idea is to update remaining 10 in the next PR.

### DIFF
--- a/services.go
+++ b/services.go
@@ -61,7 +61,7 @@ type Service struct {
 
 // ListServices gets a list of all active services.
 //
-// GitLab API docs: https://docs.gitlab.com/ee/api/integrations.html#list-all-active-services
+// GitLab API docs: https://docs.gitlab.com/ee/api/integrations.html#list-all-active-integrations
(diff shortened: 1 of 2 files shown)


svanharmelen (changes requested):
Hi @jmajaca, thanks for your PR. Looks good, except for removing the GitLabCI and HipChat services.

Removing those means that people who use older versions of GitLab can no longer update this package.

So I prefer to keep those services in for now...

jmajaca:
@svanharmelen I returned deleted services back to the source code. 

I also left URLs in function docs unchanged (and broken) because the oldest official gitlab docs archive is version 14.7 and it seems that those services were removed before that.

svanharmelen (approved):
Thanks @jmajaca! LGTM 👍🏻

PR #1927, opened 2024-05-03
Resolves #1405: only one endpoint was missing

### LINKED ISSUE #1405: External status check: creation calls
I see from #1261 that status checks has been implemented. But only the listing calls seems to be available.

There is other calls that probably could be useful to create external checks integrations

https://docs.gitlab.com/ee/api/status_checks.html#set-status-of-an-external-status-check

If there is no objections I may try to implement them

heidiberry:
Comparing the endpoints to the documentation here: https://docs.gitlab.com/ee/api/status_checks.html there is only one endpoint not currently supported. This is https://docs.gitlab.com/ee/api/status_checks.html#retry-failed-status-check-for-a-merge-request

I've added a new function for this endpoint and a very basic test.

### DIFF
--- a/external_status_checks.go
+++ b/external_status_checks.go
@@ -198,27 +198,18 @@ func (s *ExternalStatusChecksService) UpdateExternalStatusCheck(pid interface{},
 	return s.client.Do(req, nil)
 }
 
-// RetryFailedStatusCheckForAMergeRequestOptions represents the available
-// RetryFailedStatusCheckForAMergeRequest() options.
-//
-// GitLab API docs:
-// https://docs.gitlab.com/ee/api/status_checks.html#retry-failed-status-check-for-a-merge-request
-type RetryFailedStatusCheckForAMergeRequestOptions struct {
-	ExternalStatusCheckID *int `url:"external_status_check_id,omitempty" json:"external_status_check_id,omitempty"`
-}
-
 // UpdateExternalStatusCheck updates an external status check.
 //
 // Gitlab API docs:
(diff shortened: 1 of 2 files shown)


heidiberry:
https://github.com/xanzy/go-gitlab/issues/1405

external_status_checks.go
```diff
@@ -197,3 +197,31 @@ func (s *ExternalStatusChecksService) UpdateExternalStatusCheck(pid interface{},
...
+// https://docs.gitlab.com/ee/api/status_checks.html#retry-failed-status-check-for-a-merge-request
+type RetryFailedStatusCheckForAMergeRequestOptions struct {
+	ExternalStatusCheckID *int `url:"external_status_check_id,omitempty" json:"external_status_check_id,omitempty"`
+}
+
+// UpdateExternalStatusCheck updates an external status check.
+//
+// Gitlab API docs:
+// https://docs.gitlab.com/ee/api/status_checks.html#retry-failed-status-check-for-a-merge-request
+func (s *ExternalStatusChecksService) RetryFailedStatusCheckForAMergeRequest(pid interface{}, mergeRequest int, opt *RetryFailedStatusCheckForAMergeRequestOptions, options ...RequestOptionFunc) (*Response, error) {
```
svanharmelen, line 214: All the params that are part of the URL are passed as function parameters. The option structs are only used for sending options to the API which in this case is not applicable as all params are art of the URL and not request body is expected.

svanharmelen (changes requested):

svanharmelen (approved):
Thanks @heidiberry, LGTM 👍🏻

PR #1930, opened 2024-05-03
Fix: Sync project/group resource access token webhook events to match current documentation

vntw:
**Update:** After raising [this issue](https://gitlab.com/gitlab-org/gitlab/-/issues/460466), [this merge request](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/152907) updated the webhook payloads that we can expect.

This PR updates the payloads and structs to reflect that ✌️ 

---

I searched for other places in the code where a format like `2024-01-24 16:27:40 UTC` occurs and found (among others) this which I used as a reference:

https://github.com/xanzy/go-gitlab/blob/2c4b565c47917bc633bc2b0521802a0630b76f83/event_webhook_types.go#L832-L833

Hope that this is the correct way! Let me know if it's not

### DIFF
--- a/event_parsing_webhook_test.go
+++ b/event_parsing_webhook_test.go
@@ -160,8 +160,10 @@ func TestParseGroupResourceAccessTokenHook(t *testing.T) {
 		t.Errorf("Expected GroupResourceAccessTokenEvent, but parsing produced %T", parsedEvent)
 	}
 
-	if event.GroupID != 35 {
-		t.Errorf("GroupID is %v, want %v", event.GroupID, 35)
+	expectedEventName := "expiring_access_token"
+
+	if event.EventName != expectedEventName {
+		t.Errorf("EventName is %v, want %v", event.EventName, expectedEventName)
 	}
 }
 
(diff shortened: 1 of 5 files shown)


svanharmelen:
Hmm... Yeah the inconsistencies around how time is used in the API is quite frustrating. In this case I guess it would make sense to first open an issue with GitLab to see if it's a bug there which they need to fix instead?

vntw:
Good idea, I'll create an issue (if there hasn't been one already) and report back

vntw:
> In this case I guess it would make sense to first open an issue with GitLab to see if it's a bug there which they need to fix instead?

Thanks again for the hint, turns out there was more to it 😃 I updated the PR with the necessary links and info.

I think this is ready now!

svanharmelen (approved):
It's a shame the used time formats are such a mess. We could introduce a `UTCTime` struct like the `ISOTime` struct to enable parsing it as a `time.Time`, but for now I'm not going to bother and it could be an annoying change to existing users of the API.

So thanks for your efforts 👍🏻

Issue #1928, opened 2024-05-03
Missing `Status` field for container registry repositories


RaveNoX:
Hello,

According to https://docs.gitlab.com/ee/api/container_registry.html#list-registry-repositories

Status field should be present for container registry repository object, but it's not present in this library.

PR #1933, opened 2024-05-04
Add container registry status to response body

heidiberry:
Resolves https://github.com/xanzy/go-gitlab/issues/1928

Added the status field to the container repository object. The API documentation does not include a list of valid statuses but I think I've found the correct enum in GitLab's codebase and linked to it in the type declaration.


svanharmelen (approved):
Thanks @heidiberry 👍🏻

PR #1935, opened 2024-05-06
Reinstate missing fields for emails on push integration

heidiberry:
These fields were removed under https://github.com/xanzy/go-gitlab/pull/1925 but are still present in the API documentation https://docs.gitlab.com/ee/api/integrations.html#emails-on-push.

Adding them back in as updating to the latest tag causes compilation errors in the GitLab Terraform Provider https://gitlab.com/gitlab-org/terraform-provider-gitlab/-/blob/main/internal/provider/sdk/resource_gitlab_integration_emails_on_push.go?ref_type=heads#L177


svanharmelen (approved):
Guess I missed these fields being deleted as I normally try to preserve backwards compatibility...

PR #1938, opened 2024-05-11
GitLab 17.0: Fix an issue where Settings errors when ContainerRegistry is not enabled

PatrickRice-KSC:
In the 17.0-nightly build, the ContainerRegistryImportCreatedBefore attribute now returns "" when the container registry isn't enabled. This causes a parsing error because an empty string isn't parseable as a date in Go. As a result, `GetSettings` fails on every instance without a container registry enabled.

This PR implements custom parsing logic for the `Settings` struct that discards that value if it's set to an empty string to ensure that the Settings can continue to parse properly.

You can see this being detected in the Terraform Provider Nightly tests here: https://gitlab.com/gitlab-org/terraform-provider-gitlab/-/jobs/6827815421#L1850

### DIFF
--- a/settings.go
+++ b/settings.go
@@ -393,41 +393,29 @@ type Settings struct {
 	UserEmailLookupLimit int `json:"user_email_lookup_limit"`
 }
 
-// Settings requires a custom unmarshaller because the `container_registry_import_created_before` attribute (which is time.Time)
-// returns as "" when the registry has not been enabled, which is not parseable as a date. This causes an error any time
-// GetSettings is called on an instance with the registry disabled.
-// To prevent this error, Settings has a custom UnmarshalJson that nils out the empty string value prior to parsing.
+// Settings requires a custom unmarshaller in order to properly unmarshal
+// `container_registry_import_created_before` which is either a time.Time or
+// an empty string if no value is set.
 func (s *Settings) UnmarshalJSON(data []byte) error {
-	// Ignore null, like in the main JSON package.
-	if string(data) == "null" || string(data) == `""` {
-		return nil
-	}
+	type Alias Settings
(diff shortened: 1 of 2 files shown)


PatrickRice-KSC:
@svanharmelen  - This should be ready for review!

To be honest, I don't _love_ that we have to implement this here because it seems almost like a data type issue in the GitLab API. I'll chat with GitLab about it as well, but I think it's safe and relatively innocuous to handle it here as well. 

I also debated using a string replace to remove the value instead of using the unmarshal -> remarshal approach, since the double marshal is "expensive" (relatively speaking) from a CPU perspective. I landed on doing the two step marshal instead because I think it's easier to grok from a readability perspective, and I think it's overall _safer_ from a deterministic behavior perspective. Let me know if you think the other approach would be preferable though.

If you agree and merge, would you mind cutting a new version? That way I can see if the nightly builds get fixed on the TF provider :)

PatrickRice-KSC:
Update from discussions with the GitLab team - this field is only intended to be used on gitlab.com, so it's removal in 17.0 is intentional. We could either remove the field from the struct completely, or handle parsing the blank value (like this PR will do): https://gitlab.com/gitlab-org/gitlab/-/merge_requests/149168

I'd _lean_ towards handling the blank value in case someone is using those fields (and marking them as deprecated so people can remove them over the next couple of months), but I'm interested in your thoughts!

svanharmelen (approved):
I simplified the solution just a little and also removed the test to check if unmarshalling an empty string would fail.

`Settings` is an object so it would be a JSON error if you try to parse `""` into a `Settings` object (should always be `null`, which I did test and works as expected).

PR #1939, opened 2024-05-11
Add ability to import Github repos

heidiberry:
Adds a function which imports a Github repository into GitLab.

See https://docs.gitlab.com/ee/api/import.html#import-repository-from-github

### DIFF
--- a/import.go
+++ b/import.go
@@ -29,11 +29,11 @@ type ImportService struct {
 	client *Client
 }
 
-// GithubImport represents the response from an import from Github.
+// GitHubImport represents the response from an import from GitHub.
 //
 // GitLab API docs:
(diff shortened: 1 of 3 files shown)


import.go
```diff
@@ -0,0 +1,90 @@
...
+// https://docs.gitlab.com/ee/api/import.html
+type ImportService struct {
+	client *Client
+}
+
+// GithubImport represents the response from an import from Github.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/import.html#import-repository-from-github
+type GithubImport struct {
```
svanharmelen, line 36: I think we should rename `Github` to `GitHub` (in all related locations)

svanharmelen (approved):
Thanks @heidiberry 👍🏻

Issue #1479, opened 2022-05-20
Add support for "resource group" API


MichaelGissingNC:
Resource groups are kind of a semaphore to prevent multiple deployments to the same environment. If there are multiple deployment jobs queued there is a setting `process_mode` to control the order in which these jobs are processed. There is no GUI to change this setting, it's only available via API.

Therefore it would be great to have this API available in the library.

* Feature Documentation: https://docs.gitlab.com/ee/ci/resource_groups/index.html
* API Documentation: https://docs.gitlab.com/ee/api/resource_groups.html

MichaelGissingNC:
@svanharmelen Suppose I would find time to work on this (which I can't promise yet) - Is there a contribution guide or similar?

svanharmelen:
Not really, other then that I prefer to follow the naming and ordering from the docs as much as possible. And it's probably good to have a quick look around the existing code and make sure any new code is consistent with that...

PR #1940, opened 2024-05-14
Add resource group API

heidiberry:
Adds support for the Resource Group API: https://docs.gitlab.com/ee/api/resource_groups.html

Resolves https://github.com/xanzy/go-gitlab/issues/1479

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -206,6 +206,7 @@ type Client struct {
 	Repositories                 *RepositoriesService
 	RepositoryFiles              *RepositoryFilesService
 	RepositorySubmodules         *RepositorySubmodulesService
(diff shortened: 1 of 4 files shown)


svanharmelen (approved):
I'm usually a bit hold back when I see a somewhat bigger PR, as usually that means I have to polish things myself and/or give a lot of feedback.

But I must say you delivered a perfect PR in which you seem to understand exactly how I prefer to structure and order things (which isn't that difficult as there is a lot of existing code that already shows that, but you would be surprised how few PRs are consistent with the existing code base 😉).

So thanks for that and for this awesome PR @heidiberry 👍🏻

Issue #1946, opened 2024-05-25
Snippets struct is missing `project_id` field


gabrie30:
Snippets struct [here](https://github.com/xanzy/go-gitlab/blob/main/snippets.go#L37) is missing a `ProjectID` field that should map to the returned json of `project_id` in the gitlab response from `GET /snippets/all`

The gitlab snippets documentation includes this field in the example responses [here](https://docs.gitlab.com/ee/api/snippets.html#list-all-snippets)

PR https://github.com/xanzy/go-gitlab/pull/1947

PR #1947, opened 2024-05-25
Add ProjectID to snippets struct

gabrie30:
Fixes https://github.com/xanzy/go-gitlab/issues/1946

### DIFF
--- a/snippets.go
+++ b/snippets.go
@@ -50,6 +50,7 @@ type Snippet struct {
 	} `json:"author"`
 	UpdatedAt *time.Time `json:"updated_at"`
 	CreatedAt *time.Time `json:"created_at"`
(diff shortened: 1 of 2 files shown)


snippets.go
```diff
@@ -57,6 +57,7 @@ type Snippet struct {
 		RawURL string `json:"raw_url"`
 	} `json:"files"`
 	RepositoryStorage string `json:"repository_storage"`
+	ProjectID         int    `json:"project_id"`
```
svanharmelen, line 60: Please follow the ordering used in the docs (https://github.com/xanzy/go-gitlab/blob/main/CONTRIBUTING.md).

gabrie30: Thanks I've updated the order to match the docs.

PR #1950, opened 2024-06-05
Support `action` property in all Note webhooks

vntw:
This was introduced in GitLab 16.11, see https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html#comment-events

### DIFF
--- a/event_webhook_types.go
+++ b/event_webhook_types.go
@@ -102,21 +102,22 @@ type CommitCommentEvent struct {
 	} `json:"project"`
 	Repository       *Repository `json:"repository"`
 	ObjectAttributes struct {
-		ID           int    `json:"id"`
-		Note         string `json:"note"`
-		NoteableType string `json:"noteable_type"`
-		AuthorID     int    `json:"author_id"`
-		CreatedAt    string `json:"created_at"`
-		UpdatedAt    string `json:"updated_at"`
-		ProjectID    int    `json:"project_id"`
-		Attachment   string `json:"attachment"`
-		LineCode     string `json:"line_code"`
-		CommitID     string `json:"commit_id"`
-		NoteableID   int    `json:"noteable_id"`
-		System       bool   `json:"system"`
-		StDiff       *Diff  `json:"st_diff"`
-		Description  string `json:"description"`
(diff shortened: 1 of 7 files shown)


event_webhook_types_test.go
```diff
@@ -101,6 +101,10 @@ func TestCommitCommentEventUnmarshal(t *testing.T) {
 		t.Errorf("NoteableType is %v, want %v", event.ObjectAttributes.NoteableType, "Commit")
 	}
 
+	if event.ObjectAttributes.Action != "create" { //nolint:goconst
```
vntw, line 104: `//nolint` exists because if there are >2 usages of the `"create"` string, the linter will complain about
```
string `create` has 3 occurrences, but such constant `FileCreate` already exists (goconst)
```
Using `FileCreate` is the wrong intent for me, please let me know how you want that handled. I could also add a description why that is in the code

svanharmelen: Maybe create a const for it? Something like:

```go
type CommentEventAction string

const (
  CommentEventActionCreate CommentEventAction = "create"
  CommentEventActionUpdate CommentEventAction = "update"
)
```

vntw: Thanks for the idea and snippet! I also used the action struct in the event struct. Let me know if this should live somewhere else, I used `StateID` as a reference

event_webhook_types.go
```diff
@@ -38,6 +38,17 @@ const (
...
+// CommentEventAction identifies if a comment has been newly created or updated.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html#comment-events
+type CommentEventAction string
+
+const (
+	CommentEventActionCreate CommentEventAction = "create"
+	CommentEventActionUpdate CommentEventAction = "update"
+)
```
svanharmelen, line 50: Looks great, thanks! Can you move this snippet to the `types.go` file and insert it in its alphabetical order in the file?

vntw: done 👍

svanharmelen (approved):
Thanks @vntw! LGTM 👍🏻

PR #1686, opened 2023-04-08
feat: Add support for getting a single personal access token

ErKiran:
added two endpoints to get the single personal access token
* self
* token by id https://github.com/xanzy/go-gitlab/issues/1684

PR #1954, opened 2024-06-10
Fix `GetSinglePersonalAccessTokenByID` arg name

theipster:
Calling the token ID `user` is _extremely_ confusing...!

Relates to #1686.

### DIFF
--- a/personal_access_tokens.go
+++ b/personal_access_tokens.go
@@ -83,8 +83,8 @@ func (s *PersonalAccessTokensService) ListPersonalAccessTokens(opt *ListPersonal
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/personal_access_tokens.html#using-a-personal-access-token-id
-func (s *PersonalAccessTokensService) GetSinglePersonalAccessTokenByID(user int, options ...RequestOptionFunc) (*PersonalAccessToken, *Response, error) {
-	u := fmt.Sprintf("personal_access_tokens/%d", user)
+func (s *PersonalAccessTokensService) GetSinglePersonalAccessTokenByID(token int, options ...RequestOptionFunc) (*PersonalAccessToken, *Response, error) {
+	u := fmt.Sprintf("personal_access_tokens/%d", token)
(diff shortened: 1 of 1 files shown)


svanharmelen (changes requested):
I can appreciate the `user` -> `token` change (makes sense), but I do not like to ignore error checking.

In my opinion you shoud always check errors, also in tests. So please revert those changes.

Updating the error message to make it more descriptive and applicable to the actual error is of course fine.

theipster:
@svanharmelen Thanks for the quick response! 🙂 

> In my opinion you shoud always check errors, also in tests. So please revert those changes.
> 
> Updating the error message to make it more descriptive and applicable to the actual error is of course fine.

I'm happy to update my PR, but can you please explain what the purpose is / what error you might expect to see in these specific lines of code, given that it's essentially calling a standard library function with fixed arguments (so the return value will always be fixed)? Error checking would make sense if the arguments were variable, but that's not the case here.

Also, the `TestRotatePersonalAccessToken` test doesn't have error checking either.

svanharmelen:
I'm not going to spent too much words on it, but I believe that you should always check all your errors everywhere. As soon as you start to reason about which error checks make sense and which do not, you'll be on a slippery slope just waiting for a missed error check to bites you in production when you least expect it.

Next to that it will set a precedent for other users or committers of this package or maybe even to people who are new at Go. I prefer to encourage them to get used to just always check all errors as in my opinion the Go language best practices also encourages you to do so.

If `TestRotatePersonalAccessToken` doesn't do error checking then that one slipped through unintentionally. So feel free to update that test instead in order to fix it and make everything consistent again.

theipster:
@svanharmelen I've now removed the changes regarding the tests.

As a general best practice, I understand and agree with your intentions. However, in this particular case, I'm struggling to articulate what the error message should actually be (and `PersonalAccessTokens.ListPersonalAccessTokens returned error:"` is definitely not accurate because it has nothing to do with the `ListPersonalAccessTokens` function), so I'll leave it for somebody smarter than me to rename the error message.

svanharmelen (approved):
Thanks @theipster 👍🏻

PR #1957, opened 2024-06-18
Add ExtendedTrailers to snippets struct

yecaowulei:
The version of gitlab I am using is v17.0.1, and while modifying the gitlab file using the go-gitlab library, I encountered an error "json: cannot unmarshal array into Go value of type gitlab.Commit",Screening is found that the new gitlab interface/projects / : id/repository/commits Extended_trailers increases the field, so you need to in commits. Go in the structure of the Commit increase to solve this problem. See https://docs.gitlab.com/ee/api/commits.html gitlab document

### DIFF
--- a/commits.go
+++ b/commits.go
@@ -52,6 +52,7 @@ type Commit struct {
 	LastPipeline   *PipelineInfo     `json:"last_pipeline"`
 	ProjectID      int               `json:"project_id"`
 	Trailers       map[string]string `json:"trailers"`
+	Extended_trailers map[string]string `json:"extended_trailers"`
 	WebURL         string            `json:"web_url"`
 }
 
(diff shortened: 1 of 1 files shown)


svanharmelen (changes requested):
Please make sure you use Go style field names (check the rest of the package as well and please follow the used and expected naming conventions). This would mean calling the field `ExtendedTrailers` instead of`Extended_trailers`.

And also make sure to format your code (using [gofumpt](https://github.com/mvdan/gofumpt)). Thanks!

yecaowulei:
> Please make sure you use Go style field names (check the rest of the package as well and please follow the used and expected naming conventions). This would mean calling the field `ExtendedTrailers` instead of`Extended_trailers`.
> 
> And also make sure to format your code (using [gofumpt](https://github.com/mvdan/gofumpt)). Thanks!

I'm sorry, I didn't notice this naming problem, it's been fixed

svanharmelen:
I see that you fixed the naming, thanks! But it still needs to be formatted as well before I can merge it...

svanharmelen (approved):
LGTM! Thanks @yecaowulei 👍🏻

PR #1963, opened 2024-06-25
Add LFS field to GetRawFileOptions

royiarchy:
Small PR adding a missing field to the `GetRawFileOptions` struct.

https://docs.gitlab.com/ee/api/repository_files.html#get-raw-file-from-repository
gitlab lfs field


svanharmelen (approved):
Thanks @royiarchy 👍🏻  LGTM!

PR #1969, opened 2024-07-09
fix: Change VariableType type from string to VariableTypeValue

perrze:
Hi,
I changed the type of VariableType in pipelines.go, jobs.go and pipeline_schedules.go PipelineVariable Object from string to VariableTypeValue.
Indeed in group_variables.go line 123 the CreateGroupVariableOptions had this type for VariableType for example and I didn't find a reason why it would be different.

It passes test and gofumpt. 
I hope it will be useful !
Have a nice day


svanharmelen (approved):
Thanks @perrze 👍🏻

PR #1975, opened 2024-07-24
chore: add report type to MergeRequestApprovalRule

mdevilliers:
This attribute has been added to the approval_rules object - 

https://docs.gitlab.com/ee/api/merge_request_approvals.html#get-merge-request-level-rules

It should be returned by the enterprise edition and was added to gitlab with this PR - https://gitlab.com/gitlab-org/gitlab/-/merge_requests/155772.
That functionality was required as part of this PR against the terraform provider - https://gitlab.com/gitlab-org/terraform-provider-gitlab/-/merge_requests/1913

Happy to add a test but was wondering about it being an EE only feature?


svanharmelen (approved):
Thanks @mdevilliers 👍🏻

PR #1977, opened 2024-07-29
Support more options for list PATs

hjweddie:
In my production environment, hope that we can support more options for list PATs.
I have run my branch in production for a week, and it runs well yet

### DIFF
--- a/personal_access_tokens.go
+++ b/personal_access_tokens.go
@@ -57,7 +57,14 @@ func (p PersonalAccessToken) String() string {
 // https://docs.gitlab.com/ee/api/personal_access_tokens.html#list-personal-access-tokens
 type ListPersonalAccessTokensOptions struct {
 	ListOptions
-	UserID *int `url:"user_id,omitempty" json:"user_id,omitempty"`
+	CreatedAfter   *time.Time `url:"created_after,omitempty" json:"created_after,omitempty"`
+	CreatedBefore  *time.Time `url:"created_before,omitempty" json:"created_before,omitempty"`
+	LastUsedAfter  *time.Time `url:"last_used_after,omitempty" json:"last_used_after,omitempty"`
+	LastUsedBefore *time.Time `url:"last_used_before,omitempty" json:"last_used_before,omitempty"`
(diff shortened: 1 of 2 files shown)


personal_access_tokens.go
```diff
@@ -57,7 +57,14 @@ func (p PersonalAccessToken) String() string {
 // https://docs.gitlab.com/ee/api/personal_access_tokens.html#list-personal-access-tokens
 type ListPersonalAccessTokensOptions struct {
 	ListOptions
-	UserID *int `url:"user_id,omitempty" json:"user_id,omitempty"`
+	CreatedAfter   *time.Time `url:"created_after,omitempty" json:"created_after,omitempty"`
+	CreatedBefore  *time.Time `url:"created_before,omitempty" json:"created_before,omitempty"`
+	LastUsedAfter  *time.Time `url:"last_used_after,omitempty" json:"last_used_after,omitempty"`
+	LastUsedBefore *time.Time `url:"last_used_before,omitempty" json:"last_used_before,omitempty"`
```
svanharmelen, line 63: I guess these `*time.Time` fields should be `*ISOTime` looking at the docs. Can you check/confirm?

personal_access_tokens.go
```diff
@@ -57,7 +57,14 @@ func (p PersonalAccessToken) String() string {
...
+	Revoked        *bool      `url:"revoked,omitempty" json:"revoked,omitempty"`
+	Search         *string    `url:"search,omitempty" json:"search,omitempty"`
+	State          *string    `url:"state,omitempty" json:"state,omitempty"`
+	UserID         *int       `url:"user_id,omitempty" json:"user_id,omitempty"`
```
svanharmelen, line 67: Think this should be an `interface{}` as it can be either a string or an in. Not sure if we have that elsewhere and already have something for that more specific than an `interface{}`...

personal_access_tokens.go
hjweddie: I check gitlab api document and confirm that you are right，we need to use *ISOTime instead of *time.Time
I have change the type

svanharmelen (approved):
I'll take it, thanks!

PR #1978, opened 2024-07-30
Add `ContainerRegistrySize` field to `Statistics` struct

Florian3535:
The `ContainerRegistrySize` field is missing from the `Statistics`  struct.

The public GitLab API returns the `container_registry_size` field for project data. E.g. https://docs.gitlab.com/ee/api/projects.html#list-user-projects.

The public GitLab API does not return the `container_registry_size` field for group data. Thus, this change would cause that we always return the null value for `ContainerRegistrySize` when fetching groups, which is not optimal. However, we already always return the null value for the `commit_count` field in the `Statistics`  struct when fetching groups, as we use the same `Statistics`  struct for project and group data.

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -292,6 +292,7 @@ type Statistics struct {
 	PackagesSize          int64 `json:"packages_size"`
 	SnippetsSize          int64 `json:"snippets_size"`
 	UploadsSize           int64 `json:"uploads_size"`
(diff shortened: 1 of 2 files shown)


svanharmelen (approved):
Thanks @Florian3535  👍🏻

PR #1981, opened 2024-08-07
Add field "is_auditor" in user

thomasboni:


svanharmelen (approved):
Thanks @thomasboni 👍🏻

users_test.go
```diff
@@ -103,6 +103,7 @@ func TestGetUserAdmin(t *testing.T) {
 		LastActivityOn:   &lastActivityOn,
 		ColorSchemeID:    2,
 		IsAdmin:          true,
+		IsAuditor:        true,
```
svanharmelen, line 106: Guess you didn't test this 😞 Will add a commit to fix it...

thomasboni: Sorry, in my case it worked because I used an administrator user. I haven't seens the [different behavior on this param between request from admin or non admin](https://docs.gitlab.com/ee/api/users.html#list-current-user)

thomasboni: But that's the same behavior for `is_admin`, that is already present

svanharmelen: No worries, the fix was easy.

See #1982 for the details, but running `go test./...` to run the tests failed without this change.

PR #1984, opened 2024-08-09
Expose PreReceiveSecretDetectionEnabled project setting

gravis:
This change adds the `PreReceiveSecretDetectionEnabled` field to the `Project` struct. This boolean reflects the state of the project Secret Push Protection feature: https://docs.gitlab.com/ee/user/application_security/secret_detection/secret_push_protection/

See also https://gitlab.com/gitlab-org/gitlab/-/merge_requests/160960


svanharmelen (approved):
I usually don't merge PRs that add undocumented fields as I don't consider them part of the public API (if not documented it can change at any time without an indication other then in the source code), but I'll take this one as I expect it will stay around 👍🏻

PR #1991, opened 2024-08-19
build(github): Added Dependabot for timely dependency updates

flc1125:


svanharmelen:
I'm not a big fan of "blindly" updating dependencies so again not really open to merging this one. Sorry... 😞

PR #1992, opened 2024-08-19
build(github): Added support for more versions of unit tests

flc1125:


.github/workflows/lint_and_test.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     name: Lint and Test - ${{ matrix.go-version }}
     strategy:
       matrix:
-        go-version: [1.19.x, 1.20.x, 1.x]
+        go-version: [1.19.x, 1.20.x, 1.21.x, 1.22.x, 1.23.x, 1.x]
```
svanharmelen, line 13: Instead of adding the new versions to an ever increasing list, I just update the versions to test every now and then. So in this case, if you want to do that for me, I suggest to limit the versions to `1.22.x, 1.23.x, 1.x`.

svanharmelen (changes requested):

svanharmelen (approved):
Thanks @flc1125 👍🏻

PR #1995, opened 2024-08-22
Get scoped group variables

andersparslov:
Added option to set scope when getting group-level variables. Changes were based on functionality which existed for `ProjectVariables` and added corresponding test. The change in `projects.go` came from `make fmt`.


svanharmelen (approved):
Thanks @andersparslov 👍🏻  LGTM!

PR #1997, opened 2024-08-24
member role model follow 17.4.0 spec

Quanalogy:
The current definition of creating and listing member roles were not up to date with the current api. 

I've added all currently defined elements per the api spec, updated test data and simplified the tests a bit.


svanharmelen (approved):
LGTM! Thanks @Quanalogy 👍🏻

PR #2000, opened 2024-08-28
Add endpoint to lists invited groups of a project

habnux:
Adds following endoint https://docs.gitlab.com/ee/api/projects.html#list-a-projects-invited-groups


svanharmelen (changes requested):
Thanks @habnux, but it seems you missed a few bits. I've made some comments but will also add a commit with the fixes so this one can be merged 👍🏻

svanharmelen (approved):

PR #2003, opened 2024-09-02
Bug: Needs to be a list of ProjectGroup

habnux:


svanharmelen (approved):
Good catch 👍🏻 Thanks!

Issue #2008, opened 2024-09-07
Ability to immediately delete projects in Premium/Ultimate tiers


heidiberry:
For the Premium and Ultimate tiers, by default projects are not deleted immediately on request. You can override this by passing two options to the API, full_path and permanently_remove: https://docs.gitlab.com/ee/api/projects.html#delete-project.

Prior to these fields being added to the API, there were no custom options used by the delete project endpoint. Therefore, the function for this has only two parameters: https://github.com/xanzy/go-gitlab/blob/main/projects.go#L1166.

How would you go about adding a DeleteProjectOptions parameter whilst maintaining backwards compatibility?

svanharmelen:
The GitLab API changes a lot so it's pretty hard to keep this package up-to-data without breaking some of the API's every now and then. However annoying it may be, I don't really see how else to fix these issue without creating a massive additional maintenance burden. So I'm good with adding an option struct argument to the delete project method.

PR #2010, opened 2024-09-08
Breaking change: New options when calling DeleteProject

heidiberry:
Allow passing options to the DeleteProject function for immediately deleting a project. Used in Premium/Ultimate tiers where delayed deletion is enabled by default. See API docs: https://docs.gitlab.com/ee/api/projects.html#delete-project

This is a breaking change. Closes https://github.com/xanzy/go-gitlab/issues/2008


svanharmelen (approved):
Thanks @heidiberry 👍🏻

Issue #2011, opened 2024-09-11
ServiceAccounts support


Atoms:
Currently there is no way to select all ServiceAccounts using golang, it would be nice to have such functionality.

~https://docs.gitlab.com/ee/api/group_service_accounts.html#list-service-account-users~
https://docs.gitlab.com/ee/api/user_service_accounts.html

Atoms:
I could try to do this, i'm not a perfect golang programmer, but i'm learning

Atoms:
ooh i just looked at link i posted, i posted wrong link, the correct one is: https://docs.gitlab.com/ee/api/user_service_accounts.html

PR #2012, opened 2024-09-12
Add ListServiceAccounts function to list all service accounts

Atoms:
First try to contribute to this awesome project. 

Implementing ListServiceAccounts from https://docs.gitlab.com/ee/api/user_service_accounts.html#list-all-service-account-users

this is pretty straight forward implementation, only once concern i have, as GroupsService also has same functionality, but just for group service accounts.
Therefore reused ListServiceAccountOptions as they are same. If i would change this, then it could break already existing implementations.

this could fix: https://github.com/xanzy/go-gitlab/issues/2011


svanharmelen:
I'm super busy lately... Will try to have a look later this week!

svanharmelen (approved):
Made a few small tweaks, but other then that it looks good 👍🏻

PR #2015, opened 2024-09-24
add MergeStatus filed into MergeEvent changes struct

cestlive:
when a merge request was created at the first begin, the changes will only contains filed: MergeStatus, which was missing in current version

### DIFF
--- a/environments.go
+++ b/environments.go
@@ -34,16 +34,19 @@ type EnvironmentsService struct {
 //
 // GitLab API docs: https://docs.gitlab.com/ee/api/environments.html
 type Environment struct {
-	ID             int         `json:"id"`
-	Name           string      `json:"name"`
-	Slug           string      `json:"slug"`
-	State          string      `json:"state"`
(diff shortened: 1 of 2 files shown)


cestlive:
when a merge request was created at the first begin, the changes will only contains filed: MergeStatus, which was missing in current version

event_webhook_types.go
```diff
@@ -757,6 +757,10 @@ type MergeEvent struct {
 			Previous int `json:"previous"`
 			Current  int `json:"current"`
 		} `json:"updated_by_id"`
+		MergeStatus struct {
+			Previous string `json:"previous"`
+			Current  string `json:"current"`
+		}`json:"merge_status"`
```
svanharmelen, line 763: Can you put it in the alphabetical ordered location? Thanks!

cestlive: Oh. I didn't notice that, will create a new pr to update

PR #2016, opened 2024-09-25
add MergeStatus filed to struct MergeEvent

cestlive:
when a merge request was created at the first begin, the changes will only contains filed: MergeStatus, which was missing in current version.
Already put it in the alphabetically ordered location, please review.
Thanks! @svanharmelen


svanharmelen (approved):
Thanks 👍🏻

PR #2017, opened 2024-09-25
feat(runners): add maintenance_note on put/get runner details

### LINKED ISSUE #2014: Add "maintenance note" as an option while creating a new runner
When creating a new runner, there is an option in Gitlab API to add "maintenance note".
https://docs.gitlab.com/ee/api/runners.html#create-a-runner

-----
maintenance_note | string | no | Free-form maintenance notes for the runner (1024 characters).
-----

Please add it as part of the runner creation payload.
https://github.com/xanzy/go-gitlab/blob/main/runners.go#L36

jaredallard:
Adds support for the `maintenance_note` field.

Resolves #2014

### DIFF
--- a/runners.go
+++ b/runners.go
@@ -84,6 +84,7 @@ type RunnerDetails struct {
 		Name   string `json:"name"`
 		WebURL string `json:"web_url"`
 	} `json:"groups"`
+	MaintenanceNote string `json:"maintenance_note"`
 
 	// Deprecated: Use Paused instead. (Deprecated in GitLab 14.8)
 	Active bool `json:"active"`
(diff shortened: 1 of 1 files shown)


runners.go
```diff
@@ -84,6 +84,7 @@ type RunnerDetails struct {
 		Name   string `json:"name"`
 		WebURL string `json:"web_url"`
 	} `json:"groups"`
+	MaintenanceNote string `json:"maintenance_note"`
```
svanharmelen, line 87: I try to follow the order in the docs for easy lookup, so that would place this field in between `CantactedAt` and `Name`., can you update that?

jaredallard: Will do when I'm back at my personal laptop (late tomorrow PT)!

jaredallard: Done!

svanharmelen (approved):
Thanks 👍🏻

PR #2021, opened 2024-10-01
Support API endpoint: List memberships for a billable member of a group

Iridias:
Implemented support for the following API endpoint:
https://docs.gitlab.com/ee/api/members.html#list-memberships-for-a-billable-member-of-a-group


svanharmelen (approved):
Thanks @Iridias 👍🏻 Made a few small tweaks (see commit) but otherwise LGTM!

PR #2024, opened 2024-10-08
Add event_name field to the audit events.

lstout:
In https://gitlab.com/gitlab-org/gitlab/-/merge_requests/149449 support for the event_name fields were added. With this MR we enable them in in the Go client as well.


svanharmelen (approved):
Thanks @lstout I'll take it 👍🏻

PR #1792, opened 2023-09-12
Add method to list projects a user has contributed to

soggycactus:
This PR adds a method to the Projects client to allow calling the `contributed_projects` endpoint: https://docs.gitlab.com/ee/api/projects.html#list-projects-a-user-has-contributed-to

This enables users of this library to fetch all projects that a user has contributed to

PR #2028, opened 2024-10-10
Enhancement - make `projects` endpoints configurable

kolluria:
We have GitLab deployed on prem and the URLs are slightly different from what is mentioned in the API specification. For instance, to get a list of projects the user contributed to, instead of invoking `https://gitlab.example.com/api/v4/users/5/contributed_projects` we have to invoke `https://gitlab.example.com/api/v4/users/5/contributed`.

This change accommodates these discrepancies by exporting the endpoints so that users can configure then if they so need it.

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -26,6 +26,11 @@ import (
 	"github.com/hashicorp/go-retryablehttp"
 )
 
+var (
+	UserContributedProjectsEndpoint = "users/%s/contributed_projects"
+	UserStarredProjectsEndpoint     = "users/%s/starred_projects"
+)
+
 // ProjectsService handles communication with the repositories related methods
 // of the GitLab API.
 //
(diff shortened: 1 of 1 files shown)


svanharmelen:
Hmm... Interesting. Are there any docs that explain why these differences exist and/or describe what differences exist for which endpoints? I.e. are these the only two endpoints that are different?

kolluria:
@svanharmelen I might have spoken too soon. I investigated further and turns out some specific endpoints are not working on our on prem deployment like `/api/v4/users/5/contributed_projects`. Let me probe further and see if this change even makes sense.

Thanks.

svanharmelen:
Searching through the project I see this was added about a year ago and has always used these paths: https://github.com/xanzy/go-gitlab/pull/1792 So if version 15.11 is older then a year, I guess not.

RicePatrick:
Yeah, 15.11 would have been about 1.5 years ago. The `contributed_projects` API was introduced in 16.2 (here: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/124579), so I think you just need to update your gitlab version to take advantage of it via `go-gitlab` :)

kolluria:
> Yeah, 15.11 would have been about 1.5 years ago. The `contributed_projects` API was introduced in 16.2 (here: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/124579), so I think you just need to update your gitlab version to take advantage of it via `go-gitlab` :)

Unfortunately I don't know if that's gonna happen any time soon :(

PR #2032, opened 2024-10-10
Add consts for all webhook actions

puerco:
This PR defines constants for all the event webhook action types sent by GitLab.

It also reuses `EventActionCreate` and `EventActionUpdate` to define the custom CommentEventType values already defined in the module.

### DIFF
--- a/types.go
+++ b/types.go
@@ -347,6 +347,29 @@ const (
 	DORAMetricIntervalAll     DORAMetricInterval = "all"
 )
 
+// EventAction defines a constant string identifying the possible `action` values
+// of the various GitLab event webhook payloads.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html#comment-events
+type EventAction string
+
+const (
+	EventActionApproval   EventAction = "approval"
+	EventActionApproved   EventAction = "approved"
+	EventActionAward      EventAction = "award"
+	EventActionClose      EventAction = "close"
+	EventActionCreate     EventAction = "create"
+	EventActionDelete     EventAction = "delete"
+	EventActionMerge      EventAction = "merge"
+	EventActionOpen       EventAction = "open"
+	EventActionReopen     EventAction = "reopen"
+	EventActionRevoke     EventAction = "revoke"
+	EventActionUnapproval EventAction = "unapproval"
+	EventActionUnapproved EventAction = "unapproved"
+	EventActionUpdate     EventAction = "update"
+)
+
 // EventTypeValue represents actions type for contribution events.
 type EventTypeValue string
 


puerco:
I noticed that comments have their own custom action type:

https://github.com/xanzy/go-gitlab/blob/b5e0812497a6475cf9d5f55c068ee0c3a44dbe7a/event_webhook_types.go#L272-L274

but the rest of the payload types have their action typed as string:

https://github.com/xanzy/go-gitlab/blob/b5e0812497a6475cf9d5f55c068ee0c3a44dbe7a/event_webhook_types.go#L1050-L1052

so I didn't add any more specific ones.


This is not modifying any of them, only defining them

JAORMX (approved):

types.go
```diff
@@ -261,15 +261,38 @@ func BuildState(v BuildStateValue) *BuildStateValue {
...
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ee/user/project/integrations/webhook_events.html#comment-events
 type CommentEventAction string
 
 const (
-	CommentEventActionCreate CommentEventAction = "create"
-	CommentEventActionUpdate CommentEventAction = "update"
+	CommentEventActionCreate CommentEventAction = CommentEventAction(EventActionCreate)
+	CommentEventActionUpdate CommentEventAction = CommentEventAction(EventActionUpdate)
```
svanharmelen, line 295: I prefer to not tie these together and revert this change.

types.go
```diff
@@ -261,15 +261,38 @@ func BuildState(v BuildStateValue) *BuildStateValue {
...
+	EventActionCreate     EventAction = "create"
+	EventActionDelete     EventAction = "delete"
+	EventActionMerge      EventAction = "merge"
+	EventActionOpen       EventAction = "open"
+	EventActionReopen     EventAction = "reopen"
+	EventActionRevoke     EventAction = "revoke"
+	EventActionUnapproval EventAction = "unapproval"
+	EventActionUnapproved EventAction = "unapproved"
+	EventActionUpdate     EventAction = "update"
+)
```
svanharmelen, line 285: Stuff (types and consts) in this file is alphabetically ordered, so can you move these to the ordered location?

svanharmelen (changes requested):
Besides the two (minor) comments, what is the end goal with this? Just defining them without using them doesn't have much use, so are you going to add move commits to this PR to also use these?

puerco:
My main goal as a user is to expose the values of the action from the library to applications. I didn't want to modify the payload types in `event_webhook_types.go` to use them not to break compatibility but I can if you are fine with it! It would certainly be better.

types.go
puerco: OK done :+1:

types.go
puerco: Moved :+1:

svanharmelen:
Yeah while it might cause a few issues, I think it's the best what to actually guarantee the list will stay up-to-date over time. So I would say go ahead and update the `EventType` fields to use this new const.

timofurrer:
@puerco we've migrated the repository over to [GitLab.com](https://gitlab.com/gitlab-org/api/client-go).

PR #2034, opened 2024-10-13
Add alert status to project hooks

dany74q:
This commit adds the alert_status field to project hooks, similar to how we return it in group hooks.


svanharmelen (approved):
Thanks @dany74q 👍🏻

PR #2035, opened 2024-10-14
missing protected_branches#deploy_key_id

seyguai:
Protected branches can now have a deploy_key_id allowed to push (since 16.0).
The field was present for the create/update methods, but not the read/list ones.


svanharmelen (approved):
Thanks @seyguai 👍🏻

PR #2041, opened 2024-10-15
feat: added 'NewPutQueryRequest' to support the 'integrations' endpoint

### LINKED ISSUE #2040: Support 'PUT' requests that send parameters via the query string
Hot on the trails of #2039, I've discovered that the current implementation of `NewRequest` doesn't work properly w/ the `integrations` api endpoint. those methods require the parameters to be sent as part of the query string and not as a `json` body in the request. 

see: https://docs.gitlab.com/ee/api/integrations.html#set-up-gitlab-for-slack-app as an example.

jgangemi:
Resolves #2040

I refactored the code ever so slightly to not duplicate building the url or setting the headers on the request. I also didn't want to modify the existing `NewRequest` method signature and cause a cascade of changes elsewhere, so i just added a new method. 

Includes a test.

### DIFF
--- a/gitlab.go
+++ b/gitlab.go
@@ -596,22 +596,7 @@ func (c *Client) setBaseURL(urlStr string) error {
 	return nil
 }
 
-// NewRequest creates a new API request. The method expects a relative URL
-// path that will be resolved relative to the base URL of the Client.
-// Relative URL paths should always be specified without a preceding slash.
-// If specified, the value pointed to by body is JSON encoded and included
-// as the request body.
-func (c *Client) NewRequest(method, path string, opt interface{}, options []RequestOptionFunc) (*retryablehttp.Request, error) {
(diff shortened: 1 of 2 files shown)


svanharmelen:
Cross posting this update so all info is here as well: I thought to remember something about the API actually accepting both query strings and payload body and I found the docs mentioning this: https://docs.gitlab.com/ee/api/rest/#request-payload

If you read closely, you also notice that they write that usually only GET requests use query strings and POST and PUT requests use payload body (quite the default with REST). So that matches exactly with what this package does and so I don't know how you came to the conclusion that this needed to be a call that specifically needs to use query params, but I think your assumption might be incorrect...

jgangemi:
you're right - i'm not sure why this wasn't working originally but it mostly works now w/ #2042, which is still missing a few fields but i'll post there.

PR #2043, opened 2024-10-16
support ListOptions for ListMembershipsForBillableGroupMember

Iridias:
Just realized, that there was no pagination-support for this function/endpoint.
Forgot to add it in my initial commit - my bad :-/

### DIFF
--- a/group_members.go
+++ b/group_members.go
@@ -253,12 +253,18 @@ func (s *GroupsService) ListBillableGroupMembers(gid interface{}, opt *ListBilla
 	return bgm, resp, nil
 }
 
+// ListMembershipsForBillableGroupMemberOptions represents the available ListMembershipsForBillableGroupMember() options.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/members.html#list-memberships-for-a-billable-member-of-a-group
+type ListMembershipsForBillableGroupMemberOptions = ListOptions
+
 // ListMembershipsForBillableGroupMember Gets a list of memberships for a billable member of a group.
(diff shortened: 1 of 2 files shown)


group_members.go
```diff
@@ -258,14 +258,14 @@ func (s *GroupsService) ListBillableGroupMembers(gid interface{}, opt *ListBilla
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/members.html#list-memberships-for-a-billable-member-of-a-group
-func (s *GroupsService) ListMembershipsForBillableGroupMember(gid interface{}, user int, options ...RequestOptionFunc) ([]*BillableUserMembership, *Response, error) {
+func (s *GroupsService) ListMembershipsForBillableGroupMember(gid interface{}, user int, opt *ListOptions, options ...RequestOptionFunc) ([]*BillableUserMembership, *Response, error) {
```
svanharmelen, line 261: Please check the code for examples, but you should not use the `ListOptions` struct directly. Instead you should define a `ListMembershipsForBillableGroupMemberOptions` struct which could be just `type ListMembershipsForBillableGroupMemberOptions = ListOptions`...

Iridias: Ah, ok - wasn't sure about that, as I found some other occurrences, where it's used directly (e.g. in `ListProjectStatusChecks`) 
I'll change it as suggested.

svanharmelen: As far as I can see there are 3 other location where it slipped in, but good to not let the number increase 😏

svanharmelen (approved):
Thanks 👍🏻

PR #2047, opened 2024-10-24
Add username to AddGroupMemberOptions

xinau:
This change adds support for providing a `username` instead of a `user_id` to the `AddGroupMemberOptions`. See https://docs.gitlab.com/ee/api/members.html#add-a-member-to-a-group-or-project


svanharmelen (approved):
Thanks @xinau 👍🏻

Issue #2026, opened 2024-10-09
Add endpoint for update-pipeline-metadata


alexpts:
Docs: https://docs.gitlab.com/ee/api/pipelines.html#update-pipeline-metadata

alexpts:
Prepared PR - https://github.com/xanzy/go-gitlab/pull/2051

alexpts:
Could you please release a new version with new endpoint?

svanharmelen:
Will do soon... Closing this one as finished.

PR #2051, opened 2024-10-29
API pipeline update metadata

alexpts:
Implements for endpoint https://docs.gitlab.com/ee/api/pipelines.html#update-pipeline-metadata

Issue https://github.com/xanzy/go-gitlab/issues/2026

### DIFF
--- a/pipelines.go
+++ b/pipelines.go
@@ -407,18 +407,20 @@ func (s *PipelinesService) DeletePipeline(pid interface{}, pipeline int, options
 	return s.client.Do(req, nil)
 }
 
-// UpdateMetadataOptions represents the available UpdateMetadata() options.
-//
-// https://docs.gitlab.com/ee/api/pipelines.html#update-pipeline-metadata
-type UpdateMetadataOptions struct {
-	Name string `url:"ref" json:"name"`
-}
-
-// UpdateMetadata You can update the metadata of a pipeline. The metadata contains the name of the pipeline.
+// UpdatePipelineMetadataOptions represents the available UpdatePipelineMetadata() options.
 //
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/pipelines.html#update-pipeline-metadata
-func (s *PipelinesService) UpdateMetadata(pid interface{}, pipeline int, opt *UpdateMetadataOptions, options ...RequestOptionFunc) (*Pipeline, *Response, error) {
+type UpdatePipelineMetadataOptions struct {
+	Name *string `url:"name,omitempty" json:"name,omitempty"`
+}
+
+// UpdatePipelineMetadata You can update the metadata of a pipeline. The metadata
+// contains the name of the pipeline.
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/pipelines.html#update-pipeline-metadata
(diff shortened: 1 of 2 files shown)


pipelines.go
```diff
@@ -406,3 +406,35 @@ func (s *PipelinesService) DeletePipeline(pid interface{}, pipeline int, options
 
 	return s.client.Do(req, nil)
 }
+
+// UpdateMetadataOptions represents the available UpdateMetadata() options.
+//
+// https://docs.gitlab.com/ee/api/pipelines.html#update-pipeline-metadata
+type UpdateMetadataOptions struct {
+	Name string `url:"ref" json:"name"`
```
svanharmelen, line 414: This needs to be:
```suggestion
	Name *string `url:"name,omitempty" json:"name,omitempty"`
```

pipelines.go
```diff
@@ -406,3 +406,35 @@ func (s *PipelinesService) DeletePipeline(pid interface{}, pipeline int, options
 
 	return s.client.Do(req, nil)
 }
+
+// UpdateMetadataOptions represents the available UpdateMetadata() options.
+//
+// https://docs.gitlab.com/ee/api/pipelines.html#update-pipeline-metadata
```
svanharmelen, line 412:
```suggestion
// GitLab API docs:
// https://docs.gitlab.com/ee/api/pipelines.html#update-pipeline-metadata
```

pipelines.go
```diff
@@ -406,3 +406,35 @@ func (s *PipelinesService) DeletePipeline(pid interface{}, pipeline int, options
...
+type UpdateMetadataOptions struct {
+	Name string `url:"ref" json:"name"`
+}
+
+// UpdateMetadata You can update the metadata of a pipeline. The metadata contains the name of the pipeline.
```
svanharmelen, line 417: See the contributing guidelines, please keep comments below 80 chars:
```suggestion
// UpdateMetadata You can update the metadata of a pipeline. The metadata 
// contains the name of the pipeline.

```

pipelines.go
```diff
@@ -406,3 +406,35 @@ func (s *PipelinesService) DeletePipeline(pid interface{}, pipeline int, options
...
+//
+// GitLab API docs:
+// https://docs.gitlab.com/ee/api/pipelines.html#update-pipeline-metadata
+func (s *PipelinesService) UpdateMetadata(pid interface{}, pipeline int, opt *UpdateMetadataOptions, options ...RequestOptionFunc) (*Pipeline, *Response, error) {
```
svanharmelen, line 421: If we follow the conventions in the docs, this should be:
```suggestion
func (s *PipelinesService) UpdatePipelineMetadata(pid interface{}, pipeline int, opt *UpdatePipelineMetadataOptions, options ...RequestOptionFunc) (*Pipeline, *Response, error) {
```

svanharmelen (changes requested):
Few comments, but looks good 👍🏻

pipelines.go
alexpts: It is required arguments. Why it will be pointer?

svanharmelen: If you look around in the code base you'll notice this is the pattern used in all option structs. If we would determine this on a per-field bases it would be a bit nasty to use and (even more so) maintain.

alexpts: Ok. I will prepare patch

alexpts:
I fixed suggestions

svanharmelen (approved):
Thanks @alexpts 👍🏻

Issue #2020, opened 2024-09-30
Deploy keys is misssing `expires_at`


petrzjunior:
Deploy keys has supported `expires_at` field since GitLab 15.11, currently the field cannot be set.

See `POST /projects/:id/deploy_keys` docs: https://docs.gitlab.com/ee/api/deploy_keys.html#add-deploy-key

PR #2054, opened 2024-11-02
Add expires_at field when creating project deploy keys

heidiberry:
Fixes https://github.com/xanzy/go-gitlab/issues/2020

The code in this area was a bit confusing. There are two similar endpoints:
- https://docs.gitlab.com/ee/api/deploy_keys.html#add-deploy-key
- https://docs.gitlab.com/ee/api/deploy_keys.html#add-deploy-key-for-a-project

The code linked to the documentation for the first one and was named like it was the first one, but it was actually the second one.

I've updated the documentation link, but adding support for the first one will be difficult in future given it's name is already taken.

### DIFF
--- a/deploy_keys.go
+++ b/deploy_keys.go
@@ -165,10 +165,10 @@ func (s *DeployKeysService) GetDeployKey(pid interface{}, deployKey int, options
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/deploy_keys.html#add-deploy-key-for-a-project
 type AddDeployKeyOptions struct {
-	Key       *string  `url:"key,omitempty" json:"key,omitempty"`
-	Title     *string  `url:"title,omitempty" json:"title,omitempty"`
-	CanPush   *bool    `url:"can_push,omitempty" json:"can_push,omitempty"`
-	ExpiresAt *ISOTime `url:"expires_at,omitempty" json:"expires_at,omitempty"`
+	Key       *string    `url:"key,omitempty" json:"key,omitempty"`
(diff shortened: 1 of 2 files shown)


deploy_keys.go
```diff
@@ -162,19 +163,20 @@ func (s *DeployKeysService) GetDeployKey(pid interface{}, deployKey int, options
...
-// https://docs.gitlab.com/ee/api/deploy_keys.html#add-deploy-key
+// https://docs.gitlab.com/ee/api/deploy_keys.html#add-deploy-key-for-a-project
 type AddDeployKeyOptions struct {
-	Title   *string `url:"title,omitempty" json:"title,omitempty"`
-	Key     *string `url:"key,omitempty" json:"key,omitempty"`
-	CanPush *bool   `url:"can_push,omitempty" json:"can_push,omitempty"`
+	Key       *string  `url:"key,omitempty" json:"key,omitempty"`
+	Title     *string  `url:"title,omitempty" json:"title,omitempty"`
+	CanPush   *bool    `url:"can_push,omitempty" json:"can_push,omitempty"`
+	ExpiresAt *ISOTime `url:"expires_at,omitempty" json:"expires_at,omitempty"`
```
svanharmelen, line 171: `ISOTime` is actually only used for parts where only the data is required (in hindsight it should have been called `ISODate` I guess). So this needs to be:
```suggestion
	ExpiresAt *time.Time `url:"expires_at,omitempty" json:"expires_at,omitempty"`
```

svanharmelen (changes requested):

svanharmelen (approved):
I fixed it so it can be merged... Thanks!

PR #2057, opened 2024-11-06
Add project setting ci_pipeline_variables_minimum_override_role

vindvaki:
This is a companion setting to `restrict_user_defined_variables`, allowing more granular control over who can set job variables and pipeline variables in a project. Currently, the setting `restrict_user_defined_variables` acts as a toggle for the restriction set by `ci_pipeline_variables_minimum_override_role`, but it is likely that `restrict_user_defined_variable` will be deprecated in favor of just `ci_pipeline_variables_minimum_override_role` in the future.

See https://docs.gitlab.com/ee/ci/variables/#restrict-who-can-override-variables

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -157,6 +157,7 @@ type Project struct {
 	MergePipelinesEnabled                    bool               `json:"merge_pipelines_enabled"`
 	MergeTrainsEnabled                       bool               `json:"merge_trains_enabled"`
 	RestrictUserDefinedVariables             bool               `json:"restrict_user_defined_variables"`
+	CIPipelineVariablesMinimumOverrideRole   AccessControlValue `json:"ci_pipeline_variables_minimum_override_role"`
 	MergeCommitTemplate                      string             `json:"merge_commit_template"`
 	SquashCommitTemplate                     string             `json:"squash_commit_template"`
 	AutoDevopsDeployStrategy                 string             `json:"auto_devops_deploy_strategy"`
@@ -854,6 +855,7 @@ type EditProjectOptions struct {
 	CIForwardDeploymentRollbackAllowed        *bool                                `url:"ci_forward_deployment_rollback_allowed,omitempty" json:"ci_forward_deployment_rollback_allowed,omitempty"`
 	CISeperateCache                           *bool                                `url:"ci_separated_caches,omitempty" json:"ci_separated_caches,omitempty"`
 	CIRestrictPipelineCancellationRole        *AccessControlValue                  `url:"ci_restrict_pipeline_cancellation_role,omitempty" json:"ci_restrict_pipeline_cancellation_role,omitempty"`
+	CIPipelineVariablesMinimumOverrideRole    *AccessControlValue                  `url:"ci_pipeline_variables_minimum_override_role,omitempty" json:"ci_pipeline_variables_minimum_override_role,omitempty"`
 	ContainerExpirationPolicyAttributes       *ContainerExpirationPolicyAttributes `url:"container_expiration_policy_attributes,omitempty" json:"container_expiration_policy_attributes,omitempty"`
 	ContainerRegistryAccessLevel              *AccessControlValue                  `url:"container_registry_access_level,omitempty" json:"container_registry_access_level,omitempty"`
 	DefaultBranch                             *string                              `url:"default_branch,omitempty" json:"default_branch,omitempty"`
--- a/projects.go
+++ b/projects.go
@@ -129,50 +129,50 @@ type Project struct {
 		GroupFullPath    string `json:"group_full_path"`
 		GroupAccessLevel int    `json:"group_access_level"`
 	} `json:"shared_with_groups"`
-	Statistics                               *Statistics        `json:"statistics"`
-	Links                                    *Links             `json:"_links,omitempty"`
-	ImportURL                                string             `json:"import_url"`
-	ImportType                               string             `json:"import_type"`
-	ImportStatus                             string             `json:"import_status"`
-	ImportError                              string             `json:"import_error"`
-	CIDefaultGitDepth                        int                `json:"ci_default_git_depth"`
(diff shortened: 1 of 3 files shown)


projects.go
```diff
@@ -157,6 +157,7 @@ type Project struct {
 	MergePipelinesEnabled                    bool               `json:"merge_pipelines_enabled"`
 	MergeTrainsEnabled                       bool               `json:"merge_trains_enabled"`
 	RestrictUserDefinedVariables             bool               `json:"restrict_user_defined_variables"`
+	CIPipelineVariablesMinimumOverrideRole   AccessControlValue `json:"ci_pipeline_variables_minimum_override_role"`
```
vindvaki, line 160: I chose `AccessControlValue` to match existing code, but  for what it's worth, the `AccessControlValue` type seems to have been originally intended for [something completely different](https://github.com/vindvaki/go-gitlab/blob/c4c9b8a1e0290654872f642e3b9f04f084d939ac/types.go#L40-L50). Over time, it seems to have morphed into more of a catch-all string type for different access control values 🤷

vindvaki:
@timofurrer @RicePatrick looks like you are [the new maintainers](https://gitlab.com/gitlab-org/gitlab/-/issues/374736#note_2194881715)? Could either of you please review? Thanks!

RicePatrick:
@vindvaki  - Yep, I'll probably get to it tomorrow; thanks for the ping!

projects.go
RicePatrick: @vindvaki  - I agree, I'm not sure `AccessControlValue` is the right type here. Since the API seems like it's accepting a string value for the Access Level instead of the integer value (which is usually defined using the `AccessLevelValue` type), I think we should probably create a new type for it in `types.go`, especially since `no_one_allowed` seems to deviate from the normal `no_one` that's used elsewhere.

would you mind doing that? Otherwise this LGTM!

vindvaki: @RicePatrick sure, I've created a new type in https://github.com/xanzy/go-gitlab/pull/2057/commits/fc7656ee9b99ff91c8a174ef2456576d0ad4d420. Let me know what you think 🏓

RicePatrick: @vindvaki  - I apologize for nitpicking, but would you mind adding a comment on the new type to match the values of the others within the file? I tried to push a commit for you, but I don't have permissions to do so :)

I'd suggest something like this:

```golang
// CIPipelineVariablesMinimumOverrideRoleValue represents an access control value
// used for managing access to the CI Pipeline Variable Override feature.
//
// GitLab API docs: https://docs.gitlab.com/ee/api/projects.html
type CIPipelineVariablesMinimumOverrideRoleValue = string

// List of available CIPipelineVariablesMinimumOverrideRoleValue levels.
//
// GitLab API docs: https://docs.gitlab.com/ee/api/projects.html
const (
	CIPipelineVariables_NoOneAllowedRole CIPipelineVariablesMinimumOverrideRoleValue = "no_one_allowed"
	CiPipelineVariables_OwnerRole        CIPipelineVariablesMinimumOverrideRoleValue = "owner"
	CiPipelineVariables_MaintainerRole   CIPipelineVariablesMinimumOverrideRoleValue = "maintainer"
	CIPipelineVariables_DeveloperRole    CIPipelineVariablesMinimumOverrideRoleValue = "developer"
)

```

RicePatrick:
Approved the GitHub workflows, and one last nitpick then this LGTM!

projects.go
vindvaki: @RicePatrick thanks, done 🏓

types.go
```diff
@@ -1002,3 +1002,19 @@ func (t *BoolValue) UnmarshalJSON(b []byte) error {
...
+type CIPipelineVariablesMinimumOverrideRoleValue = string
+
+// List of available CIPipelineVariablesMinimumOverrideRoleValue values.
+//
+// GitLab API docs: https://docs.gitlab.com/ee/api/projects.html
+const (
+	CIPipelineVariables_NoOneAllowedRole CIPipelineVariablesMinimumOverrideRoleValue = "no_one_allowed"
+	CiPipelineVariables_OwnerRole        CIPipelineVariablesMinimumOverrideRoleValue = "owner"
+	CiPipelineVariables_MaintainerRole   CIPipelineVariablesMinimumOverrideRoleValue = "maintainer"
+	CIPipelineVariables_DeveloperRole    CIPipelineVariablesMinimumOverrideRoleValue = "developer"
```
timofurrer, line 1019: I think the code base does not use underscores to delimit "type" and "value". So lets do this instead?

```suggestion
	CIPipelineVariablesNoOneAllowedRole CIPipelineVariablesMinimumOverrideRoleValue = "no_one_allowed"
	CiPipelineVariablesOwnerRole        CIPipelineVariablesMinimumOverrideRoleValue = "owner"
	CiPipelineVariablesMaintainerRole   CIPipelineVariablesMinimumOverrideRoleValue = "maintainer"
	CIPipelineVariablesDeveloperRole    CIPipelineVariablesMinimumOverrideRoleValue = "developer"
```

projects_test.go
```diff
@@ -280,13 +280,15 @@ func TestListOwnedProjects(t *testing.T) {
 func TestEditProject(t *testing.T) {
 	mux, client := setup(t)
 
-	var developerAccessLevel AccessControlValue = "developer"
+	var developerRole AccessControlValue = "developer"
+	developerPipelineVariablesRole := CIPipelineVariables_DeveloperRole
```
timofurrer, line 284: Goes together with the other comment:

```suggestion
	developerPipelineVariablesRole := CIPipelineVariablesDeveloperRole
```

timofurrer (changes requested):
@vindvaki sorry for another nitpick. I'm trying to apply myself and merge.

timofurrer (approved):

timofurrer:
@vindvaki @RicePatrick I'm going to merge here. Thanks Hordur! I assume you need a release?

PR #2059, opened 2024-11-12
feat: implement 'allowed_email_domains_list' group attribute

jeremad:


groups.go
```diff
@@ -384,6 +385,7 @@ type CreateGroupOptions struct {
 	SharedRunnersMinutesLimit       *int                                    `url:"shared_runners_minutes_limit,omitempty" json:"shared_runners_minutes_limit,omitempty"`
 	ExtraSharedRunnersMinutesLimit  *int                                    `url:"extra_shared_runners_minutes_limit,omitempty" json:"extra_shared_runners_minutes_limit,omitempty"`
 	IPRestrictionRanges             *string                                 `url:"ip_restriction_ranges,omitempty" json:"ip_restriction_ranges,omitempty"`
+	AllowedEmailDomainsList         *string                                 `url:"allowed_email_domains_list,omitempty" json:"allowed_email_domains_list,omitempty"`
```
timofurrer, line 388: @jeremad is this really a supported field for the create endpoint? It's at least not documented, see https://docs.gitlab.com/ee/api/groups.html#create-a-group.

groups_test.go
```diff
@@ -1124,3 +1124,51 @@ func TestEditGroupPushRules(t *testing.T) {
...
+	if !reflect.DeepEqual(want, group) {
+		t.Errorf("Groups.CreateGroup returned %+v, want %+v", group, want)
+	}
+}
+
+func TestUpdateGroupWithAllowedEmailDomainsList(t *testing.T) {
+	mux, client := setup(t)
+
+	mux.HandleFunc("/api/v4/groups/1",
+		func(w http.ResponseWriter, r *http.Request) {
```
timofurrer, line 1158: I think that we should actually test here if the request contain the `allowed_email_domains_list` field, right?

timofurrer (changes requested):
@jeremad I did a first round of review and left some comments. Back to you 🏓

groups.go
jeremad: No you are right, it is only available for updates, I copied the pattern of `ip_restriction_ranges` which works the same way. But this may be bugged as well. I'll dig into that :pick:

groups_test.go
```diff
@@ -783,49 +783,40 @@ func TestUnshareGroupFromGroup(t *testing.T) {
...
+			var bodyJson map[string]interface{}
+			err = json.Unmarshal(body, &bodyJson)
+			if err != nil {
+				t.Fatalf("Failed to parse the request body into JSON. Error: %v", err)
+			}
 
-func TestUpdateGroupWithIPRestrictionRanges(t *testing.T) {
-	mux, client := setup(t)
+			if bodyJson["ip_restriction_ranges"] != ipRange {
+				t.Fatalf("Test failed. `allowed_email_domains_list` expected to be '%v', got %v", ipRange, bodyJson["ip_restriction_ranges"])
```
timofurrer, line 806:
```suggestion
				t.Fatalf("Test failed. `ip_restriction_ranges` expected to be '%v', got %v", ipRange, bodyJson["ip_restriction_ranges"])
```

jeremad: oh thanks

jeremad:
@timofurrer fixed!

timofurrer (approved):
@jeremad LGTM 🎉

PR #2063, opened 2024-11-19
feat: add dependency list export client service

lmphil:
This PR adds support for the GitLab Dependency List Export API:
https://docs.gitlab.com/ee/api/dependency_list_export.html


dependency_list_export.go
```diff
@@ -0,0 +1,107 @@
...
+// SBOM exports can be only accessed by the export’s author.
+//
+// GitLab docs:
+// https://docs.gitlab.com/ee/api/dependency_list_export.html#create-a-pipeline-level-dependency-list-export
+func (s *DependencyListExportService) CreateDependencyListExport(id int, options ...RequestOptionFunc) (*DependencyListExport, *Response, error) {
+	// POST /pipelines/:id/dependency_list_exports
+	createExportPath := fmt.Sprintf("pipelines/%d/dependency_list_exports", id)
+
+	opt := &createDependencyListExportOptions{
+		ExportType: *Ptr("sbom"),
```
timofurrer, line 48: That's interesting 🤔 I wonder if we should already expose the `CreateDependencyListExportOptions` as a parameter so that if there will ever be another `export_type` than `"sbom"` it wouldn't introduce a breaking change.

To guarantee the same user-friendliness we could just set the `ExportType` to "sbom" in case either the `CreateDependencyListExportOptions` is `nil` or the `ExportType` field is `nil`. 

WDYT?

dependency_list_export.go
```diff
@@ -0,0 +1,107 @@
...
+// detected in a pipeline.
+//
+// If an authenticated user does not have permission to read_dependency, this request returns a 403
+// Forbidden status code.
+//
+// SBOM exports can be only accessed by the export’s author.
+//
+// GitLab docs:
+// https://docs.gitlab.com/ee/api/dependency_list_export.html#create-a-pipeline-level-dependency-list-export
+func (s *DependencyListExportService) CreateDependencyListExport(id int, options ...RequestOptionFunc) (*DependencyListExport, *Response, error) {
```
timofurrer, line 43: WDYT about being a little bit more explicit with the argument name?

```suggestion
func (s *DependencyListExportService) CreateDependencyListExport(pipelineID int, options ...RequestOptionFunc) (*DependencyListExport, *Response, error) {
```

dependency_list_export.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	return export, resp, nil
+}
+
+// DownloadDependencyListExport downloads a single dependency list export.
+//
+// GitLab docs:
+// https://docs.gitlab.com/ee/api/dependency_list_export.html#download-dependency-list-export
+func (s *DependencyListExportService) DownloadDependencyListExport(id int, options ...RequestOptionFunc) (string, *Response, error) {
```
timofurrer, line 91: This seems to be one of the rare cases where it probably doesn't make sense to introduce a type for the return value. However, let's return a byte slice instead of a string.

And what do you think about pointing out `github.com/CycloneDX/cyclonedx-go` as a potential way to consume the returned `[]byte` ?

Seems to be fairly simple:

```go
bom := new(cdx.BOM)
decoder := cdx.NewBOMDecoder(reader, cdx.BOMFileFormatJSON)
if err = decoder.Decode(bom); err != nil {
	panic(err)
}
```

WDYT?

dependency_list_export_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestCreateDependencyListExport(t *testing.T) {
+	mux, client := setup(t)
+
+	mux.HandleFunc("/api/v4/pipelines/1234/dependency_list_exports", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, http.MethodPost)
```
timofurrer, line 15: Let's check here if the request actually got the correct `export_type`.

timofurrer (changes requested):
@lmphil thanks for the contribution! I did a first round of review and left a couple of comments for you - back to you 🏓

dependency_list_export.go
lmphil: I agree. I'll change it.

lmphil: I defaulted the `CreateDependencyListExportOptions` to contain "ExportType: sbom" if it's `nil`, but for the `ExportType`, since it cannot be `nil` I decided to let it send whatever string value the user sets to the API, because in that case the user has to have explicitly set it to `""` .

dependency_list_export.go
lmphil: I did the change to return a byte slice.

What would be the preferred way to suggest using `github.com/CycloneDX/cyclonedx-go` to parse the return value?
Would a function comment like this work?
```go
// DownloadDependencyListExport downloads a single dependency list export.
//
// The github.com/CycloneDX/cyclonedx-go package can be used to parse the returned bytes.
//
// GitLab docs:
// https://docs.gitlab.com/ee/api/dependency_list_export.html#download-dependency-list-export
func (s *DependencyListExportService) DownloadDependencyListExport...
```

timofurrer: Yes, I've imagined a function comment. The idea behind this is that a user may be surprised and ask themselves what to do if we just return a byte slice.

dependency_list_export.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	var sbomBuffer bytes.Buffer
+	resp, err := s.client.Do(req, &sbomBuffer)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return sbomBuffer.Bytes(), resp, nil
+}
```
timofurrer, line 113: Thinking about the interface a little bit more. WDYT about returning an `io.Reader` instead? That way it can directly be used with `cyclondx-go` or otherwise converted to bytes if pleased:

```suggestion
// DownloadDependencyListExport downloads a single dependency list export.
//
// The github.com/CycloneDX/cyclonedx-go package can be used to parse the returned data from the io.Reader.
//
// GitLab docs:
// https://docs.gitlab.com/ee/api/dependency_list_export.html#download-dependency-list-export
func (s *DependencyListExportService) DownloadDependencyListExport(id int, options ...RequestOptionFunc) (io.Reader, *Response, error) {
	// GET /dependency_list_exports/:id/download
	downloadExportPath := fmt.Sprintf("dependency_list_exports/%d/download", id)

	req, err := s.client.NewRequest(http.MethodGet, downloadExportPath, nil, options)
	if err != nil {
		return nil, nil, err
	}

	var sbomBuffer bytes.Buffer
	resp, err := s.client.Do(req, &sbomBuffer)
	if err != nil {
		return nil, resp, err
	}

	return sbomBuffer, resp, nil
}
```

timofurrer (changes requested):
@lmphil thanks for the changes! I have some more additions - back to you 🏓

dependency_list_export.go
lmphil: That works. I also added your concrete example to the function comment, if you don't mind.

lmphil:
@timofurrer Done and tested. Thanks for the feedback. 🏓

dependency_list_export.go
```diff
@@ -0,0 +1,125 @@
...
+//	decoder := cdx.NewBOMDecoder(reader, cdx.BOMFileFormatJSON)
+//
+//	if err = decoder.Decode(sbom); err != nil {
+//		panic(err)
+//	}
+//
+// ----------------------------------------------
+//
+// GitLab docs:
+// https://docs.gitlab.com/ee/api/dependency_list_export.html#download-dependency-list-export
```
timofurrer, line 108: We should use a [Go doc code block](https://tip.golang.org/doc/comment#code) for this:

```suggestion
// DownloadDependencyListExport downloads a single dependency list export.
//
// The github.com/CycloneDX/cyclonedx-go package can be used to parse the data from the returned io.Reader.
//
//	sbom := new(cdx.BOM)
//	decoder := cdx.NewBOMDecoder(reader, cdx.BOMFileFormatJSON)
//
//	if err = decoder.Decode(sbom); err != nil {
//		panic(err)
//	}
//
// GitLab docs:
// https://docs.gitlab.com/ee/api/dependency_list_export.html#download-dependency-list-export
```

timofurrer (changes requested):
@lmphil I've left one last suggestion - then we are good for the merge.

Thanks for pulling through 🤝

timofurrer:
I've actually applied the patch myself and will merge :)

dependency_list_export.go
lmphil: @timofurrer  Nice. I didn't know about those.

PR #2064, opened 2024-11-21
Add support for hidden field to group variables

yogeshlonkar:


group_variables.go
```diff
@@ -127,6 +128,7 @@ type CreateGroupVariableOptions struct {
 	Description      *string            `url:"description,omitempty" json:"description,omitempty"`
 	EnvironmentScope *string            `url:"environment_scope,omitempty" json:"environment_scope,omitempty"`
 	Masked           *bool              `url:"masked,omitempty" json:"masked,omitempty"`
+	Hidden           *bool              `url:"hidden,omitempty" json:"hidden,omitempty"`
```
timofurrer, line 131: From the [docs](https://docs.gitlab.com/ee/api/group_level_variables.html#create-variable:~:text=variable%20is%20masked.-,masked_and_hidden,-boolean) it seems the field is called `masked_and_hidden`. Did `hidden` work for you?

timofurrer (dismissed):
@yogeshlonkar thanks for the contribution 🤝 

I've left some clarification questions similar to the ones I've left in the other PR #2065.

Back to you 🏓

group_variables.go
yogeshlonkar: Same comment as https://github.com/xanzy/go-gitlab/pull/2065#pullrequestreview-2452420467

RicePatrick:
Thanks for fixing the models for group to align with the models for project. Merging so this is available during the migration :)

RicePatrick (approved):

PR #2065, opened 2024-11-21
Add support for hidden field to project variables

yogeshlonkar:
GitLab has inconsistent naming of the filed for hidden variables.
It seems for keeping UI consistant, GitLab chose to introduce different
naming for the same field. This could have been easy handled in UI by
setting two variables on API request instead of introducing combined
field for UI.

I have added 2 test cases for create and update project API as per the
gitlab.com API that I manually tested and added similar unit tests.

### DIFF
--- a/project_variables.go
+++ b/project_variables.go
@@ -41,7 +41,7 @@ type ProjectVariable struct {
 	VariableType     VariableTypeValue `json:"variable_type"`
 	Protected        bool              `json:"protected"`
 	Masked           bool              `json:"masked"`
-	Hidden           bool              `json:"hidden"`
+	MaskedAndHidden  bool              `json:"masked_and_hidden"`
 	Raw              bool              `json:"raw"`
 	EnvironmentScope string            `json:"environment_scope"`
 	Description      string            `json:"description"`
@@ -133,7 +133,7 @@ type CreateProjectVariableOptions struct {
 	Description      *string            `url:"description,omitempty" json:"description,omitempty"`
 	EnvironmentScope *string            `url:"environment_scope,omitempty" json:"environment_scope,omitempty"`
 	Masked           *bool              `url:"masked,omitempty" json:"masked,omitempty"`
-	Hidden           *bool              `url:"hidden,omitempty" json:"hidden,omitempty"`
+	MaskedAndHidden  *bool              `url:"masked_and_hidden,omitempty" json:"masked,omitempty"`
 	Protected        *bool              `url:"protected,omitempty" json:"protected,omitempty"`
 	Raw              *bool              `url:"raw,omitempty" json:"raw,omitempty"`
 	VariableType     *VariableTypeValue `url:"variable_type,omitempty" json:"variable_type,omitempty"`
@@ -175,7 +175,6 @@ type UpdateProjectVariableOptions struct {
 	EnvironmentScope *string            `url:"environment_scope,omitempty" json:"environment_scope,omitempty"`
 	Filter           *VariableFilter    `url:"filter,omitempty" json:"filter,omitempty"`
 	Masked           *bool              `url:"masked,omitempty" json:"masked,omitempty"`
-	Hidden           *bool              `url:"hidden,omitempty" json:"hidden,omitempty"`
 	Protected        *bool              `url:"protected,omitempty" json:"protected,omitempty"`
 	Raw              *bool              `url:"raw,omitempty" json:"raw,omitempty"`
 	VariableType     *VariableTypeValue `url:"variable_type,omitempty" json:"variable_type,omitempty"`
--- a/project_variables.go
+++ b/project_variables.go
@@ -41,7 +41,7 @@ type ProjectVariable struct {
 	VariableType     VariableTypeValue `json:"variable_type"`
 	Protected        bool              `json:"protected"`
 	Masked           bool              `json:"masked"`
-	MaskedAndHidden  bool              `json:"masked_and_hidden"`
+	Hidden           bool              `json:"hidden"`
 	Raw              bool              `json:"raw"`
 	EnvironmentScope string            `json:"environment_scope"`
 	Description      string            `json:"description"`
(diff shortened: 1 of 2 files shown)


project_variables.go
```diff
@@ -132,6 +133,7 @@ type CreateProjectVariableOptions struct {
 	Description      *string            `url:"description,omitempty" json:"description,omitempty"`
 	EnvironmentScope *string            `url:"environment_scope,omitempty" json:"environment_scope,omitempty"`
 	Masked           *bool              `url:"masked,omitempty" json:"masked,omitempty"`
+	Hidden           *bool              `url:"hidden,omitempty" json:"hidden,omitempty"`
```
timofurrer, line 136: From the [docs](https://docs.gitlab.com/ee/api/project_level_variables.html#create-a-variable:~:text=masked.%20Default%3A%20false-,masked_and_hidden,-boolean) it looks like it's called `masked_and_hidden`. Did `hidden` work for you?

project_variables.go
```diff
@@ -173,6 +175,7 @@ type UpdateProjectVariableOptions struct {
 	EnvironmentScope *string            `url:"environment_scope,omitempty" json:"environment_scope,omitempty"`
 	Filter           *VariableFilter    `url:"filter,omitempty" json:"filter,omitempty"`
 	Masked           *bool              `url:"masked,omitempty" json:"masked,omitempty"`
+	Hidden           *bool              `url:"hidden,omitempty" json:"hidden,omitempty"`
```
timofurrer, line 178: From the [docs](https://docs.gitlab.com/ee/api/project_level_variables.html#update-a-variable) I don't see that this field is supported in the update API. Did `hidden` work for you?

timofurrer (changes requested):
@yogeshlonkar thanks for the contribution 🤝  I've left some clarification questions on the correctness of the field / field naming. Back to you 🏓

project_variables.go
yogeshlonkar: API in docs perspective `masked_and_hidden` is consistent name. I have not tested this against the api.

project_variables.go
yogeshlonkar: This feature is not matured, You can close the MR or leave it open until this field is available in update api.

project_variables.go
yogeshlonkar: There also might be validation error as masked and masked_and_hidden should be mutually exclusive as per UI

project_variables.go
timofurrer: I think it's up to you. We can already merge what is supported and iterate. No need to wait for a "matured" upstream API. 

IMHO, we should move on with what is supported upstream.

project_variables.go
timofurrer: Well, we need to support in go-gitlab exactly with what the upstream API supports. 

> I have not tested this against the api.

I highly recommend doing this :)

yogeshlonkar: The name is changed as per api documentation.

>  > I have not tested this against the api.
>
> I highly recommend doing this :)

Still not tested as its not easy to setup GitLab by any means

project_variables.go
yogeshlonkar: Resolving this as I have removed the field from update options since GitLab doesn't support it

project_variables.go
```diff
@@ -41,6 +41,7 @@ type ProjectVariable struct {
 	VariableType     VariableTypeValue `json:"variable_type"`
 	Protected        bool              `json:"protected"`
 	Masked           bool              `json:"masked"`
+	MaskedAndHidden  bool              `json:"masked_and_hidden"`
```
RicePatrick, line 44:
```suggestion
	Masked  bool              `json:"masked"`
```

When reading the variable, you get the `masked` result back, not `masked_and_hidden` - Masked and Hidden have a value of `true` for this.

project_variables.go
```diff
@@ -132,6 +133,7 @@ type CreateProjectVariableOptions struct {
 	Description      *string            `url:"description,omitempty" json:"description,omitempty"`
 	EnvironmentScope *string            `url:"environment_scope,omitempty" json:"environment_scope,omitempty"`
 	Masked           *bool              `url:"masked,omitempty" json:"masked,omitempty"`
+	MaskedAndHidden  *bool              `url:"masked_and_hidden,omitempty" json:"masked,omitempty"`
```
RicePatrick, line 136:
```suggestion
	MaskedAndHidden  *bool              `url:"masked_and_hidden,omitempty" json:"masked_and_hidden,omitempty"`
```

Similarly when creating the variable, we need to pass in `masked_and_hidden` whether we're using the URL or the JSON, so this value needs to be updated.

RicePatrick (changes requested):
I went ahead and tested this against a locally running GitLab with the 2 suggested changes, and the tests pass. If you'd please make those changes, I'm good to merge.

RicePatrick:
Here is the test I used to test the above 2 suggestions:

```golang
func TestProjectVariablesService_MaskAndHidden_RealGitLab(t *testing.T) {
	client, err := NewClient("<token>",
		WithBaseURL("http://localhost:8085"),
	)
	if err != nil {
		t.Fatalf("Failed to create client: %v", err)
	}
	// create a project
	project, _, err := client.Projects.CreateProject(&CreateProjectOptions{
		Name: Ptr(fmt.Sprintf("test-%d", time.Now().UnixMicro())),
	})
	if err != nil {
		t.Fatalf("Failed to create project: %v", err)
	}
	projectVar, _, err := client.ProjectVariables.CreateVariable(project.ID, &CreateProjectVariableOptions{
		Key:             Ptr("test"),
		Value:           Ptr("testtesttest"), // must have a min of 8 characters
		MaskedAndHidden: Ptr(true),
	})
	if err != nil {
		t.Fatalf("Failed to create project variable: %v", err)
	}

	assert.True(t, projectVar.Hidden)
}
```

yogeshlonkar:
@RicePatrick  @timofurrer can you please review PR again

RicePatrick:
Updated code looks good; merging so it's available for the migration tomorrow :)

RicePatrick (approved):

PR #2069, opened 2024-11-28
add 'name' and 'description' to project hooks structs

cgahlon:
Adds the `name` and `description` fields to the `ProjectHook`, `AddProjectHookOptions`, and `EditProjectHookOptions` structs.  These fields are supported by the GitLab 17.x API.  Also added them to the existing test.

I have been using this code in my fork to automatically generate hooks without any issues.  Thanks for all your work on this project!

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -1270,6 +1270,8 @@ type HookCustomHeader struct {
 type ProjectHook struct {
 	ID                        int                 `json:"id"`
 	URL                       string              `json:"url"`
+	Name                      string              `json:"name,omitempty"`
+	Description               string              `json:"description,omitempty"`
 	ConfidentialNoteEvents    bool                `json:"confidential_note_events"`
 	ProjectID                 int                 `json:"project_id"`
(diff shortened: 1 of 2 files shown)


projects.go
```diff
@@ -1270,6 +1270,8 @@ type HookCustomHeader struct {
 type ProjectHook struct {
 	ID                        int                 `json:"id"`
 	URL                       string              `json:"url"`
+	Name                      string              `json:"name,omitempty"`
+	Description               string              `json:"description,omitempty"`
```
RicePatrick, line 1274: You don't need the `omitempty` on the main struct used for reads, if you wouldn't mind removing it.

RicePatrick (changes requested):
One small change, otherwise I think we're good to merge. I'll kick off the actions as well to validate linting and tests.

projects.go
cgahlon: Fixed!

RicePatrick (approved):
Looks good; merging!

Issue #2060, opened 2024-11-15
Migration Announcement: go-gitlab moves to GitLab.com


timofurrer:
# 🗣️ Migration Announcement: go-gitlab moves to GitLab.com

We're excited to announce that this Go package previously hosted at `github.com/xanzy/go-gitlab`
is moving to a new home at `gitlab.com/gitlab-org/client-go`! :tada:

-> **When**: soon, we'll provide an update in the coming days.

## What does this mean for you?

This migration will be straightforward:

- To continue using the package, simply update your import path and
  `go.mod` reference from `github.com/xanzy/go-gitlab` to `gitlab.com/gitlab-org/client-go`.
- No other breaking changes are expected.

## Why the move?

This transition ensures better alignment with GitLab's ecosystem and simplifies maintenance.
If you are contributing to this package you are most likely a familiar GitLab user already.

The project remains open source with the same license and community-driven, 
with ongoing contributions encouraged and welcomed.

## Maintainers Update

Sander (@svanharmelen), the original author of the package, has stepped down and [offered](https://gitlab.com/gitlab-org/gitlab/-/issues/374736) 
GitLab contributors to maintain it.
We extend our heartfelt thanks to Sander for his amazing work in creating and maintaining this library! :pray:
Moving forward, Timo (@timofurrer) and Patrick (@RicePatrick) will continue maintaining
the project to ensure its stability and growth.

## Next Steps

- We'll update this issue with an exact date of the migration.
- Feel free to share your feedback and suggestions here in the issue.

Thank you for being part of this journey!
We're looking forward to continuing the development of this library together with the vibrant Go and GitLab communities.

Happy coding! :rocket:

\- Timo

timofurrer:
## Migration date: 2024-12-10 at 14:00 UTC

**Quick Updates**:

- Decided on final migration time: **Tuesday 2024-12-10 at around ~14:00 UTC** we'll start the migration.
- Decided on final location: we'll migrate to **https://gitlab.com/gitlab-org/api/client-go**.

**Important information**:

- If you want your issues, pull requests, comments etc. to be linked to your GitLab account, please make sure your e-mail addresses can be mapped. See https://docs.gitlab.com/ee/user/project/import/github.html#accounts-for-user-contribution-mapping
- All contributions to this project during the migration will be "dangling on GitHub" - we'll make sure to let you know if this has happened and help to migrate the contribution over to GitLab.
- This project will be archived once the migration is complete.
[...]

PR #2070, opened 2024-12-03
Extend project lint result struct with includes

TheCodear:
The Gitlab API exposes a list of (recursive) includes when linting a projects ci configuration. The ProjectLintResult struct representing the API response does not reflect this list, so a user of the SDK cannot use this information from the API. It is also not possible to grab this information via response body parsing outside of the go-gitlab client, as the body is already closed when the client code can access it.

So, in short, this change allows the client code to access the list of includes returned by the Gitlab API.

My only concern is the documentation of the includes type. It is not documented in the api docs of the lint API (see https://docs.gitlab.com/ee/api/lint.html#validate-a-projects-cicd-configuration). Instead, one can find the specification in the openapi spec, see https://gitlab.com/gitlab-org/gitlab/-/blob/master/doc/api/openapi/openapi_v2.yaml?plain=1#L52442. But relying on this spec is inconsistent with the rest of the go-gitlab library...

Please let me know if this still is an acceptable change - and of course whether there is anything else missing (e.g. if further tests are required).

Thank you!

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -44,10 +44,26 @@ type LintResult struct {
 // GitLab API docs:
 // https://docs.gitlab.com/ee/api/lint.html#validate-a-projects-ci-configuration
 type ProjectLintResult struct {
-	Valid      bool     `json:"valid"`
-	Errors     []string `json:"errors"`
-	Warnings   []string `json:"warnings"`
-	MergedYaml string   `json:"merged_yaml"`
+	Valid      bool      `json:"valid"`
+	Errors     []string  `json:"errors"`
+	Warnings   []string  `json:"warnings"`
+	MergedYaml string    `json:"merged_yaml"`
+	Includes   []Include `json:"includes"`
+}
+
+// Include contains the details about an include block in the .gitlab-ci.yml file.
+// It is used in ProjectLintResult.
+//
+// Reference can be found at the lint API endpoint in the openapi yaml:
+// https://gitlab.com/gitlab-org/gitlab/-/blob/master/doc/api/openapi/openapi_v2.yaml
+type Include struct {
+	Type           string                 `json:"type"`
+	Location       string                 `json:"location"`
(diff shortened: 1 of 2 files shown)


RicePatrick:
@timofurrer  -what're your thoughts about "undocumented" features that are only in the OpenAPI spec? I believe that OpenAPI spec is auto-generated, so it may include fields that aren't technically meant for "public" use, but it seems to me as long as it's present there it seems reasonable to support it.

I suppose we could open a documentation update MR and see what comes from that too; if it's denied to include in the docs, I don't think we should support it here.

timofurrer:
@RicePatrick 

> I suppose we could open a documentation update MR and see what comes from that too; if it's denied to include in the docs, I don't think we should support it here.

Yes, probably this or an issue. There have been other instances where we asked for doc changes or made the updates. But also, where we've just supported them here. I very much prefer the former though.

TheCodear:
Hey @RicePatrick and @timofurrer thanks for joining in!

There is a similar thing with the jobs field, there is an `include_jobs` attribute documented in the request (see https://docs.gitlab.com/ee/api/lint.html#validate-a-projects-cicd-configuration) but the response in case this attribute is set to `true` is omitted - so it might be that the documentation (resp. the examples) is just not up to date to the latest api changes..

I have started an issue on Gitlab on this: https://gitlab.com/gitlab-org/gitlab/-/issues/508011 

Although I'm not sure if the issue is placed correctly - I'll maybe route this also via our Gitlab Support...

timofurrer:
@TheCodear thanks for creating the upstream issue - I've triaged it. Let's see.

TheCodear:
Perfect! thanks @timofurrer

timofurrer:
@TheCodear we've migrated the repository over to [GitLab.com](https://gitlab.com/gitlab-org/api/client-go).

PR #2072, opened 2024-12-06
add ci_id_token_sub_claim_components to editProjectOptions

duncan485:
This adds the ci_id_token_sub_claim_components to the gitlab project. 

Note that it is currently only possible to edit this setting, so I did not add in the createProjectOptions (awaiting API implementation from gitlab: https://gitlab.com/gitlab-org/gitlab/-/issues/477260 ). 

Also see https://gitlab.com/gitlab-org/gitlab/-/issues/360592#note_2060395819

### DIFF
--- a/projects.go
+++ b/projects.go
@@ -143,6 +143,7 @@ type Project struct {
 	CIOptInJWT                               bool                                        `json:"ci_opt_in_jwt"`
 	CIAllowForkPipelinesToRunInParentProject bool                                        `json:"ci_allow_fork_pipelines_to_run_in_parent_project"`
 	CIRestrictPipelineCancellationRole       AccessControlValue                          `json:"ci_restrict_pipeline_cancellation_role"`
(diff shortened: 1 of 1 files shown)


RicePatrick:
This seems a little risky - 
1. The field isn't documented as part of the official API
2. The example posted isn't a GitLab employee (to my knowledge), so it's not necessarily a "sanctioned" workaround
3. There's explicitly a story open to add separate API support.

It feels like we should wait until there is an official API support; the stance on this library is to not support "non-public" APIs unless there's a very compelling reason to do so.

Any thoughts here @timofurrer ?

duncan485:
Hi @RicePatrick I get what you are saying. And can't say I disagree, shall I convert this one to WIP, and revisit once the api implementation on the giltab side has been done?

RicePatrick:
@duncan485  - given we'll be migrating this project to gitlab.com (hopefully) tomorrow, I'd say let's close it and we can revisit it there once the gitlab side is done, yeah :)

